PR #5, opened 2016-11-05
feat: computed pathParts

mattiamanzati:
Like the todo says, this PR make the "pathParts" (a.k.a. the _path) property of the "Node" a computed value. This is done by letting the parent node tell where a child node is located (if direct children).

Have a nice day!

### DIFF
--- a/src/core/node.ts
+++ b/src/core/node.ts
@@ -21,8 +21,10 @@ export abstract class Node /* TODO: implements INode*/ {
         }
 
         // get the key
-        const key = this._parent.getPathForNode(this)
-        if(key !== null){
+        const keys = this._parent.getChildNodes()
+            .filter(([key, node]) => node === this)
+        if(keys.length > 0){
+            const [key] = keys[0]
             return this._parent.pathParts.concat([key])
         }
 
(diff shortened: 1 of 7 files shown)


src/core/node.ts
```diff
@@ -103,7 +112,7 @@ export abstract class Node /* TODO: implements INode*/ {
     // TODO: needs improvements, now called too often. Should be set propertly when applying snapshots / calling factories immediately!
     // TODO: should not be possible to change just subpath?
     setParent(newParent: Node | null, subpath: string | null = null) {
-        if (this.parent === newParent && this.pathParts[this.pathParts.length - 1] === subpath)
```
mweststrate: Not sure whether this can be removed; item might be moved within same parent? Or maybe this is always the combination of two operations indeed

mattiamanzati: Ideally, that subpath was used before only to update the children paths. This if avoided unecessary subpath updates. Since now it is computed, we could safely set the parent whenever we want and the path will be computed accordly.

mweststrate:
Ah good point! Thanks for the explanation

Op di 8 nov. 2016 om 08:53 schreef mattiamanzati <notifications@github.com>:

Issue #20, opened 2017-01-04
Idea: Basic type system


mattiamanzati:
Right now, mobx-state-tree has a very basic type system. Basically, it knows if a node is an array, an object, a reference or a primitive.

Imagine this real world scenario:
I have a stacked navigation in my app, so I will store all my app screens state in an array named "stack" and I will push/pull instances from it. But this array will contain a lot of screens; I would say that in typescript would be an array of an union type of all the available application screens.

Sample TypeScript typing.
```typescript
interface LoginForm{
    type: 'login'
    username: string
    password: string
}

interface SignupForm{
    type: 'signup'
    username: string
    email: string
    email_repeat: string
    password: string
    password_repeat: string
}

type ApplicationForms = LoginForm | SignupForm

interface ApplicationState{
    stack: ApplicationForms[]
}
```

This will need union types in mobx-state-tree. To implement them, we need a standard way to handle types in mobx-state-tree.

My proposal is to modify the factory definition in something like this:
```typescript
export interface IModelFactory<S, T> {
    (snapshot?: S, env?: Object): T & IModel
    isModelFactory: true
    factoryName: string
    is: (snapshotOrNode: any) => boolean
    dispatchType: (snapshotOrNode: any) => IModelFactory<any, any>
    identity: boolean
    config: Object
}
```
It basically introduces 3 new props:
- **is**: Given a snapshot or a Node, it returns if the snapshot is assignable to that Model Factory type.
- **dispatchType**: Given a snapshot that is assignable to this type, resolve the exact type. In union types this will resolve to the correct subtype.
- **identity**: If true, no dispatch is needed and you can assume that dispatchType(snapshot) will return the type itself. Primary types, arrayOf(), mapOf() are identiy. Union types are not.

NB: Types will always receive the snapshot as input to determine the type.

PS: Thanks to @gcanti because this environment reminded it's awesome work on tcomb!

mattiamanzati:
Closed and done in #21

PR #21, opened 2017-01-04
WIP: Feat: initial type system

mattiamanzati:
As of #20!

### DIFF
--- a/src/core/factories.ts
+++ b/src/core/factories.ts
@@ -1,36 +1,69 @@
 import {action} from "mobx"
-import {extend} from "../utils"
+import {extend, fail} from "../utils"
 import {getNode, hasNode, NodeConstructor} from "./node"
 
 export type IModel = {
     $treenode: any // Actually Node, but that should not be exposed to the public...
 } & Object
 
-export interface IModelFactory<S, T> {
+export type IModelFactoryChecker = (snapshot: any) => boolean
+export type IModelFactoryDispatcher = (snapshot: any) => IModelFactory<any, any>
+export interface IModelFactoryConstructor<S, T>{
     (snapshot?: S, env?: Object): T & IModel
-    isModelFactory: true
+}
+
+export interface IModelFactory<S, T> extends IModelFactoryConstructor<S, T>{
     factoryName: string,
+    kind: string
+    is: IModelFactoryChecker
+    irreducible: boolean
+    dispatch: IModelFactoryDispatcher
     config: Object
+    isModelFactory: boolean
+}
+
+export function createFactoryConstructor<S, T>(
+    name: string,
+    nodeClass: NodeConstructor,
+    config,
+    instanceCreator: () => any
+): IModelFactoryConstructor<S, T>{
+    let factory = extend(
+        action(name, function(snapshot?: any, environment?: Object) {
+            const instance = instanceCreator()
+            const adm = new nodeClass(instance, environment, factory, config)
+            if (arguments.length > 0)
+                adm.applySnapshot(snapshot)
+            return instance
+        }),
+        {
+            config
+        }
+    )
+    return factory
 }
 
 export function createFactory<S, T>(
     name: string,
-    nodeClass: NodeConstructor,
-    configuration,
-    instanceCreator: () => any
+    kind: string,
+    is: IModelFactoryChecker,
+    dispatch: IModelFactoryDispatcher,
+    constructor: IModelFactoryConstructor<S, T>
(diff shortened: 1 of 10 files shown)


src/core/factories.ts
```diff
@@ -1,36 +1,69 @@
...
+    kind: string
+    is: IModelFactoryChecker
+    irreducible: boolean
+    dispatch: IModelFactoryDispatcher
     config: Object
+    isModelFactory: boolean
 }
 
-export function createFactory<S, T>(
+export function createFactoryConstructor<S, T>(
```
mattiamanzati: I had to split this because we need to create 2 different types of factories. The "basic" ones which also attach administration objects, the other just attach the type system.

mattiamanzati:
**Question:** How should we deal when trying to restore a snapshot and the instance changed its type to another of the possible union types? e.g. Box changes to Square and vice-versa.

Possible solution:
- Throw when applying directly to the union type.
- If applied to a map, array or object, discard the children and create a new one.

src/types/array-node.ts
```diff
@@ -100,15 +100,23 @@ export class ArrayNode extends Node {
...
+    let factory = createFactory(
         "array-factory",
-        ArrayNode,
-        {
-            subType: subtype,
-            isArrayFactory: true
-        } as IArrayFactoryConfig,
-        () => observable.shallowArray()
+        "array",
+        snapshot => (Array.isArray(snapshot) && snapshot.every(item => subtype.is(item))),
```
mweststrate: @mattiamanzati so you want to be able to do `unionOf(arrayOf(X), arrayOf(Y))`? But that still cannot pick the correct type for an empty snapshot I would guess, so maybe the type system should not be _that_ flexible, and only allow unions of model types?

src/types/object-node.ts
```diff
@@ -188,15 +188,27 @@ export class ObjectNode extends Node {
...
-        },
-        () => observable.shallowObject({})
+    let name = typeof arg1 === "string" ? arg1 : "unnamed-object-factory"
+    let config: Object = typeof arg1 === "string" ? arg2 : arg1
+    let primitiveKeys = Object.keys(config).filter(key => isPrimitive(config[key]))
+
+    let factory = createFactory(
+        name,
+        "object",
+        snapshot => typeof snapshot === 'object' && Object.keys(snapshot).length === primitiveKeys.length && primitiveKeys.every(key => key in snapshot),
```
mweststrate: Note that a snapshot currently a snapshot is allowed to be a subset of the target object, leaving the other fields on their initial value

mweststrate: I think the subtypes of each key/ value should also be checked to be fully correct? E.g. there could be a `unionOf({ x: 3 }, { x: "hello")`

mweststrate:
For the last question: I currently a hard contents replace is done for maps and arrays (I have a commit open that fixes this for maps), and instances are reused for sub-objects. We should add a typecheck for that indeed, and not reuse the instance if the type changes

mattiamanzati:
@mwestrate Nice! This will align the behaviour of the existing node types!

mattiamanzati:
@mweststrate IMHO the answer is yes. The identity of an object is it's position. Action is also an example of that, you can't pass models as parameters because they are not serializable, and they could be dropped. The source of truth are the path and the root node, am I right? Also on connecting side we should prefer refering to the root node and a path instead of a reference to the leaf node.
That being said we have 2 types of entities in the game. Instance-creator types and modifying types. I am not too happy about that :/

src/types/array-node.ts
mattiamanzati: @mweststrate Yes, I'd like to support that as well. The default type problem occurs also with simple unions of structs. So my guess is to throw if there is no guessed type. You can change easily that by providing at first a dispatch function (to manually detect the type), or in near future use withDefault(unionOf(arrayOf(X), arrayOf(Y)), arrayOf(X)) to default to an empty array of X.

src/types/object-node.ts
mattiamanzati: @mweststrate Yeah, I was defaulting to strict factories. I think I should reverse this check and default to true and return false if a primitive or subtype is invalid.

mweststrate:
Yeah for the action params I already have an idea: we could just (de)serialize the path, and just need a structure to make clear that the param is a path :)

mattiamanzati:
@mweststrate Yeah, like mongodb or document databases do! Something like {$ref: "/path}! Love it!

PR #32, opened 2017-01-16
Type System Improvements and fixes

mattiamanzati:
Replaces #21

### DIFF
--- a/src/core/node.ts
+++ b/src/core/node.ts
@@ -98,7 +98,7 @@ export class Node {
     }
 
     public applySnapshot(snapshot) {
-        invariant(this.type.is(snapshot), "Not assignable from " + snapshot)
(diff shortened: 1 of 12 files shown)


src/core/node.ts
```diff
@@ -151,15 +151,24 @@ export class Node {
...
+            // TODO: reason about this... Mh... :/
+            // we are adding a node with no parent (first insert in the tree)
+            if(node.parent === null){
+                node.setParent(this, subpath)
+                return child
+            }
+
             // TODO: fail here or implicitly clone?
-            child = getNode(child).snapshot
+            child = node.snapshot
```
mweststrate: I think we should throw here, implicit cloning might be to subtle, especially since we don't clone anymore for complete trees

Issue #19, opened 2017-01-02
Idea: Use TypeScript 2.1 to avoid wrong typings to make createFactory typings work.


mattiamanzati:
The following snippet won't work right now, because arrayOf returns IObservableArray (createArrayFactory returns instead a factory) and same for mapOf.

I think that was made to make arrayOf and mapOf work with createFactory.
```javascript
    const Row = createFactory({
        article_id: 0
    })

    const Collection = arrayOf(Row)

    const coll = Collection()
```

I think that we could use something like this to avoid this problem. What do you think @mweststrate ?

```typescript
export type NodeFactory<T> = (state?: T) => T
export type GraphType<T> = NodeFactory<T> | T

function structOf<T>(def: {[P in keyof T]: GraphType<T[P]>}): NodeFactory<T>{}
function arrayOf<T>(def: GraphType<T>): NodeFactory<T[]>{}
```

Issue #16, opened 2016-12-28
Idea: allow complex values to be initialized with a default value


mweststrate:
Example:

```
const Todo = createFactory({
   author: {
       name: "Michel"
   }
})
```

(Ok, bad example). 

`author` would be a fresh object for each new `Todo`.

mattiamanzati:
You mean something like this?

```js
const Author = createFactory({
  name: "Mattia"
})

const Todo = createFactory({
  author: withDefault(Author, {name: "Michel"})
})
```

mweststrate:
Yep, much better :)

tomitrescak:
Was this removed from 0.6.1 ? I cannot find a way how to use complex values in store. withDefault does not exists and I'm lost how can following be achieved:

```
const childModel = types.model('childModel', { prop: 1 });

const myModel = types.model('myModel', {
  child: childModel.create(),
});
```

PR #35, opened 2017-01-18
feat: add default type

mattiamanzati:
As of #16 

`withDefault(arrayOf(Row) as any, [{name: 'test'}])`

Notice that as any is required due to #19


mweststrate:
No further questions your honor.

PR #36, opened 2017-01-18
feat: constant and maybe types

mattiamanzati:
Builds on top of #35

### DIFF
--- a/src/types/constant.ts
+++ b/src/types/constant.ts
@@ -1,8 +1,8 @@
 import {isFactory, IFactory} from "../core/factories"
 import {invariant, fail, isPrimitive} from "../utils"
-import {PrimitiveType} from "./primitive"
+import {Type} from "../core/types"
 
-export class Constant extends PrimitiveType {
+export class Constant extends Type {
     readonly value: any
 
     constructor(value: any) {
@@ -10,12 +10,17 @@ export class Constant extends PrimitiveType {
         this.value = value
     }
 
+    create(value) {
+        invariant(isPrimitive(value), `Not a primitive: '${value}'`)
+        return value
+    }
+
     describe(){
-        return this.value
+        return JSON.stringify(this.value)
     }
 
     is(value) {
-        return value === this.value && super.is(value)
+        return value === this.value && isPrimitive(value)
     }
 
 }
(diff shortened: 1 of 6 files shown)


src/types/constant.ts
```diff
@@ -0,0 +1,26 @@
...
+export class Constant extends PrimitiveType {
+    readonly value: any
+
+    constructor(value: any) {
+        super("" + value)
+        this.value = value
+    }
+
+    describe(){
+        return this.value
```
mweststrate: return this.name? (stringified)

src/types/constant.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    is(value) {
+        return value === this.value && super.is(value)
```
mweststrate: what does the `super.is` add?

src/types/constant.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+}
+
+export function createConstantFactory(value: any): any {
+    invariant(isPrimitive(value), 'Constant types can be built only on top of primitives')
```
mweststrate: Future work could be to support (plain) complex types and deep Freeze?

src/types/maybe.ts
```diff
@@ -0,0 +1,8 @@
+import {createUnionFactory} from "./union"
+import {createConstantFactory} from "./constant"
+import {createDefaultValueFactory} from "./with-default"
+import {isFactory, IFactory} from "../core/factories"
+
+export function createMaybeFactory(type: IFactory<any, any>): IFactory<any, any>{
+    return createDefaultValueFactory(createUnionFactory(createConstantFactory(null), type), null)
```
mweststrate: Singleton for null factory?

src/types/constant.ts
mattiamanzati: I inteded to do it in a separate type named "forzen" or "serializable", the problem with complex object is if the equality check should be done on reference or deep equals.

src/types/constant.ts
mattiamanzati: Just the isPrimitive check, we can remove the extends Primitive if you want

src/types/constant.ts
mattiamanzati: This is done to make typescript-like describes. TypeScript itself shows "a" | "b" | "c", so maybe it should be `JSON.stringify(this.value)`

src/types/maybe.ts
mattiamanzati: Makes sense!

mweststrate:
Great!

PR #37, opened 2017-01-24
fix: minor correction to type definition of extend utility

steve8708:
Was browsing through the source and happened to find this :)


mweststrate:
Woah you have a sharp eye :)

PR #67, opened 2017-03-30
Type system improvements, fixes #66

### LINKED ISSUE #66: Additional fields
Hey, guys!
Thanks a lot for this great lib 👍 

Can you clarify one question for me?
I have code:
```typescript
const Item = createFactory({
  id: types.number,
  name: types.string,
});

const SomeScheme = createFactory({
  items: types.array(Item),
});

const someScheme = SomeScheme(blabla);

const response = await this.fetch('/api/');
someScheme.items.replace(response.data);
```
Sheme from API:
```json
[
  {
    "id": 2,
    "name": "Name"
  }
]
```
Everything is ok.

But then at the API backend guys adding new non-breaking (not for us) field.
For example:
```json
[
  {
    "id": 2,
    "name": "Name",
    "description": "Text"
  }
]
```

And we getting error and stack trace.

How can I avoid/handle this case?

Any solution for ignoring `additional` fields for the scheme?

mweststrate:
This PR ensures a few things:

* Correct primitive type is inferred from the default value
* If a default value is present, it can be omitted in snapshots
* `null` and `undefined` are not acceptable as model attributes (should specify `maybe`)
* Superfluous attributes in snapshots are ignored (see #66)

### DIFF
--- a/src/core/mst-node-administration.ts
+++ b/src/core/mst-node-administration.ts
@@ -214,10 +214,8 @@ export class MSTAdminisration {
             else
                 current = current!.getChildMST(pathParts[i])
             if (current === null) {
-                if (failIfResolveFails) {
-                    debugger
+                if (failIfResolveFails)
                     return fail(`Could not resolve '${pathParts[i]}' in '${joinJsonPath(pathParts.slice(0, i - 1))}', path of the patch does not resolve`)
-                }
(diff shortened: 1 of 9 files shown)


mweststrate:
Merged, feel free to leave comments nonetheless

Issue #48, opened 2017-02-21
Print warning if MobX is not in strict mode?


mweststrate:
see also: https://github.com/mobxjs/mobx-state-tree/issues/47

mweststrate:
Not needed, see #68

Issue #53, opened 2017-03-08
it should not be possible to change state outside an action


mweststrate:
question: does that also hold for applySnapshot etc?

mweststrate:
update relevant tests in `test/action.ts`

mattiamanzati:
question: does that also hold for applySnapshot etc?
@mweststrate IMHO it should'nt, applySnapshot is implicitely an action

mweststrate:
See #68 for a proposal, using `protect`

The goal to achieve is that actions are transactional; e.g. replaying actions brings us best to the same state. To ensure that if a tree is protected, not even `applySnapshot` can update it (unless of course applySnapshot is called from inside an action of the model)

mweststrate:
Fixed in 0.2.2

PR #68, opened 2017-04-03
Introduced `protect(tree)`, to defined a tree against action-less modifications

mweststrate:


mweststrate:
Fixes #48, #53

RainerAtSpirit:
What about the other way around e.g `unprotect(tree)` ?
mobx is a peer dependency of mobx-state-tree, so setting `useStrict(true)` somewhere won't allow any longer modifying models directly.

PR #77, opened 2017-04-07
WIP: Cleanup

mweststrate:


mattiamanzati:
I would also replace t.identifier() with either t.identifier or t.identifier(t.string) in this release ;)

PR #93, opened 2017-04-23
fix MSTAdminisration typo

RainerAtSpirit:


mweststrate:
Thanks for noticing! PR was quite conflicted to redid the same name in my IDE

PR #111, opened 2017-05-04
box example fix

robinfehr:
as far as i know the api might change again but here a quick pull request for the guys who want to try out the box example.


mweststrate:
Thanks!

Issue #74, opened 2017-04-06
Introduce forward types to support circular types


mweststrate:
```javascript
import {getParent, types} from 'mobx-state-tree'

const Store = types.define((self) => types.model({
   todos: types.array(Todo), // without types.define, would be `null` at this point
   subStores: types.array(self) // for recursive types
})

type IStore = typeof Store.Type // already known at compile time

const Todo = types.model({
  title: "",
  getParent(): IStore {
     return getParent(this) as IStore
  }
})

Store.finalizeType() // make sure the lambda is actually called
```

mweststrate:
cc: @mattiamanzati what do you think? It's like #52 but generalized

mweststrate:
Beste name? `forward | define | declare`

mattiamanzati:
@mweststrate Why not let generic types accept a lambda function that resolves to a type? GraphQL does something similar if I remember correclty

```javascript
import {getParent, types} from 'mobx-state-tree'

const Store = types.model({
   todos: types.array(() => Todo), // lambda will be called at create() time
   subStores: types.array(self) // for recursive types
})

type IStore = typeof Store.Type // already known at compile time

const Todo = types.model({
  title: "",
  getParent(): IStore {
     return getParent(this) as IStore
  }
})
```

mweststrate:
Solved through #113

PR #113, opened 2017-05-05
Feat: Late type, expects a type that gets defined later.

mattiamanzati:
#74


src/index.ts
```diff
@@ -5,6 +5,9 @@ import "./types/complex-types/complex-type"
 export {
     types
 } from "./types"
+export {
+    IType
```
mweststrate: just `IType` on line 6 should work as well

src/types/utility-types/recursive.ts
```diff
@@ -1,28 +0,0 @@
-import { IType, Type } from "../type"
```
mweststrate: I justed wanted to ask you about that :). Seems you are reading my mind...

src/index.ts
mattiamanzati: Unfortunately "./types" does not exports IType, so should be the same as requiring it here :)

src/types/utility-types/recursive.ts
mattiamanzati: Yeah, only reason to keep that would be to provide a better describe() in circulars, but that's not bat at all as it works now.

PR #121, opened 2017-05-10
Various changes to internal

mattiamanzati:
- [X] Internal name renames
- [X] Array, Map and literals does not have a default value anymore

### DIFF
--- a/src/types/complex-types/array.ts
+++ b/src/types/complex-types/array.ts
@@ -9,7 +9,7 @@ import {
 import { addHiddenFinalProp, identity, nothing } from "../../utils"
 import { IType, IComplexType, isType } from "../type"
 import { ComplexType } from "./complex-type"
-import { createDefaultValueFactory } from "../utility-types/with-default"
+import { optional } from "../utility-types/optional"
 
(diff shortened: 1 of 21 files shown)


src/types/complex-types/array.ts
```diff
@@ -149,7 +148,7 @@ export class ArrayType<S, T> extends ComplexType<S[], IObservableArray<T>> {
     }
 
     getDefaultSnapshot() {
-        return []
+        return undefined
```
mweststrate: I think the only use case for `getDefaultSnapshot`  is `arrayFactory.create(/* no args*/)`, not skipping values from snapshots, so I don't think this change is needed.

src/types/complex-types/map.ts
```diff
@@ -188,7 +187,7 @@ export class MapType<S, T> extends ComplexType<{[key: string]: S}, IExtendedObse
     }
 
     getDefaultSnapshot() {
-        return {}
+        return undefined
```
mweststrate: idem

mweststrate:
Not entirely sure about not defaulting `.create()` either. But beyond that, looking good so merged anyway :)

PR #129, opened 2017-05-16
Export also simple type to allow for easier typing of frozen types

tomitrescak:
Example

```js
export const NumberFieldModel = types.model("Field", {
  value: types.number,
  validation: types.frozen as ISimpleType<RegExp>
});
```

instead of

```js
export const NumberFieldModel = types.model("Field", {
  value: types.number,
  validation: types.frozen as IType<RegExp, RegExp>
});
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,7 +8,7 @@ import "./types/complex-types/complex-type"
 // unescapeJsonPath
 
 export {
-    types, IType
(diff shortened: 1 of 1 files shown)


mattiamanzati:
@tomitrescak Thanks! Can you please create a new PR with both changes? :)

PR #132, opened 2017-05-17
argument names within onAction middlewares.

robinfehr:
within middlewares added to onAction, the arguments now are an object
that includes pairs of argument name and argument value.

### DIFF
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -3,7 +3,7 @@ import { action as mobxAction, isObservable } from "mobx"
 export type ISerializedActionCall = {
     name: string;
     path?: string;
-    args?: any[];
+    args?: any;
 }
 
 export type IRawActionCall = {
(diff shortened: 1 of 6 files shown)


mattiamanzati:
@mweststrate Are we sure that we want that? The way to get argument names is kinda hacky :/

robinfehr:
@mattiamanzati i agree i don't like how reflection is implemented in js either. the decision is about what is more intuitive and how are the users going to use it. I for example wanted to tag my actions somehow so that i can say: this action is recordable for the time machine, and others are not. You cant do that using an object like:

action:
```
move({ dx, dy, optimistic, recordable = true }) {...}
```
middleware:
```
const args = action.args[0]
const { optimistic, recordable } = args || {};
// recordable is always undefined here (default param problem) - tagging not possible like that
```
component:
```
store.move({20, 20, true}) // you'd need to add the recordable here
```



whereas with named arguments you could do that:

action:
```
move(dx, dy, optimistic, recordable) {...}
```
middleware:
```
const { optimistic } = args;
const recordable = Object.keys(action.args).includes('recordable'); // taggable
```
component:
```
store.move({20, 20, true})
```

i agree the solution is not perfect. i'm very open to another way to tag an action - just couldn't find another way until now.

robinfehr:
I'd also like to add that it would be nice to have the possibility to add optional parameters like i tried above with the optimistic argument - which is undefined if not provided and can luckily be handled like a bool in javascript. currently im thinking of something like action.dispatch(box.move(20,20), { optimistic: false })

PR #139, opened 2017-05-22
upgrade MST & mobx version in redux-todomvc example

mondaychen:
mobx 2.6.0 does not work with mobx-state-tree 0.5.0

You will see error message like this when you run `npm start`:

image

This PR will fix this issue


mattiamanzati:
Thanks for your PR! :D

PR #140, opened 2017-05-22
fix nextState bounds checking in box example

mondaychen:
It's too late when it's `states.length`...

image


mattiamanzati:
Thanks for your PR! :D

PR #147, opened 2017-05-24
fixed rendering issue for "clone" in api overview

alonbardavid:
The markdown for "clone" method in the api overview is malformed due to not escaping the pipe symbol.


mweststrate:
Thanks!

Issue #134, opened 2017-05-19
Refereced objects that use types.union died when applying snapshot


abruzzihraig:
Hi guys,
I have to say this is an amazing project. It always gives me all kinds of surprises when I use it to build a project with state management and observable features. I really appreciate your talent work.

But today I got a weird issue and it took me a few hours to dig until now. I think the document is not perfect by now, and I still don't really understand parts of them. So I want to describe the confusing issue I met to you, then you may help me to know if it is indeed an issue from the mobx-state-tree or just a wrong way I wrote. Thanks.

The scene has a little bit complicated so I try my best to simplify it for you.

##### Stage 1
At the beginning, there was a simple store that held two properties like below:

```
const BuildingStore = types.model('BuildingStore', {
    objects: types.array(types.union(Text, Post)),
    selection: types.reference(types.union(Post, Text))
}, {
    setSelection(selection: IUIObject) {
        if (this.selection === selection) return;

        if (this.selection && isAlive(this.selection)) {
            if (this.selection.setEditablility) {
                this.selection.setEditablility(false);
            }
        }

        this.selection = selection;
})
```

As you can see, I used `types.union` there in both of `types.array` and `types.reference`. I think it makes sense although I cannot find any example for how to use `types.union` in `types.reference`.

During this stage, I got an error when I was trying to `applySnapshot` to `BuildingStore`. The error message showed it's an ambiguous snapshot. So I just put a dispatcher function for the property `objects`:

##### Stage 2

```
const BuildingStore = types.model('BuildingStore', {
    objects: types.array(types.union((snapshot: any) => {
        if (snapshot.type === 'Post') return Post;
        if (snapshot.type === 'Text') return Text;
    }, Post, Text)),
    selection: types.reference(types.union(Post, Text))
})
```

At this time, the ambiguous error disappeared. But when I did something like undo to `applySnapshot` to `BuildingStore` again, I found I cannot get the computed properties from the `selection` anymore. At the mean time, if I do an operation that triggers the method `setSelection`, I will get a new error like below:

```
utils.js:14 Uncaught Error: [mobx-state-tree] Failed to assign a value to a reference; the value should already be part of the same model tree
    at Object.invariant (utils.js:14)
    at Reference.setNewValue (reference.js:64)
    at BuildingStore.set (reference-property.js:40)
    at BuildingStore.setSelection (buildingStore.ts:22)
    at executeAction (mobx.js:663)
    at BuildingStore.res (mobx.js:654)
[...]

PR #151, opened 2017-05-26
fix(union): Union.create() shouldn't ignore extra arguments

abruzzihraig:
Children in type Union could be any type, currently means it could be either `SimpleType` or `ComplexType` so the rest arguments such as `environment, parent, subpath` shouldn't be ignored in method `Union.create()`.

I would have tried to declare the method `create` into the type `IComplexType` such as:
```
export interface IComplexType<S, T> extends IType<S, T & ISnapshottable<S> & IMSTNode> {
    create(snapshot?: S, environment?: any, parent?: MSTAdministration, subpath?: string): T
}
```
Then it might be a type `IComplexType<any, any>|IType<any, any>` or `IComplexType<any, any>|ISimpleType<any, any>` could be used in the Class Union.

But I'm confused with the current structure in MST:
```
IComplexType extends IType
ISimpleType extends IType
Type implements IType
```
So far so good, but why `ComplexType extends Type` rather than `ComplexType extends Type implements IComplexType`? It doesn't make sense to me.

Besides, if I use the union type like `IComplexType<any, any>|IType<any, any>` or `IComplexType<any, any>|ISimpleType<any, any>`, I wouldn't know how to distinguish them in runtime so that I cannot give it a specific assertion.

That's why finally I just changed the parameter declarations in `IType.create()`.

### DIFF
--- a/src/types/type.ts
+++ b/src/types/type.ts
@@ -18,7 +18,7 @@ export interface IType<S, T> {
     flags: TypeFlags
     is(thing: any): thing is S | T
     validate(thing: any, context: IContext): IValidationResult
-    create(snapshot?: S, environment?: any): T
+    create(snapshot?: S, environment?: any, parent?: MSTAdministration | null, subpath?: string): T
     isType: boolean
     describe(): string
     Type: T
@@ -67,3 +67,4 @@ export abstract class Type<S, T> implements IType<S, T> {
 import { fail } from "../utils"
 import { IMSTNode } from "../core/mst-node"
 import { IContext, IValidationResult } from "./type-checker"
+import { MSTAdministration } from "../core/mst-node-administration"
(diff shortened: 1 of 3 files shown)


abruzzihraig:
Link to #134

mattiamanzati:
Hi! Thanks for your PR!
The other parameters doesnt appear in IType because we dont want them to be visible, can you please revert only the IType declaration? :)

abruzzihraig:
@mattiamanzati if I revert the change in IType, the following lines have to be changed by an assertion `as any`:

https://github.com/mobxjs/mobx-state-tree/pull/151/commits#diff-7536e74658295b48b68949f5100a9989R37 
https://github.com/mobxjs/mobx-state-tree/pull/151/commits#diff-7536e74658295b48b68949f5100a9989R45

Is that ok for your code foundation? Using assertion `as any` here seems is not a good way... But currently, we cannot use IComplexType or something else to restrict it.

mattiamanzati:
Its fine for now, soon when all types will be complex by default, that will no more be an issue :)

abruzzihraig:
@mattiamanzati Pls check it again, thanks.

mattiamanzati:
Now its fine! Thanks for your hard work! :)

PR #152, opened 2017-05-27
Box all the things

mweststrate:
First class boxes, references, etc. Should address irregularities related to references, add support for references inside `maybe`, `map`, `array` etc


robinfehr:
awesome - reached that problem just now :)

mattiamanzati:
Possible enhancement: Since typecheck() is recursive, we can remove all the typechecks from instantiate and assume that snapshot is a valid snapshot. Maybe typecheck should be performed only in create() and applySnapshot()

src/core/node.ts
```diff
@@ -11,6 +11,7 @@ export class Node  {
     // optimization: these fields make MST memory expensive for primitives. Most can be initialized lazily, or with EMPTY_ARRAY on prototype
     readonly nodeId = ++nextNodeId
     readonly type: IType<any, any>
```
mweststrate: maybe rename type -> actualType to make it more clear?

src/types/complex-types/map.ts
```diff
@@ -42,6 +42,7 @@ export class MapType<S, T> extends ComplexType<{[key: string]: S}, IExtendedObse
     createNewInstance() {
         // const identifierAttr = getIdentifierAttribute(this.subType)
         const map = observable.shallowMap()
+        extras.getAdministration(map).dehancer = unbox
```
mweststrate: dehancer is the worst name, looking for a better alternative :-P

mattiamanzati: boxer and unboxer? 👖

PR #176, opened 2017-06-09
implement computed as type

mattiamanzati:
Only 1 test failing for computed setter. Should we support that? Maybe we should let the computed be read only and not able to set? What do you think @mweststrate?


mweststrate:
I think this PR makes computed properties enumerable on objects. Could that be a problem? Maybe we should have `enumerable` or `snapshottable` as type property. That avoids the need for omit-from-snapshot, and would nicely support `types.localState` as well. 

I think computed setters should be supported as well. In MobX that is a quite commonly used feature

mattiamanzati:
Uh, using a snapshottable:boolean property would also allow future optimization ad avoidable checks if the reconcile should create a new node or keep the existing one.
The problem with supporting setters @mweststrate is that ATM isValidSnapshot() only accepts value and context, so it does'nt know if you are trying to create a new node or apply to an existing one. Maybe we should also pass currentNode: Node | null as first property, so if currentNode is null, we check the snapshot to be undefined, and if currentNode already exists, we allow any value. How does it sounds?

mattiamanzati:
Uhm, downsides with the "snapshottable" flag, types like union if have childs both snapshottable and non snapshottable, it's unable to predict if is snapshottable :/

mattiamanzati:
Will be reimplemented later over the new branch.

PR #177, opened 2017-06-09
applySnapshot fails on types.maybe(complexType)

geowarin:
Hello, thanks for your work, we love MST!

We stumbled upon a bug when applying snapshots to a `types.maybe`.
When we first set the value to a legal object, then set it to null, it fails type checking:

```
[Error: [mobx-state-tree] Error while converting `null` to `AnonymousModel`:
    value `null` is not assignable to type: `AnonymousModel`, expected an instance of `AnonymousModel` or a snapshot like `{ value: string }` instead.]
```

We noticed that the bug was fixed on the `box-all-the-things` branch and we're looking forward to testing it.

We thought we could contribute a test to ensure no regression happens in further versions.

### DIFF
--- a/test/object.ts
+++ b/test/object.ts
@@ -92,6 +92,20 @@ test("it should apply snapshots", (t) => {
     t.deepEqual(getSnapshot(doc), {to: 'universe'})
 })
 
+test("it should apply and accept null value for types.maybe(complexType)", (t) => {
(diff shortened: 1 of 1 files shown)


mattiamanzati:
Thanks a lot for your contribution! Your test will help prevent it will happen again :)

PR #186, opened 2017-06-12
Add applyPatch test case for reference array

cpunion:
See https://github.com/mobxjs/mobx-state-tree/issues/185

### DIFF
--- a/test/reference.ts
+++ b/test/reference.ts
@@ -1,5 +1,5 @@
 import { reaction, autorun } from "mobx"
-import { types, getSnapshot, applySnapshot, unprotect, resolvePath, detach, resolveIdentifier, getRoot } from "../src"
+import { types, getSnapshot, applySnapshot, onPatch, applyPatch, unprotect, resolvePath, detach, resolveIdentifier, getRoot } from "../src"
 import { test } from "ava"
 
 test("it should support prefixed paths in maps", t => {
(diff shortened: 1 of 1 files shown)


test/reference.ts
```diff
@@ -610,4 +610,64 @@ test("References in recursive structures", t => {
...
+        hovers: types.array(types.reference(Item))
+    }, {
+        addObject(item) {
+            this.objects.put(item)
+        },
+        addHover(item) {
+            this.hovers.push(item)
+        },
+        removeHover(item) {
+            this.hovers = this.hovers.filter(i => i.id === item.id)
```
mweststrate: note: same as `this.hovers.remove(item)`

cpunion: For `removeHover`, should be `!==`, or use `remove`

mattiamanzati:
Ok, I think I got where the problem is, basically after the item is removed from the array, dehanceValues is called and calls getValue over the reference, that has'nt got yet an identifier cache.
We could have 2 solutions; one is to remove somehow that call, and the other one is make by default the actual type of reference(T) as T | null where null is when the reference is'nt resolvable due to being removed from the tree, etc..

cpunion:
Found another place, just called `applySnapshot`, the stack trace is:
```
The error was thrown at: 
    at ReferenceType.getValue
    at Node.getValue
    at Node.unbox
    at Array.map
    at ObservableArrayAdministration.dehanceValues
    at ObservableArrayAdministration.spliceWithArray
    at ObservableArray.replace
    at Node.pseudoAction
    at ArrayType.applySnapshot
    at executeAction
    at ArrayType.res
    at Node.applySnapshot
    at ArrayType.ComplexType.reconcile
    at ValueProperty.willChange
    at ObjectType.willChange
    at Array.<anonymous>
    at interceptChange
    at setPropertyValue
    at Project.set
    at ValueProperty.deserialize
    at Node.pseudoAction
    at ObjectType.applySnapshot
    at executeAction
    at ObjectType.res
    at Node.applySnapshot
    at applySnapshot
```

mattiamanzati:
@cpunion Thanks again for reporting and for your test! Can you please update this branch from master, and add an example fail case for applySnapshot as separate test? :)

Thanks again! :)

cpunion:
Sure, I will try. :)

cpunion:
Added a test case, and rebased on master.

mattiamanzati:
Nice! Seems to be fixed! Thanks for your help @cpunion! :)

cpunion:
Yeah! Thanks!

Please cut a release today, I was blocked. :)

PR #190, opened 2017-06-13
isMST is replaced by isStateTreeNode

ssured:


mattiamanzati:
You're right! Thanks for your PR! :D

PR #194, opened 2017-06-15
Add cross references in changelog

ssured:
I like to learn the rationale behind changes in code. Referencing the original issues & PR's helps to learn the why and how of the code


mattiamanzati:
Definetely agree with that, will keep in mind in next changes ;)

Issue #179, opened 2017-06-10
Switch from webpack to rollup for bundling


mweststrate:
Does rollup support peer deps properly? Rollup will make the build smaller and faster, support tree shaking and more debugger friendly code

jbrantly:
Out of curiosity, doesn't webpack2 support tree shaking? Is there a difference between the tree shaking support in webpack2 vs rollup?

mweststrate:
Afaik webpack 2 was quite limited in this regard. Anyway rollup bundles are much cleaner; they inline all the sources, rather than webpack that tries to preserve the original source modules, which is totally not interesting for a library like this, and hampers the opportunities for minification

mweststrate:
@rossipedia, interested to roll up this one as well? No legacy default exports here ;-)

rossipedia:
sure, why not :)

rossipedia:
62.7KB with Webpack
44.5KB with Rollup

I think just for the heck of it I'm going to try webpack@3 with scope hoisting, see what that ends up looking like

rossipedia:
Roughly the same with webpack 3 and scope hoisting. Rollup definitely the clear winner re: size.

mweststrate:
Released in 0.8.0. The standard export is now rolled-up as well :)

mweststrate:
And added es6 module for the fancyness of it :

PR #196, opened 2017-06-15
convert from webpack to rollup for bundling

rossipedia:
As requested in #179


mattiamanzati:
Thanks @rossipedia for your awesome work! :)

PR #202, opened 2017-06-19
Add todomvc example

mondaychen:
I know we already have a redux-todomvc example, but it will be nice to have a pure MST style todomvc as well. It can demonstrate how modularized your app can be (no more "findById", Todo Model speaks for itself, etc.), and also shows off useful APIs like `destroy`.


mweststrate:
Thanks for this PR!

examples/todomvc/src/models/todos.js
```diff
@@ -0,0 +1,48 @@
+import { types, destroy } from 'mobx-state-tree'
+
+const Todo = types.model(
+  {
+    text: 'Learn Redux',
```
mweststrate: Default `text` to `''` ?

examples/todomvc/src/models/todos.js
```diff
@@ -0,0 +1,48 @@
...
+    completed: false,
```
mweststrate: note: source code doesn't seem pretty printed. If you are using VSCode it should do so automatically, otherwise see the package json scripts on how to do that

examples/todomvc/src/models/todos.js
```diff
@@ -0,0 +1,48 @@
...
+    id: 0
```
mweststrate: `types.identifier(types.number)` will be better here, it will result in better reconciliation when applying snapshots etc

examples/todomvc/src/components/Footer.js
```diff
@@ -0,0 +1,72 @@
...
+import classnames from 'classnames'
+import { SHOW_ALL, SHOW_COMPLETED, SHOW_ACTIVE } from '../constants/TodoFilters'
+
+const FILTER_TITLES = {
+  [SHOW_ALL]: 'All',
+  [SHOW_ACTIVE]: 'Active',
+  [SHOW_COMPLETED]: 'Completed'
+}
+
+export default class Footer extends Component {
```
mweststrate: Suggestion: I think Footer and the call to `Footer` will be simpler if you just pass in the store, and mark it `observer`?

examples/todomvc/src/models/todos.js
mweststrate: or `Learn MST` ;-)

examples/todomvc/src/models/todos.js
mondaychen: fixed in 2d2b2b7 let me know if it looks good now

examples/todomvc/src/models/todos.js
mondaychen: is `types.string` better here? bb89f97

examples/todomvc/src/models/todos.js
mondaychen: bb89f97

examples/todomvc/src/components/Footer.js
mondaychen: Great idea! Done in 0d1989b.

I'm thinking -- maybe it's also a good idea to move "filter" to store as well. It will make everything even easier.

examples/todomvc/src/models/todos.js
mondaychen: I noticed that my vscode is switching between single quote and double quote... maybe we should define it in the settings.json as well?

mondaychen:
Hi @mweststrate 

I've set `prettier.singleQuote` to `false` in 991bde0bbaa9b4dba05bf7cd24b7148495853744 and move `filter` to the store in 88cfe5643b43749a98daed86381b09e228de1f58

Let me know if the new commits look good to you

Thx for reviewing

mweststrate:
Nice work @mondaychen! Cool demonstration of optional, union and literal with the filters.

examples/todomvc/src/models/todos.js
mweststrate: Yep that was missing indeed. Thanks!

PR #220, opened 2017-06-30
Allow Mobx observables as snapshots

alessioscalici:
Case scenario:
The user's state tree contains an array (types.array). The user wants to store a snapshot in a normal mobx observable. The array in the snapshot becomes an ObservableArray, thus it won't be matched anymore by Array.isArray (As described in https://mobx.js.org/best/pitfalls.html).
If the user tries to apply this snapshot with _applySnapshot_, the type check will fail.

This PR is to allow applying observable snapshots

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist
 test-lib
 coverage
 .nyc_output
+.idea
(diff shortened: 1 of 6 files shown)


alessioscalici:
@mattiamanzati could you give a look to this? :)

mattiamanzati:
@alessioscalici Yeah, it seems a fair point. As observable arrays behaves and extends plain arrays it should be fine! If also @mweststrate is fine with that, we can merge! :)

mweststrate:
I concur. We could check the code base for other changes where an Array.isArray check is done and apply the same

alessioscalici:
@mattiamanzati sure, I will at my first!

mattiamanzati:
Thanks! And don't forget to add the change to the CHANGELOG.md with your name! :)

alessioscalici:
@mattiamanzati I added a test, and refactored the "isArray" check as an Utils function, and updated usages.
I think this will fix https://github.com/mobxjs/mobx-state-tree/issues/221

mattiamanzati:
Thanks for your work and your time! Going to merge it! :)

PR #222, opened 2017-06-30
Increase validation information for OptionalValue

caesarsol:
Type validation errors at runtime were masked if the type was wrapped in an `OptionalValue`.

This was very annoying for large snapshots, since the specific model key that caused the error was difficult to find.

With this PR the model keys are explicitly listed in the error.

```js
const { types: t } = require('mobx-state-tree')
const Aaa = t.optional(t.array(t.model('Aaa', {a: t.number})), [])
Aaa.create([{a:1}, {a:2}, {a:3}, {}, {a:5}, {}])
```

Before:

    Error: [mobx-state-tree] Error while converting `[{"a":1},{"a":2},{"a":3},{},{"a":5},{}]` to `Aaa[]`:
    snapshot `[{"a":1},{"a":2},{"a":3},{},{"a":5},{}]` is not assignable to type: `Aaa[]`, expected an instance of `Aaa[]` or a snapshot like `{ a: number }[]?` instead.

After:

    Error: [mobx-state-tree] Error while converting `[{"a":1},{"a":2},{"a":3},{},{"a":5},{}]` to `Aaa[]`:
    at path "/3/a" value `undefined` is not assignable to type: `number`.
    at path "/5/a" value `undefined` is not assignable to type: `number`.

@mweststrate I'd like to work on console errors verbosity, I have in mind a couple of PRs on improving the message and details, are you open to it?

### DIFF
--- a/src/types/utility-types/optional.ts
+++ b/src/types/utility-types/optional.ts
@@ -1,6 +1,12 @@
 import { Type, IType } from "../type"
 import { TypeFlags } from "../type-flags"
-import { IContext, IValidationResult, typecheck, typeCheckSuccess, typeCheckFailure } from "../type-checker"
+import {
+    IContext,
+    IValidationResult,
+    typecheck,
+    typeCheckSuccess,
+    typeCheckFailure
+} from "../type-checker"
 import { isStateTreeNode, getStateTreeNode, Node } from "../../core"
 
 export type IFunctionReturn<T> = () => T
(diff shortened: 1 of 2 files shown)


caesarsol:
You're welcome @mattiamanzati! 
What about the question on the validation error messages?

mattiamanzati:
We're open for sure, maybe first open an issue describing the changes if they are lot or breaking :/

PR #226, opened 2017-06-30
Guards against called isFrozen on primitives.

### LINKED ISSUE #225: Object.isFrozen can only be called on Objects (android/react-native)
Quick sanity check here.  

I think there's an issue with calling `Object.isFrozen()` on a `number` on Android (`react-native@0.45.1` on Android 6 on an `LG-G4`).

`Object.isFrozen can only be called on Objects.`

image

The [line that's dying](https://github.com/mobxjs/mobx-state-tree/blob/48567d7ccdd3894da348f1a07e6a80f6b230bf1a/src/utils.ts#L94) is in the `deepFreeze()` function.  I'm feeding it an object with a property that is a number.

I noticed in [`freeze()`](https://github.com/mobxjs/mobx-state-tree/blob/48567d7ccdd3894da348f1a07e6a80f6b230bf1a/src/utils.ts#L86) you're checking for a [`isPrimitive`](https://github.com/mobxjs/mobx-state-tree/blob/48567d7ccdd3894da348f1a07e6a80f6b230bf1a/src/utils.ts#L73) and not freezing anyway, so I tried adding this extra guard:

```
           if (!isPrimitive(value[propKey]) && ... 
```

... and it fixes it.

I guess on older JS environments, you can't ask if you're frozen on a primitive... or something.

Is there a `types.frozen` alternative that was based on the honour system?  😆 

I can PR this if you think it would be useful.

skellock:
Fixes #225

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -91,7 +91,10 @@ export function deepFreeze<T>(value: T): T {
 
     if (isPlainObject(value)) {
         Object.keys(value).forEach(propKey => {
-            if (!Object.isFrozen((value as any)[propKey])) {
+            if (
+                !isPrimitive((value as any)[propKey]) &&
+                !Object.isFrozen((value as any)[propKey])
+            ) {
                 deepFreeze((value as any)[propKey])
             }
         })


mattiamanzati:
Thanks again for your time!! Will release a 0.9.1 later to distribute all this fixes :)

Issue #107, opened 2017-05-02
applySnaphshot / applyPatch should show up in action middleware as well?


mweststrate:
Possible as predefined actions, e.g. `$applySnapshot` / `$applyPatch` ? Or, alternatively, applyPatch / applySnapshot should not be allowed if a tree is protected. But should rather be made part of an action

mattiamanzati:
Already merged, will be out with next release!

PR #227, opened 2017-07-01
Killed pseudo actions, applySnapshot and applyPatch will now properly…

mweststrate:
… emit an action as well. See #107

### DIFF
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -120,10 +120,10 @@ export function applyAction(target: IStateTreeNode, action: ISerializedActionCal
     const node = getStateTreeNode(resolvedTarget)
 
     // Reserved functions
-    if (action.name === "$APPLY_PATCHES") {
+    if (action.name === "@APPLY_PATCHES") {
         return applyPatch.call(null, resolvedTarget, action.args![0])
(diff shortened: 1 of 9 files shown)


mattiamanzati:
Uhm, just an opinion, may be a lazy idea: Why not @applyPatches and @applySnapshot?
@ is not a valid character for starting a function name, so may be better to intend that this is a system action :)
On the other hand, why not just use the real names (without @)? So if I re-run the actions sequentially that will ensure correct replayability :)

src/core/action.ts
```diff
@@ -122,6 +118,15 @@ export function applyAction(target: IStateTreeNode, action: ISerializedActionCal
     const resolvedTarget = tryResolve(target, action.path || "")
     if (!resolvedTarget) return fail(`Invalid action path: ${action.path || ""}`)
     const node = getStateTreeNode(resolvedTarget)
+
+    // Reserved functions
+    if (action.name === "$APPLY_PATCHES") {
```
mattiamanzati: Why not use @APPLY_PATCHES?
$ is a valid function name character, @ is not :)

Issue #251, opened 2017-07-17
Optimize reconciler by using `===` when applying snapshots


mweststrate:
If a node produced a snapshot, and later on the same snapshot is applied, but the node didn't change in the mean time, it doesn't need to reconcile. This could greatly improve features like snapshot based time travelling if this is checked first

mattiamanzati:
Merged, will be out with next release! :)

PR #253, opened 2017-07-18
avoid apply snapshot if it is the same as the current one

mattiamanzati:
Should address #251 
I think that checks in reconciler and applySnapshot should be enough, and they are actually pretty cheap. Am I missing something else?

### DIFF
--- a/src/types/type.ts
+++ b/src/types/type.ts
@@ -89,7 +89,9 @@ export abstract class ComplexType<S, T> implements IType<S, T> {
     }
 
     reconcile(current: Node, newValue: any): Node {
-        const { parent, subpath } = current
+        if (current.snapshot === newValue)
+            // newValue is the current snapshot of the node, noop
+            return current
         if (isStateTreeNode(newValue) && getStateTreeNode(newValue) === current)
             // the current node is the same as the new one
             return current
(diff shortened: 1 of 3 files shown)


src/types/type.ts
```diff
@@ -89,7 +89,9 @@ export abstract class ComplexType<S, T> implements IType<S, T> {
     }
 
     reconcile(current: Node, newValue: any): Node {
-        const { parent, subpath } = current
+        if (current.snapshot === newValue)
```
mweststrate: Q: is this check performed _before_ apply preProcessedSnapshot hooks? I think it should be before, for symmetry purposes, as the stored snapshot is the postProcessedSnapshot. Not sure about it, but there might be problems if this optimization is applied after preProcessing (although I can't think of one, maybe you can :-P).

Further LGTM :)

mattiamanzati: A: This is always called before preProcess, as this check is performed in node "applySnapshot", and after that the type applySnapshot kicks in, so should be everything fine  :) I can't think of issues too, maybe will just end up skipping always the perf improvement ;)

mweststrate:
Check, feel free to merge :)

Op di 18 jul. 2017 om 23:43 schreef Mattia Manzati <notifications@github.com

mattiamanzati:
Thanks for your time! :)

mweststrate:
Thanks for the PR 😁

Op di 18 jul. 2017 23:52 schreef Mattia Manzati <notifications@github.com>:

Issue #232, opened 2017-07-04
Introduce `types.enum`


mweststrate:
It is too convenient and commont to not have :)

signature: `types.enum(name?, [options], defaultOption?`)`, e.g. 

`const stopLight = types.enum("StopLight", ["Red", "Orange", "Green"], "Red")`

Simple sugar around `types.optional(types.union(...options.map(types.literal)), defaultValue)`

mattiamanzati:
Uhm, I'd avoid the defaultOption and leave that to the types.optional, I think it's more consistent with other types, what do you think? :)

mweststrate:
Fair point

xaviergonz:
@mweststrate @mattiamanzati Is there a way to use this new construct easily with the new Typescript 2.4 string enums? I mean, an alternative for the second argument that probably also supports passing the enum type itself (as long as it is not a const enum of course) instead of every single enum value

xaviergonz:
Ok, apparently for string enums it is easy, just pass as second argument Object.values(StopLight), however for old (number) enums it might be not so easy, since for example for this:

export enum ClipType { Intersection, Union, Difference, Xor }

it generates the following code:
```js
var ClipType;
(function (ClipType) {
    ClipType[ClipType["Intersection"] = 0] = "Intersection";
    ClipType[ClipType["Union"] = 1] = "Union";
    ClipType[ClipType["Difference"] = 2] = "Difference";
    ClipType[ClipType["Xor"] = 3] = "Xor";
})(ClipType = exports.ClipType || (exports.ClipType = {}));
```

which means that both the number themselves as well as "Intersection" and so on are keys and values of the object... (though the numbers are actual numbers in the values and strings in the keys)

mweststrate:
See PR #263

xaviergonz:
I see that you used literal("" + option), is there any reason to not supports enum of any kind (e.g. numbers by passing an array of numbers)

xaviergonz:
only either strings or numbers would be cool too

mweststrate:
Numbers seems such an edge case (since there is no translation table
involved) that it is fine imho to use a union of literals in that case

Op za 22 jul. 2017 22:46 schreef xaviergonz <notifications@github.com>:

mweststrate:
Released as 0.9.4

PR #263, opened 2017-07-22
Introduced types.enumeration, implements #232

mweststrate:


src/types/utility-types/enumeration.ts
```diff
@@ -0,0 +1,29 @@
...
+ * })
+ * ```
+ *
+ * @export
+ * @alias types.enumeration
+ * @param {string} name descriptive name of the enumeration (optional)
+ * @param {string[]} options possible values this enumeration can have
+ * @returns {ISimpleType<string>}
+ */
+export function enumeration(name: string | string[], options?: string[]): ISimpleType<string> {
```
xaviergonz: could be <T> - T[] - ISimpleType<T>
or if only strings and numbers are supported then string | number

Issue #266, opened 2017-07-25
Introduce `async`


mweststrate:
Like: https://github.com/mobxjs/mobx-utils#asyncaction

mweststrate:
Done (called `process` now). See #273 for future extensions

Issue #260, opened 2017-07-21
onAction not trigger when invoke from parent store


AlexanderChen1989:
```javascript
const Store = t.model({
  owner: t.model({}, {
    hello() {
      console.log('Hello')
    }
  }),
}, {
  doIt() {
    this.owner.hello()
  }
})

const store = Store.create({owner: {}})

onAction(store.owner, action => {
  console.log(action)
})

window.store = store 
```

![image](https://user-images.githubusercontent.com/2957936/28448902-4cff4860-6e0e-11e7-8529-1361e7f8ad0c.png)

mattiamanzati:
By design only the outermost action will trigger action events, and refers to action owned by that model you subscribe to, so Thats intended. @mweststrate may be more precise than me here :)

AlexanderChen1989:
I know, thanks

mweststrate:
Actually I think we have run here in a potential issue with the middle ware model, e.g. given this:

```javascript
const Todo = types.model({
  done: types.boolean,
}, {
  toggle() {
    this.done = !this.done
  }
})

const Store = types.model({
  todos: types.array(Todo)
}, {
  toggleAll() {
    this.todos.forEach(t => t.toggle())
  }
}
```

If you now add some middleware to a store instance, everything will work as expected. However, if you add middleware to a specific todo (e.g. `addMiddleWare(store.todos[0], someHandler)`), it will trigger on `store.todos[0].toggle()`, but not(!) on `toggleAll`

I think there are 3 potential fixes for this:
1 . make middleware smarter, and somehow detect that the middleware on todo should be triggered as well
2. forbid attaching middleware on a non-root
3. trigger middleware always, also if an action is called from another action

---

For 3: the reason why indirectly called actions don't call middleware is that it ensures replayability, indirectly called actions don't have to be replayed, as they should be the deterministic result of calling the 'wrapping' action. 

Note that this is currently not entirely true; async handlers need to be actions currently, but are in fact also the result of calling another action. Still tinkering on how this can be handled more elegantly

For 2: this is the easiest and conceptually most simple solution, but makes MST less composable

For 1: This can be achieved by keeping track of whether listeners have already fired.

So I think either 1 or 2 should be implemented to fix this properly

olivoil:
Hi! Is there a way to receive actions called from another action in a middleware at all?

[...]

PR #267, opened 2017-07-25
Introduce async actions, implements #266

mweststrate:
* [x] impl
* [x] middleware support
* [x] tests
* [x] docs
* [x] middleware docs
* [x] Test & address #260 
* [x] typings
* [x] expose explicit `async` for better typings (and in future to combine with `optimistic`)


test/async.ts
```diff
@@ -0,0 +1,66 @@
...
+    })
+}
+
+test.cb("can handle async actions", t => {
+    const Todo = types.model(
+        {
+            title: "get coffee"
+        },
+        {
+            *fetchData(kind: string) {
```
mattiamanzati: Are we sure about autowrapping with async? Are typings fine with it "T" wise? :/

mattiamanzati:
Will this PR also implement call mode: direct | indirect? If so maybe asyncId should be something like transactionid?

mweststrate:
No, I don't think so, it doesn't close the door either. But I started first with that, before doing async, but then somehow it didn't make sense in might head (and it started hurting) that middleware starts kicking off in the middle of a function run (indirect mode). Like in Express where middleware only works on the outside. Middleware will be kicked of for individual steps in async btw.

test/async.ts
mweststrate: Typings are an open issue, but beyond that is syntactically really nice. Especially because `fetchData: async(function* { ... })` is kinda ugly, but maybe that is the way forward for better typings. Really like the current syntax though :)

src/core/action.ts
```diff
@@ -8,6 +8,11 @@ export type ISerializedActionCall = {
 }
 
 export type IRawActionCall = {
+    /**
+     * AsyncId indicates whether this action is part of an `async` based action. Id is zero if it isn't.
+     */
+    asyncId: number
+    asyncMode: "none" | "start" | "yield" | "end"
```
mattiamanzati: What about use mode: invoke | start | yield | end and transaction: number (or random string)? I'd always generate an id even for plain actions, and on standard actions use invoke or return, whilst in async have a series of start, yield, throw, return?

mattiamanzati:
Ok, here's the proposal for an API that would allow either safe time-travelling, and nice and fine TS typings.

Given the following piece of code as example which is taken from a currently working app.
```typescript
{
            recoverFromStorage() {
                return promise(this)(function*() {
                    // read user id from async storage
                    const userId = yield AsyncStorage.getItem("user_id")

                    // if has one, restore it
                    yield getUserService(this).recoverFromStorage()
                    if (userId && getUserService(this).get(parseInt(userId, 10))) {
                        this.userId = parseInt(userId, 10)
                    }
                })
            }
}
```

Instead of allowing generators as members of actions, we introduce a new function "promise" which takes as only argument the context of the async action generator, and returns a function that accepts the generator and starts it immediately. The result will be Promise<T>, where T can be optionally be specified when calling promise<T>(this).

The action log of "recoverFromStorage" will be exactly as is now, except the addition of a id field.
`{name: "recoverFromStorage", path: "/auth", args: [], id: 1, mode: "return"}`
This way, middleware can handle it with no problem as happens currently with sync actions.

After that, the promise execution will eventually start, and a spawn event will be emitted. (could be avoided, but its nice to have).
`{name: "recoverFromStorage", path: "/auth", args: [], id: 2, mode: "spawn", spawnedBy: 1}` (spawnedBy could eventually be stored in args)
The spawnedBy, name, etc.. params of the promise initialization will be read and stored to be used later when invoking promise(this). Basically promise(this) is some sort of "lock", telling that something will happen later on this.

When the action will end, a throw or return could eventually be emitted.
`{name: "recoverFromStorage", path: "/auth", args: [undefined], id: 3, mode: "return", spawnedBy: 1}`
`{name: "recoverFromStorage", path: "/auth", args: ["ERROR!"], id: 3, mode: "throw", spawnedBy: 1}`

Now we can have 2 options:

1. Serialize as top level call:
We could now treat the code inside the generator as it would be global code executed by the user. The only difference is that spawnedBy will be present in any of the emitted actions. So in the given example code the following actions will be emitted:
`{name: "recoverFromStorage", path: "/user", args: [], id: 3, mode: "return", spawnedBy: 2}`
...and eventually applyied patches without actions can be serialized as applyPatches invoke
`{name: "@APPLY_PATCHES", path: "/auth", args: [{op: "replace", path: "/userId", value: 1}], id: 4, mode: "invoke"}`
This also ensures composability, because as we can see in action #3 another spawn action event could be triggered.
This way of handling is nice because it emits nicely replayable actions, but while replaying won't actually re-run the generator. It will instead perform the state changes the generator did. We could even assume that everything we do in the generator is impure (e.g. http calls, storage reads, etc...), but will be actually converted onto pure actions/patches calls.

2. Serialize as yields happens
The other option involves serializing the value emitted/read by the generator. This will re-run the generator while replaying, but notice that it will also perform any promise code. So if we were for examply replaying some fetch/post to an endpoint it will actually perform it even if the generator will actually .next the replay log value, resulting in some issues in the database.

PR #281, opened 2017-08-01
Remove usage of eval to support usage of CSP self

pioh:
Because of the fact that it uses a new Function(""), I have to use Content Security Policy script-src: 'unsafe-eval' insted of just 'self'.
Look's like Object.defineProperty also should work.

https://stackoverflow.com/a/40918734
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name#Description


mattiamanzati:
Seems legitimate! Was already mentioned somewere if I remember correclty :)
Thanks for your hard work!

adamsanderson:
You might also want to update the comment right above the change since now it does what is says it didn't do ;)

Issue #282, opened 2017-08-01
Introduce new syntax to declare actions and volatile state


mweststrate:
[...]

Issue #279, opened 2017-07-31
Async actions don't work using TypeScript and `es5` target


praxxis:
Per https://github.com/Microsoft/TypeScript/issues/15088, TS using `"target": "es5"` doesn't brand a generator function with a proper `name` or `displayName`, which causes [`isGeneratorFunction`](https://github.com/mobxjs/mobx-state-tree/blob/master/src/utils.ts#L91-L97) to return a false negative. 

Furthermore, the MST [tests tsconfig](https://github.com/mobxjs/mobx-state-tree/blob/master/test/tsconfig.json#L5) is set to the `es2015` target, which is why this is currently passing in tests (the base tsconfig is set to `es5`).

TBH I'm not sure what a good fix for this is. Checking an action for a `next` function or some such might work, but feels wrong. There's nothing else in the TS [`__generator` polyfill ](https://github.com/Microsoft/tslib/blob/master/tslib.es6.js#L69-L95) that could be used to discover a generator function.

KatSick:
+1 . blocker for me

mweststrate:
Sorry guys, nasty oversight. Is this something that needs to be fixed urgently? #282 will address it properly, but otherwise we could introduce an `async(function* () {})` method that marks the method properly as work around. (it screws typings a bit though). Basically undoing this [commit](https://github.com/mobxjs/mobx-state-tree/commit/5469bd2f44c0c1d1c8ec7529ecf03d8036ac5003). But instead of check for `isGenerator`, explicitly marking it as generator.

KatSick:
hi @mweststrate. not urgent at all. Helper `async(function * () {})` is great workaround except for Typescript users. With `strictImplicitAny: true` all `this` inside actual generator `function * () {}` will be any..

[...]

PR #286, opened 2017-08-03
[WIP] New model syntax

mweststrate:
See #282 

Todo:
* [x] implement
* [x] verify types
* [x] update base tests
* [x] add tests for volatile state / private funcs etc
* [x] restory async
* [x] add tests for nested async
* [x] make sure async is properly bound
* [x] restore snapshot processing
* [x] example updates
* [x] document fluent api
* [x] address async generator syntax with es5 target & typings (see #279)
* [x] implement fluent type api with`actions(self => actions).props(propDeclaration).name("some name")`. Each step should produce a new type (not modifying the current one)
* [x] express compose using fluent api (or drop it?)
* [x] support old syntax or provide a codemod
* [x] restore all skipped tests
* [x] document babel setup for generators / update bookshop example
* [x] doc updates
* [x] update async-action docs
* [x] tutorial updates
* [x] document code mod
* [ ] update codesandbox examples (ASAP after merge)
* [ ] update getting started tutorial and related sandboxes (ASAP after merge)
* [ ] update mobx-state-tree deps in example package jsons (ASAP after merge)


src/types/complex-types/object.ts
```diff
@@ -98,7 +98,15 @@ export class ObjectType<S, T> extends ComplexType<S, T> implements IModelType<S,
             // todo assert plain object
             if (actions && isPlainObject(actions)) {
                 Object.keys(actions).forEach(name => {
-                    addHiddenFinalProp(self, name, createActionInvoker(self, name, actions[name]))
+                    const action = actions[name]
+                    if ((action as any).isAsyncMSTAction === true)
```
mattiamanzati: Is this only to avoid async actions emit also a normal action?

mweststrate:
Yep.

Op do 10 aug. 2017 09:13 schreef Mattia Manzati <notifications@github.com>:

Bnaya:
Shouldn't this PR be in the 1.0 milestone?

mweststrate:
Yes. I think the related issue is

Op do 10 aug. 2017 16:32 schreef Bnaya Peretz <notifications@github.com>:

src/types/complex-types/object.ts
mattiamanzati: Uhm, what about having a modifier like "silent(...action...)" instead of specifically tie it to async actions?

Bnaya:
So add it there ;) its seems so empty 
https://github.com/mobxjs/mobx-state-tree/pulls?q=is%3Aopen+is%3Apr+milestone%3A%221.0+-+initial+release%22

mattiamanzati:
I completely agree and fallen in love with these last commits <3
As said I'd only change the fact that we should emit yield events only if promises are wrapped in task(() => promise()) to ensure that no side effects are performed on the tree and only a value is returned to ensure correct time travel replay :)

src/core/async.ts
```diff
@@ -48,42 +48,56 @@ export function asyncAction(asyncAction: any): any {
     return createAsyncActionInvoker(asyncAction.name, asyncAction)
```
mattiamanzati: Uhm, can we add an optional first parameter for the name? Not sure if most generator polyfills will retain the function name.

docs/getting-started.md
```diff
@@ -178,14 +181,18 @@ const User = types.model({
...
+    function addTodo(id, name) {
+        self.todos.set(id, Todo.create({ name }))
     }
+
+    return {addTodo}
 })
 ```
 [View sample in playground](https://goo.gl/Sj17Nq)
 
+Please notice the use of "self". Self is the object being constructed when an instance of your model is created. Thanks to the self object instance actions are "this-free", allowing you to be sure that they are correclty bound.
```
praxxis: `correctly`

PR #292, opened 2017-08-07
[Docs] Fix some wrong code blocks

caesarsol:
Were in wrong markdown syntax


caesarsol:
Did not realize these docs were generated from source... :(

PR #295, opened 2017-08-10
Update README.md

homura:
make types.late example correct


mattiamanzati:
Thanks for your time and PR! :D

PR #306, opened 2017-08-17
Various Object cleanups

mattiamanzati:
- [X] deprecate ValueProperty
- [X] optimization: persist properties names in array to iterate easily
- [X] added check for views return
- [X] object will no longer initialize undefinedType under the hood
- [X] Node snapshot reaction will initialize after node is finalized
- [X] Added package manager lock files to ignore list


mweststrate:
can you remove package-lock and stick to `yarn` for now? we shouldn't be using both I think

mattiamanzati:
Done&Done! :)
I've added them to the ignore list, so they wont bother anymore in the future :)

.gitignore
```diff
@@ -5,4 +5,6 @@ dist
 test-lib
 coverage
 .nyc_output
-.idea
\ No newline at end of file
+.idea
+yarn.lock
```
mweststrate: yarn.lock should *not* be in git ignore. Assuming we pick yarn for now as primarly way to maintain the repo

src/core/node.ts
```diff
@@ -31,15 +31,19 @@ export class Node {
...
-        storedValue: any
+        initialValue: any,
+        createNewInstance: (initialValue: any) => any = identity,
+        finalizeNewInstance: (node: Node, initialValue: any) => void = noop
     ) {
         this.type = type
         this._parent = parent
         this.subpath = subpath
-        this.storedValue = storedValue
+        this.storedValue = createNewInstance(initialValue)
```
mweststrate: @mattiamanzati did you see this comment?

src/types/complex-types/object.ts
```diff
@@ -312,28 +350,26 @@ export class ObjectType<S, T> extends ComplexType<S, T> implements IModelType<S,
...
 
-    private forAllProps(fn: (o: Property) => void) {
-        // optimization: persists keys or loop more efficiently
-        Object.keys(this.parsedProperties).forEach(key => fn(this.parsedProperties[key]))
+    private forAllProps(fn: (name: string, type: IType<any, any>) => void) {
+        this.propertiesNames.forEach(key => fn(key, this.properties[key]))
     }
 
     describe() {
         // TODO: make proptypes responsible
```
mweststrate: this todo can be removed

src/core/node.ts
mattiamanzati: @mweststrate Just seen :)
Btw it does'nt matter the order because afaik and if you watch the old code, createNewInstance is called before node creation, so has not any concern about environment theorically.

Swapped anyway to make code more readable!

PR #328, opened 2017-08-23
dev-mode only checks

### LINKED ISSUE #275: Type refinement not checked on action (but checked on applySnapshot)
Hi everybody :)

## Summary
We have some weird behaviour about `identifier` checking : 
 - it works well with `applySnapshot`
 - it seems bypassed with an action.

## Context
We have this helper to declare identifiers :
```es6
export default (name: string) => types.identifier(
  types.refinement(
    'id',
    types.string, identifier => identifier.indexOf(`${name}_`) === 0,
  ),
)
```

We have this model (Author): 
```es6
import { types } from 'mobx-state-tree'
import { getIdType } from './helpers'

const name: string = 'Author'

export default types.model(
  name,
  {
    id: getIdType(name),
    name: types.string,
    image: 'https://cdn.pixabay.com/photo/2016/11/18/23/38/child-1837375_1280.png',
  },
)
```

## Error
Using an action to set our store: the `Author` identifier is not checked. 😢 
Using `applySnapshot`: the `Author` identifier is checked. 👍 

----

## With an action
image

## With `applySnapshot`
PR: https://github.com/fabienjuif/gni-mst/pull/4/files
```es6
`AnonymousModel`:
at path "/recipes/0/author/id" value `"char_1"` is not assignable to type: `identifier(id)`, expected an instance of `identifier(id)` or a snapshot like `identifier(id)` instead.
```

## With `applySnapshot` fix
PR: https://github.com/fabienjuif/gni-mst/pull/5/files
image

mattiamanzati:
- [X] introduce checks for all the top-level types definitions
- [x] introduce checks for all top-level MST utility functions
- [X] run type-checks only in dev-mode
- [X] fixed #275 

cc @mweststrate

### DIFF
--- a/README.md
+++ b/README.md
@@ -792,6 +792,10 @@ A _disposer_ is a function that cancels the effect it was created for.
 
 # Tips
 
+### Building with production environment
+
+MobX-state-tree provides a lot of dev-only checks. They checks the correctness of function calls and perform runtime type-checks over your models. It is recommended to disable them in production builds. To do so, you should use webpack's DefinePlugin to set environment as production and remove them. More information could be found in the [official webpack guides](https://webpack.js.org/plugins/environment-plugin/#usage).
+
 ### `optionals` and default value functions
 
 `types.optional` can take an optional function parameter which will be invoked each time a default value is needed. This is useful to generate timestamps, identifiers or even complex objects, for example:
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,7 @@
+# next
+
+* Added dev-only checks, typecheck will be performed only in dev-mode and top-level API-calls will be checked.
+
 # 0.10.2
 
 * Object model instances no longer share a prototype.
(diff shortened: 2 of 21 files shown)


mweststrate:
Impressive! Is the goal:

1. smaller prod builds?
2. runtime perfromance?
3. both :)?

Also, could `process.env.NODE_ENV !== "production"` be rewritten to some `const isProduction` in `utils`, or would that kill the minification?

mattiamanzati:
Main goal is DX, I discovered through user issues that the lack of call checks resulted in weird issue when the user make mistakes. This aims to make the DX better by throwing whenever something is calling not in the correct way! :)
Some checks have been moved to be dev-only, so yeah, will definetely result in smaller build for production :)
I need to re-check that, slept on it and figured out why the first time i've done that it did'nt worked out.

mattiamanzati:
Nope, tried out declaring a global __DEV__, but ends up not being removed in prod build :/

mattiamanzati:
@mweststrate Can I merge this or do you prefer to first merge your others PR and then merge this since this may break the other PRs? :)

mweststrate:
LGTM, feel free to merge. But I think the readme might give some attention how to set it up properly :) (at least for webpack builds, which is probably the most common)

mattiamanzati:
Added to the readme on the tips section, if build passes, I'll merge to master :)

PR #332, opened 2017-08-24
WIP introduce `invariant` / `getSelf`

mweststrate:


src/core/action.ts
```diff
@@ -28,6 +28,14 @@ export type IMiddleWareEvent = {
 let nextActionId = 1
 let currentActionContext: IMiddleWareEvent | null = null
 
+export function getSelf(): IStateTreeNode {
+    if (!currentActionContext || !currentActionContext.context)
```
mweststrate: initializer functions (`.views` / `.actions`) don't create an action context. How do we make them return the correct self?

mattiamanzati: What about renaming currentActionContext to currentExecutionContext and add a field that tells if it is in view, action, and in the future, task execution?

mweststrate:
Closing, I think it becomes to obscure for now and it is not really needed

PR #345, opened 2017-08-31
docs: dedupe headers

aretecode:


mweststrate:
`API.md` get's generated from the source code (see `npm run build-docs`). So this PR won't fix it. Didn't figure out yet where those empty headers are coming from though. There don't seem to be superfluous JSDoc comments in the code that are responsible for this. It might be an issue with the processing of the `@alias` tags.

PR #349, opened 2017-09-01
Fix references of enum -> enumeration

NgoKnows:
There were a few places where `enum` was being used instead of `enumeration`.


mattiamanzati:
Thanks for your effort! :)

PR #380, opened 2017-09-13
fixing incomplete types.identifier() call and default value for maps

the-noob:
` id: types.identifier,` throws
> Error: [mobx-state-tree] Functions are not supported as properties, use views instead

`const s = TodoStore.create({ todos: [] })` throws

> [mobx-state-tree] Error while converting `{"todos":[]}` to `AnonymousModel`:
> at path "/todos" snapshot `[]` is not assignable to type: `map<string, Todo>` (Value is not a plain object), expected an instance of `map<string, Todo>` or a snapshot like `Map<string, { id: identifier(string); task: string }>` instead.


mattiamanzati:
Good catch! Thanks a lot!

PR #384, opened 2017-09-15
[ready to merge] New array reconciler algo

mattiamanzati:
This is a new algo for the array reconciler.
Instead of performing 3/4 for loops through all the nodes, it will instead perform all the steps of reconciling in one pass.
The algo will try to do its best job and also supports reconciling based on snapshot reference identity.
If you pick the snapshot of an array element without identifier and move it along the array, this algo will identify it by checking if oldNode.snapshot === newValue.
This means that if you move around snapshot of array elements with nodes without identifiers, it will reuse the existing node for that snapshot.
This could open for new ways of managing arrays of !isStateTreeNode() elements that should be less memory-stressfull as it would reuse existing nodes instead of dropping and recreating them.

### DIFF
--- a/src/types/complex-types/array.ts
+++ b/src/types/complex-types/array.ts
@@ -18,7 +18,7 @@ import {
     isStateTreeNode,
     IStateTreeNode
 } from "../../core"
-import { addHiddenFinalProp, fail, isMutable, isArray } from "../../utils"
+import { addHiddenFinalProp, fail, isMutable, isArray, isPlainObject } from "../../utils"
 import { ComplexType, IComplexType, IType } from "../type"
 import { TypeFlags, isType } from "../type-flags"
 import {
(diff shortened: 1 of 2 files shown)


src/types/complex-types/array.ts
```diff
@@ -250,85 +250,116 @@ function reconcileArrayChildren<T>(
...
+        } else if (areSame(oldNode, newValue)) {
+            oldNodes[i] = valueAsNode(childType, parent, "" + newPaths[i], newValue, oldNode)
+            // nothing to do, try to reorder
         } else {
-            // create a fresh MST node
-            res[index] = childType.instantiate(parent, subPath, undefined, newValue)
+            oldMatch = undefined
+
+            // find a possible candidate to reuse
+            for (let j = i; j < oldNodes.length; j++) {
```
mweststrate: Note that this makes the reconciliation basically O(N)^2. For that reason there was in the original implementation a map of old instances build first (so performance is ~ O(N)). I think that could still be done in the new setup? Create a map `identifier => { target, stillAvailable }` ?

mattiamanzati: Uhm, I am not good at this kind of things so I may be totally wrong :)
Was'nt also the old one O(N)^2 because findReconciliationCandidate would loop through all unused array items with an identifier (for lett attr in oldNodesBy etc...), with decreasing amount of loops each step the old nodes get used up.
I think here's basically the same thing, findReconciliationCandidate is mostly equal to node loop + areSame?

PR #387, opened 2017-09-15
Allow setting a parent node if the child and parent envs are the same.

### LINKED ISSUE #354: Do not throw if given environment is the same as the target one
Is there any reason why I can't do something like this if the two environments passed on create are the same ones?

```js
const env = whatever;
const parent = Parent.create({}, env);
const child = Child.create({}, env);
parent.children.set('asdf', child); // exception
parent.children.set('asdf', getSnapshot(child)); // ugly workaround, but works
```

I'd understand the error if the two envs were not the same, but when they are?

xaviergonz:
Fixes #354


mattiamanzati:
@xaviergonz Thanks a lot for this PR!
Could you please add also a test for not throwing if env is === same? :)

xaviergonz:
@mattiamanzati there you go!

xaviergonz:
anything else needed?

mattiamanzati:
I think just the approval of @mweststrate :)

mweststrate:
@xaviergonz approved! Thanks :)

PR #388, opened 2017-09-16
fix changelog import path

aretecode:


mattiamanzati:
Good catch!
Thanks for your time! ;)

PR #390, opened 2017-09-17
Fix types.map example errors

rchanou:
The types.map example appears to be incomplete with numerous errors. I updated it so it actually works correctly.


rchanou:
Ah, I just found the documentation on the `put` method for identifiers. I still think the example is broken, but I'll close this for now and see if I can come up with something better using the `put` method.

PR #391, opened 2017-09-18
fix types.map example

rchanou:


mattiamanzati:
Uhm, I think that there are 2 thing to consider before merging this:
1. Old example was fine, it was just missing unprotect(store) to enter unprotected mode
2. API.md is generated from jsdoc of the API, so any change needs to be mabe there!
Thanks for your work, feel free to relocate the changes so I can merge :)

PR #394, opened 2017-09-19
fix: types.map example

rchanou:
OK, I edited the `map.ts` source file and rebuilt `API.md` from that.


mattiamanzati:
Thanks a lot for your time! :)

Issue #395, opened 2017-09-21
wrong value in inversePatch for nested map


sqxieshuai:
`MST version is 1.0.0`

hi,

 i use the test code like below:

```
var mst = require('mobx-state-tree');

var types = mst.types;
var onPatch = mst.onPatch;
var getSnapshot = mst.getSnapshot;

var Store = types.model({
  value: types.map(types.map(types.map(types.number))),   // KEY POINT
}).actions(self => ({
  remove() {
    self.value.delete('k1');   // KEY POINT
  },
}))

var store = Store.create({
  value: {
    k1: {
      k2: {
        k3: 3,
      },
    },
  },
});

console.log('store before remove: ', JSON.stringify(store.toJSON()), JSON.stringify(getSnapshot(store)));

onPatch(store, (patch, inversePatch) => {
  console.log('store onPatch ', 'patch: ', JSON.stringify(patch), 'inversePatch: ', JSON.stringify(inversePatch));
});

store.remove();

console.log('store after remove: ', JSON.stringify(store.toJSON()));

```

and after run it, the output is like:

```
store before remove:  {"value":{"k1":{"k2":{"k3":3}}}} {"value":{"k1":{"k2":{"k3":3}}}}
store onPatch  patch:  {"op":"remove","path":"/value/k1"} inversePatch:  {"op":"add","path":"/value/k1","value":{"":{"k3":3}}}
store after remove:  {"value":{}}
```

the `value` in inversePatch i expected is :

```
{"k2":{"k3":3}}
```

but in fact the value of the `value` and the expected value is not the same.

is the way I use the MST map type incorrectly? 
or 
it's a bug in MST?

mweststrate:
That looks like a bug indeed. Would you mind creating a PR with the above
example as unit test?

Op do 21 sep. 2017 om 08:24 schreef shuai.xie <notifications@github.com>:

PR #396, opened 2017-09-21
fix: reversepatch key is empty when deleted an item from a nested map

0neSe7en:
Fixes: #395

### DIFF
--- a/test/map.ts
+++ b/test/map.ts
@@ -282,3 +282,21 @@ test("it should not throw when removing a non existing item from a map", t => {
         t.is(store.something(), false)
     })
 })
+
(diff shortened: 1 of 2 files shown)


test/map.ts
```diff
@@ -282,3 +282,21 @@ test("it should not throw when removing a non existing item from a map", t => {
         t.is(store.something(), false)
     })
 })
+
+test.cb("it should get map keys from reversePatch when deleted an item from a nested map", t => {
```
mattiamanzati: Should'nt be just test()? :)

0neSe7en: done

PR #405, opened 2017-09-25
fix wrong un/escaping of json patches and add test for it

mattiamanzati:
Added a test for it, so we won't type it wrong again :)


mweststrate:
Awesome!

Op ma 25 sep. 2017 om 17:03 schreef Mattia Manzati <notifications@github.com

PR #407, opened 2017-09-28
getEnv(this) -> getEnv(self)

aretecode:


mattiamanzati:
Whops! Good catch! Thanks for your work!

Issue #399, opened 2017-09-22
`process` is a very bad function name


s-panferov:
Hello. Recently I faced with a bug in my code that I can't make ReactHotLoader work because it uses `process.env.NODE_ENV` global variable to check an environment. But if you have `import { process } from 'mobx-state-tree'` require, webpack fails to inline `process.env.NODE_ENV` variable.

This is very bad because a lot of people may face it and it's really hard to understand what is going on.

tinchoz49:
I think the same as you. What I'm doing to solve the problem is using 
`import { process as co } from 'mobx-state-tree'`

aretecode:
`fromGeneratorToAsync`

nickbreaton:
What if it was renamed to `fromGeneratorToAsync` as suggested by @aretecode, but it was implied by using a generator function in actions.
```js
types
  .model({
    name: 'default'
  })
  .actions(self => ({
    changeName(name) {
      self.name = name
    },
    *fetchRandomName() { 
      const response = yield fetch('.....')
      self.name = response.data
    }
  }))
```

instead of

```js
  types
    .model({
      name: 'default'
    })
    .actions(self => ({
      changeName(name) {
        self.name = name
      },
      /* this would replace 'process', and would still be valid, but verbose */
      fetchRandomName: fromGeneratorToAsync(function* () { 
        const response = yield fetch('.....')
        self.name = response.data
      })
    }))
```

This can be detected by checking the constructor name:

```js
(function a() {}).constructor.name  // => "Function"
(function* b() {}).constructor.name  // => "GeneratorFunction"
```

mattiamanzati:
@nickbreaton In the past it worked like that, but was changed due to problems imlied by transpilation. See #279.

nickbreaton:
@mattiamanzati Sorry about that I checked Babel, but not TS. 

I'm just spit balling, but what about something like this?
```js
types
.model({ name: '' })
.actions(self => ({
  changeName(name) {
    self.name = name
  }
}))
.processes(self => ({
  *fetchRandomName() {
    // ...
  }
}))
```

It would allow people to avoid importing `process`.

mattiamanzati:
Could potentially be a solution, I would like more something like

.compose(processes(self =>({
....
})

so that eventually other community-provided model enhancer could be provided.

nickbreaton:
[...]

PR #410, opened 2017-09-28
Proposal: Introduce `.processes`

nickbreaton:
### What?

A way to define a process, without having to wrap each action with `process()`.

```ts
import { types } from 'mobx-state-tree'

const User = types
  .model({
    name: types.string
  })
  .processes(self => ({
    *getData() {
      const response = yield fetch('...')
      const data = yield response.json()
      self.name = data.name
    }
  }))
```

### Why?
- Avoids having to import `process`, see #399 
- Allows for syntax that has been used in the past, without the TypeScript bugs, see #279 
- Introduces no breaking changes
- Cleans up verbose `process` definitions

  <table>
    <thead>
      <tr>
        <th>Proposal</th>
        <th>Current</th>
      </tr>
    </thead>
    <tr>
      <td>
  <pre lang="ts">.processes(self => ({
    *getData() {
      const response = yield fetch('...')
      self.data = yield response.json()
    }
  }))</pre>
        </td>
        <td>
  <pre lang="ts">.actions(self => ({
    getData: process(function* () {
      const response = yield fetch('...')
      self.data = yield response.json()
    })
  }))</pre>
        </td>
      </tr>
    </table>
- Typings represent a process as a _method_ and not a _property_
  <table>
    <thead>
      <tr>
        <th>Proposal</th>
        <th>Current</th>
      </tr>
    </thead>
    <tr>
      <td>
  screen shot 2017-09-28 at 6 52 37 pm
        </td>
        <td>
  screen shot 2017-09-28 at 6 53 24 pm
        </td>
      </tr>
  </table>

### Why not... 

...implement a more generic `.compose` function as proposed [here](https://github.com/mobxjs/mobx-state-tree/issues/399#issuecomment-331657527), and allow the community to implement `processes`?
- I think it is important to provide a solid way to handle async actions as almost every application needs it. MST is supposed to be opinionated, and I believe it should have an opinion about the way async actions work.
- I like the idea of `.compose`, but think it is out of the scope of this PR.

### Notes

- If `process()` is aliased or renamed, #399, then this `.processes` should be as well.

__Edit:__ Thanks for taking the time to read. If this seems like a terrible idea to you, please shoot it down. As always, thanks for maintaining this awesome project.

### DIFF
--- a/src/types/complex-types/object.ts
+++ b/src/types/complex-types/object.ts
@@ -39,6 +39,7 @@ import {
 } from "../type-checker"
 import { getPrimitiveFactoryFromValue, undefinedType } from "../primitives"
 import { optional } from "../utility-types/optional"
+import { createProcessSpawner } from "../../core/process"
 
 const PRE_PROCESS_SNAPSHOT = "preProcessSnapshot"
 
@@ -174,6 +175,16 @@ export class ObjectType<S, T> extends ComplexType<S, T> implements IModelType<S,
             return self
         }
         return this.extend({ initializers: [actionInitializer] })
+    }
+
+    processes<A extends { [name: string]: Function }>(fn: (self: T) => A): IModelType<S, T & A> {
+        return this.actions((self: T) => {
+            const actions = fn(self)
+            Object.keys(actions).forEach((name: string) => {
+                actions[name] = createProcessSpawner(name, actions[name])
+            })
+            return actions
+        })
     }
 
     named(name: string): IModelType<S, T> {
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks for the PR. But I think it doesn't address all use cases yet:

1. how to create process that are private (not exposed as api)
2. how to share closure (internal state) with actions?

These are downsides simply renaming `process` to `flow` doesn't have. But  yes, not needing to wrap with `process` is a nice upside ;-). But I think the current proposals sets processes a bit too much apart from the normal actions

nickbreaton:
Both are excellent points I neglected to think about. I may have gotten a little too excited to help contribute. 😅

mweststrate:
No problem, and if you are interested in creating a PR for process -> flow
(backward compatible with deprecated message), a PR is welcome! Most work
is probably updating docs

Op vr 29 sep. 2017 om 14:58 schreef Nick Breaton <notifications@github.com>:

Issue #377, opened 2017-09-11
Parent's view items are not undefined in child's hooks


KevinOrtman:
After updating to v0.11.0 of MST, references to a parent's view items are not resolved (null) in the child's afterCreate() and afterAttach() hooks.  

This did work in v0.9.5. 

This can be reproduced with the code snippet below (modified bookstore example) and this sandbox:  
https://codesandbox.io/s/6yzz6kj7jw
(NOTE: you will need to open the codesandbox preview pane in it's own window for the alert to be visible)

```
.actions(self => ({        
        afterCreate() {
          alert(`After create - Fetch is undefined: ${typeof self.shop.fetch === "undefined"}`);
          alert(`After create - isLoading is undefined: ${typeof self.shop.isLoading === "undefined"}`);
        },
        afterAttach() {
            alert(`After attach - Fetch is undefined: ${typeof self.shop.fetch === "undefined"}`);
            alert(`After attach - isLoading is undefined: ${typeof self.shop.isLoading === "undefined"}`);
        },
```

mweststrate:
@KevinOrtman thanks for reporting! Are you interested in creating a PR for a unit test to reproduce this behavior?

KevinOrtman:
Certainly.

maxgallo:
I run into the same bug.
I've got a workaround, even if you're not going to like it.

**Not working**
```
.actions(self => ({
    afterCreate() {
        console.log(getParent(self).anyViewProperty); // this is undefined
    }
}))
```

**Working**
```
.actions(self => ({
    afterCreate() {
        setTimeout(() => {
            console.log(getParent(self).anyViewProperty); // this has a value
        }, 0);
    }
}))
```

mweststrate:
Will be fixed in next MST release; where `afterAttach` can be used to make assumptions about parent. `afterCreate` hook will run depth-first, `afterAttach` breadth first

PR #413, opened 2017-09-30
adding parent-properties test

KevinOrtman:
Adding unit test for issue #377

### DIFF
--- a/test/parent-properties.ts
+++ b/test/parent-properties.ts
@@ -1,0 +1,45 @@
+import { types, getEnv, getParent } from "../src"
+import { test } from "ava"
+
+const ChildModel = types
+    .model("Child", {
+        parentPropertyIsNullAfterCreate: false,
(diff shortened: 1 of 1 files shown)


mweststrate:
Sorry for the huge delay! Investigated a bit more, an made a follow up MR based in your tests. For afterCreate it is correct that you cannot access the views of a parent yet, as children are created first recursively, before finalizing the types with actions etc. 

However, to support your case, the MR changes the semantics of `afterAttach`, which will now first on run on the parent, and then on the children, only after the parent has finalized. This should result in the correct behavior. See also the updated tests: https://github.com/mobxjs/mobx-state-tree/blob/6cae9468128fc89a9049ee0e83d4eee12b05ae61/packages/mobx-state-tree/test/parent-properties.ts#L37

KevinOrtman:
That works.   Thank you, Michel.

PR #424, opened 2017-10-05
Updated TypeScript typings on enumeration

danielduwaer:
This is a raise of the upper bound on the number of options that the current TypeScript typings of the enumeration supports.

Now it will support up to, and including, 100 options while preserving the typing information. The current upper bound of 20 is too low for a lot of practical applications.


mattiamanzati:
Uhm, 100 seems a lot, is'nt 40 / 50 enough?

danielduwaer:
@mattiamanzati Agreed, 50 seems like a reasonable amount of options. I have updated my PR.

mattiamanzati:
@danielduwaer Thanks a lot for your time! :D

PR #435, opened 2017-10-08
Rename `process` to `flow`, deprecate `process`

nickbreaton:
In #399, an issue was raised about `process` being a poor choice for a function name, as it collides with the Node.js `process` global. This is especially problematic because it interferes with build tools such as webpack when trying to strip development code from production builds.

The solution determined was to rename `process`. The word `flow` was chosen as an acceptable synonym.

This PR contains the following:
  - Deprecation of `process`, without breaking API compatibility. 
  - Renaming `process` to `flow`.
  - All internal uses of `process` changed to `flow`.
  - A `deprecated` helper function to display warning messages in the console. 

I need some additional guidance on:
  - [x] __Test Coverage__: `process` was renamed to `flow` in all test cases. This lowers test coverage, as the process file is no longer being tested. `flow` wasn't simply aliased to `process` because the action types in the function body needed to be renamed as well (`"process_*"` => `"flow_*"`). I'm not sure if we should duplicate the test cases, (seems messy) or ignore `process.ts` from the coverage report (doesn't seem safe). Please let me know what you think and I will address it.

  - [x] __Example MST version__: ~~Since the examples were changed, the version number will have to be updated, but the next version of MST is not yet published to NPM. How should I go about doing this?~~

  
      Addressed by #446, ~~will update this PR when that is complete.~~


mweststrate:
Hmm the middleware names is nasty indeed. Maybe we should just make this a breaking change? (I don't think there are much manually written middlewares yet anyway that use the event names)

nickbreaton:
I'm all for that. Then we can simply alias process to the flow functions. Just let me know and I'll take care of it.

nickbreaton:
I can also add a note of that in the deprecation warning of `process()` so people are aware.

mweststrate:
You agree @mattiamanzati? Make it a breaking change (cause of the event names), and make `process` a simple delegator with deprecation message?

mweststrate:
@nickbreaton sorry we just moved today to lerna, so the project structure has changed quite a bit. I hope just rebasing will work, otherwise conflicts are easy to solve by copying your conflicted files over `packages/mobx-react-state-tree` folder. Apologies for the inconvenience and thanks a lot for the work so far!

nickbreaton:
@mweststrate No worries at all! It solves my issue with the example version anyway.

mattiamanzati:
@nickbreaton @mweststrate Yeah I am totally fine with it, there are not many great usages of middlewares yet, so yeah, perfectly fine :)
Sorry for the inconvenient of moving to lerna :(

nickbreaton:
@mattiamanzati 

Don't worry about lerna being inconvenient! However, I do have a question regarding the examples. Is there anyway to have the `mobx-state-tree` version in the examples' `package.json` point to the current version in this repository? 

I ask because with this API change, I want to update the example code to reflect it, but the `mobx-state-tree` version in the `package.json` is `^0.10.1`. Clearly this PR is not in that version, so if someone tried to run that code it would not work. 

Maybe the `mobx-state-tree` version could point to the actual directory? 

```json5
// example package.json

{
  "dependencies": {
    "mobx-state-tree": "file:../mobx-state-tree"
  }
}
```

If its not that big of a deal, I won't worry about it. 

I just figure for future PRs that touch example code and source code, the example version will always point to the actual code containing the changes.

mattiamanzati:
@nickbreaton
I moved to lerna because lerna allows exaclty that feature!
Clone the repo, npm install && npm run bootstrap and the folder of mobx-state-tree will be automatically symlinked inside the node_modules of the examples :)

mattiamanzati:
@nickbreaton Ping me when you feel like this is ready to merge, and remember to update CHANGELOG.md :)

nickbreaton:
This is the deprecation message presented when `process` is being used:

image

nickbreaton:
@mattiamanzati I have added information to the changelog, and I believe this is now ready to merge.

mattiamanzati:
Thanks a lot for your awesome work!
It's merged now! :D

nickbreaton:
No problem. I look forward to contributing in the future!

PR #446, opened 2017-10-10
Lerna MonoRepo

mattiamanzati:
As discussed with @mweststrate , this PR initializes a lerna monorepo.
This way we can easily make examples depend on mobx-state-tree itself, and move eventually middlewares to a separate package that will be compiled and bundled.


mattiamanzati:
@mweststrate Is it fine for you to have all the middleware as separate packages like mst-middleware-redux, etc... or do you prefer a single package, mst-middlewares with all the middlewares?

mattiamanzati:
Pro of the new setup: building the library also builds the examples and run test over them, so if any change breaks an example, it needs to be updated to make CI happy.

mattiamanzati:
@mweststrate Should be done now, feel free to merge whenever you want :)

skellock:
2 more and I can build a card game.

mattiamanzati:
@skellock LOL
Not sure why all of those errors where'nt running on my machine :/ Some of them appeared only in CI

mattiamanzati:
Going to merge right now, cross your finger everyone! :)

Issue #445, opened 2017-10-10
ES6 in published middleware


paul-veevers:
I ran into an issue when using the new TimeTraveller middleware. Distributed `npm` code hasn't been fully transpiled to `es5`. It says [here](https://github.com/mobxjs/mobx-state-tree/blob/master/middleware/README.md#contributing) "The middlewares must be written in valid ES5. ES imports / exports are allowed."

I wouldn't have noticed, but uglifyjs complains :)

Why not run babel on each file with module transpilation disabled in the build step?

I'm happy to do a PR with whatever you think is best.

mweststrate:
Feel free to PR! Preferably we do keep the ES imports / exports (I think it
depends on the uglify version whether that is supported or not), as without
ES modules mobx-state-tree cannot be tree-shaken anymore when using these
middlewares

Op di 10 okt. 2017 om 12:30 schreef Paul Veevers <notifications@github.com>:

mattiamanzati:
I've encountered this issue while working on #446 , and I was thinking about moving middlewares to a separate package, have them written in TS and fully transpiled. Is fine for anyone?

mweststrate:
+1

Op di 10 okt. 2017 om 19:13 schreef Mattia Manzati <notifications@github.com

mattiamanzati:
Ok, just done in #446 , now create-react-app is also able to build and does'nt complain about missing full-es5 transpilation. This should fix your use-case @paul-veevers !

mattiamanzati:
Fixed, will be out with next release

PR #447, opened 2017-10-10
Fix for not being able to add to a map of a map when using environments

xaviergonz:
Basically I stumbled upon a problem while trying to add a node to a map of a map when using environments, failing saying that the environments were not the same.

The newParent environment was undefined and this._environment was correctly set.
Upon checking getEnv (which was actually giving the proper value) I found out that the proper way to get a node environment is to use node.root._environment, and thus this fix.

### DIFF
--- a/src/core/node.ts
+++ b/src/core/node.ts
@@ -314,7 +314,7 @@ export class Node {
             if (
                 !this._parent &&
                 !!this._environment &&
-                this._environment !== newParent._environment
+                this._environment !== newParent.root._environment
             ) {
                 fail(
                     `A state tree cannot be made part of another state tree as long as their environments are different.`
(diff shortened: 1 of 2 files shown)


xaviergonz:
makes me wonder now that I think about it, should this._environment actually be this.root._environment?

mattiamanzati:
Yep, I think so too. :)
While you are at it, can you also add a test for the failing case with the map, so we can avoid regressions in the future? :)

xaviergonz:
@mattiamanzati there you go!

mattiamanzati:
Awesome! Thanks for your work and your time!!!

xaviergonz:
@mattiamanzati my pleasure, any estimates on the release date for the new version?

mattiamanzati:
Not sure yet, wanna release a bunch of fixes together with #446 and #445  :)

xaviergonz:
no worries, guess I'll just have to live with the exception for a while longer :)

PR #448, opened 2017-10-10
Improve Unprotect Example

DyingAlbatross:
The prior example was more appropriate for demonstrating using protect on something that had been previously unprotected. This commit updates the example to use the newer syntax with .actions() and I think makes the use of unprotect a little more clear.


API.md
```diff
@@ -947,17 +947,18 @@ In that case you can disable this protection by calling `unprotect` on the root
...
+    done: false
+}).actions(self => ({ 
     toggle() {
-        this.done = !this.done
+        self.done = !self.done
     }
-})
+}))
 
 const todo = new Todo()
```
mattiamanzati: Whops, this should be Todo.create() :P

mattiamanzati:
Thanks for your PR! Unfortunately API.md is autogenerated from JSDocs in *.ts files, can you please move the changes over mst-operations.ts where the same comment exists? :)

API.md
DyingAlbatross: Whoops is right! :)

DyingAlbatross:
Sure thing!

DyingAlbatross:
All set @mattiamanzati :thumbsup:

mattiamanzati:
Awesome! Thanks for all of your work! :D

PR #465, opened 2017-10-13
Fixed links to examples in README.md

dsumer:
fixed example links as they moved from /examples to /packages/mst-example-*


mattiamanzati:
Uh, good catch! Thanks for reporting and your pr! :)

PR #466, opened 2017-10-13
Documentation Updates

DyingAlbatross:
- Adds an example for onAction
- Updates syntax in the frozen example
- Updates syntax in addDisposer example
- Adds unprotect to the array example


mattiamanzati:
Thanks a lot for your work!! :)

Issue #440, opened 2017-10-09
Orders of magnitude slower than mobx


knoopx:
Hi I'm migrating an app from mobx to mobx-state-tree and I'm facing some performance issues.

Here is the simplest example case:

https://mattiamanzati.github.io/mobx-state-tree-playground/#src=import%20%7B%20observable%20%7D%20from%20'mobx'%0Aimport%20%7B%20types%20%7D%20from%20%22mobx-state-tree%22%0A%0Aconst%20MSTStore%20%3D%20types.model(%7B%0A%20%20%20%20items%3A%20types.optional(types.array(types.string)%2C%20%5B%5D)%0A%7D).actions(self%20%3D%3E%20(%7B%0A%20%20%20%20setItems(items)%7B%0A%20%20%20%20%20%20%20%20self.items%20%3D%20items%0A%20%20%20%20%7D%0A%7D))%0A%0Aclass%20MobxStore%20%7B%0A%20%20%20%20constructor()%7B%0A%20%20%20%20%20%20%20%20this.items%20%3D%20observable(%5B%5D)%20%0A%20%20%20%20%7D%0A%20%20%20%20setItems(items)%7B%0A%20%20%20%20%20%20%20%20this.items%20%3D%20items%0A%20%20%20%20%7D%0A%7D%20%0A%0Aconst%20items%20%3D%20Array.from(Array(100000).keys()).map(x%20%3D%3E%20%60Item%20%24%7Bx%7D%60)%0A%0Aconst%20mstStore%20%3D%20MSTStore.create()%0A%0Aconsole.time(%22mst%22)%0AmstStore.setItems(items)%0Aconsole.timeEnd(%22mst%22)%0A%0Aconst%20mobxStore%20%3D%20new%20MobxStore()%0A%0Aconsole.time(%22mobx%22)%0AmobxStore.setItems(items)%0Aconsole.timeEnd(%22mobx%22)

Replacing an observable array with 100.000 items takes ~5sec when using MST and just 0.02sec when using mobx. The worst part is that it completely blocks the UI while doing so. Is there something I could do to improve speed?

**EDIT:**  Running with `NODE_ENV=production` does not improve things too much.

```
development
mst: 744.248291015625ms
mobx: 0.052001953125ms
```

```
production
mst: 645.447021484375ms
mobx: 0.064208984375ms
```

mattiamanzati:
[...]

PR #474, opened 2017-10-18
perf: introduce ImmutableNode

mattiamanzati:
After the discussion in #440, this is one of the possible improvements in MST core to make things faster.

We have split up Node into Node and ImmutableNode; the Immutable one does not allow applying snapshots, patches or emits any event. ImmutableNode will be used whenever the value does not allow to attach a "$treenode" property.

The speedtest runt does not measure those actions performance, but as you can see we obtained a performance gain because we dont allocate the arrays for the listeners and the loops.

Sample breakdown

Future steps:
- [ ] make isAlive always return true
- [ ] skip isProtected/isRunningAction checks; the value won't change, a new node will be created instead


mweststrate:
@mattiamanzati what was the status of this PR?

mattiamanzati:
Ideally is mergeable, but this is only a first step I think :)
Not enogh time in these days unfortunately :/

robinfehr:
quickly checked the action performance - would say it didn't really change much there.
but my actions might be too little to see much.
```
test one: create with depth 1, depth 2
master:
0.04119373378717422 "avg action time"
0.048729987542637805 "avg action time"
0.03871607269745846 "avg action time"
0.030418973542940635 "avg action time"
0.037991712434218027 "avg action time"
0.038821287658188465 "avg action time"

feature/immutable-node:
0.038274623226561534 "avg action time"
0.04373809615146877 "avg action time"
0.0420638497940525 "avg action time"
0.04047102012683906 "avg action time"
0.04185781622611554 "avg action time"
0.04029827629955466 "avg action time"

Test 2: create nodes depth 1, move node
master:
0.04226841871699275 "avg action time"
0.02680902302315396 "avg action time"
0.03834891667170472 "avg action time"
0.02730177676072416 "avg action time"
	
feature/immutable-node:
0.028105118988837337 "avg action time"
0.036905018670854876 "avg action time"
0.028750588500201832 "avg action time"
0.025291647144669048 "avg action time"

Test 3: create node depth 1, zooming
Master:
0.08172772939205757 "avg action time"
0.09013432225053801 "avg action time"	
0.08650827861457419 "avg action time"
0.05078110070350211 "avg action time"
Feature/immutable-node:
0.07391971327567357 "avg action time"
0.06865690040787775 "avg action time"
0.058948347130167354 "avg action time"
0.07294465531685651 "avg action time"
```

mattiamanzati:
"but my actions might be too little to see much"
This improvement should gain performances when you have actions that modify lot of models with lots of properties or arrays of scalar values :)

mattiamanzati:
@mweststrate I think we could merge as is, what do you think about it?

mweststrate:
Could you split it to 3 files? Further LGTM!

mattiamanzati:
@mweststrate I moved all the things aroud, but had no luck with fixing circular deps, it seems like that isNode wont be exported, so I think there is some circular-magic going on, when you have spare time, please check it out :)

Bnaya:
Whats the connection between `types.frozen` and ImmutableNode?
Is it a replacement, or not something completely different ?

mattiamanzati:
Oh, right, better give some explanation here :)
The way MST works, is by attaching "$treenode" to any possible instance. In that treenode, a "Node" instance is stored. In the node instance, there are infos like the instance type and path relative to the root parent. Nodes are also the point where onSnapshot/onPatch/etc... listener are attached and stored.
Before this PR, we had a Node for each instance of a tree. For e.g. if you have a model with 3 properties, you would have 4 Nodes, one that will be attached at the root, and 3 that will represent the model properties. Unfortunately $treenode property can't be attached to a string, so MST will simply drop off that Node, and create a new one each time. This is very expensive, because we allocate listeners array and snapshot reaction even for immutable value.
onSnapshot("string") would'nt also make much sense. With this PR, we introduce a new type of node that will store the same information the old one did for mutable values, except the one that are useless for "unattacchable" value.
With the latest commit, this has been renamed to ScalarNode and ObjectNode, as it will represent better that any JS scalar value will have a ScalarNode, and any other mutable value will have an ObjectNode instead. Hope I explained it at least a bit :)

alexvcasillas:
This looks really promising, I would like to test it into my “monster” MST store because I have like a lot of large model with lots of childrens and at the current version it’s working very smooth :)
Great job @mattiamanzati !!

mweststrate:
Merged!

PR #483, opened 2017-10-23
Expected 'than' to be 'then'

Dafrok:


mattiamanzati:
Good catch! Thanks! :D

PR #493, opened 2017-10-27
Add version information about flow function

dargue3:
Ran into this problem myself after recently adding MST to my project. `yarn add mobx-state-tree` gave me `v1.0.1` which does not include `flow()`. Figured I'd possibly save someone some time and add a quick note about the required version.


mattiamanzati:
Makes sense! Thanks for the PR! :)

PR #502, opened 2017-11-02
enable model base type -- types.model<SomeBaseType>(...)

devdoomari:
Here's a use case for this PR:
```typescript
export type WalletType = {
   currency: string,
   amount: number,
}
// note: Wallet is used on API side:
export async function getWallet(user: UserType): Promise<WalletType> {
  ...
}
export const Wallet = types.model<WalletType>('wallet', {
  currency: types.string,
  amount: types.number,
})
.actions(self => ({
  action1() {
  },
});

const wallet = Wallet.create( ... )
// before this PR, wallet loses types defined in actions
wallet.action1(); // works with PR
```

### DIFF
--- a/packages/mobx-state-tree/src/types/complex-types/model.ts
+++ b/packages/mobx-state-tree/src/types/complex-types/model.ts
@@ -439,7 +439,7 @@ export type Snapshot<T> = {
     [K in keyof T]?: Snapshot<T[K]> | any // Any because we cannot express conditional types yet, so this escape is needed for refs and such....
 }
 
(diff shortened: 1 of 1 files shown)


mattiamanzati:
Uhm, I am not able to reproduce the bug... thats strange :/

image

devdoomari:
@mattiamanzati that's strange...
I just rechecked and ...well I guess I can just close the PR...

mattiamanzati:
Thanks anyway for the PR :)

PR #504, opened 2017-11-02
undoManger - grouping & declarative withoutUndo

robinfehr:
- the undoManager can now handle actions within actions

 - similarly to withoutUndo users can now add groups like:
```
handleStop = (mousePosition, { dx, dy }) => {
  this.stopTrackingDrag();
  undoManager.stopGroup();
}

handleDrag = (mousePosition, { dx, dy }) => {
  if (dx === 0 && dy === 0) return;

  const { view, parentNode } = this.props;
  undoManager.startGroup(() =>
    parentNode.moveSelectedNodes({ dx: dx / view.zoom, dy: dy / view.zoom })
  );
}
```

### DIFF
--- a/packages/mst-middlewares/README.md
+++ b/packages/mst-middlewares/README.md
@@ -360,13 +360,13 @@ import {undoManager} from './Store/'
 })
 ```
 
-StartGroup, EndGroup - within a react  component:
+StartGroup, StopGroup - within a react  component:
 ```js
 import {undoManager} from '../Store'
 ...
 handleStop = (mousePosition, { dx, dy }) => {
   this.stopTrackingDrag();
-  undoManager.endGroup();
+  undoManager.stopGroup();
 }
 
 handleDrag = (mousePosition, { dx, dy }) => {
--- a/packages/mst-middlewares/README.md
+++ b/packages/mst-middlewares/README.md
@@ -177,67 +177,139 @@ if (timeTraveller.canUndo)
 
 # UndoManager
 
-The `UndoManager` model is quite similar to the `TimeTraveller`. However, it has a few unique features.
-Because it records patches instead of snapshots, it is better capable in dealing with concurrency and asynchronous processes.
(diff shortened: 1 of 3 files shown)


robinfehr:
not thought through idea: what if we'd put the skipping = false later - after the action is not applied to the history - like that it might be possible to add the undo-manager in a more declarative way to the actions too.

```
actions(self => {
   return {
       actionName: ({ prop1 }) => undoManger.withoutUndo(() => {
          self.prop1 = prop1;
       }
   }
})
```

chemitaxis:
This is amazing @robinfehr can you please tell me more or showing an example of this? Thanks!

mweststrate:
@robinfehr I didn't entirely get your last comment? Could you elaborate a bit on it?

robinfehr:
@mweststrate was a bit tricky to explain in words, check out the last commit to see what I've meant.
it is now also possible use the withoutUndo within the actions in a declarative way. (see example above)

will add the readme changes in a bit :)

chemitaxis:
@robinfehr Thanks! ;)

robinfehr:
quickly tested flow's - it's not working yet but it's not due to the changes within this branch (yet) but due to that, the flow_spawn is somehow not within createActionTrackingMiddleware. will dig deeper.

mweststrate:
@robinfehr so far the PR looks good :) 

* [ ] I would mention that undo groups will record *any* change during their execution period). 
* [ ] I think (with the changes that are just merged to master), that if the `loadBooks` etc is kicked off from `afterAttach` instead of `afterCreate` (or the closure), everything would be recorded correctly. Can you confirm that solves the problem, or do I not understand it correctly?
* [ ] can you add unit tests for the new features?

robinfehr:
should be ready - included the PR #460 thanks @builden

builden:
great, hope merge to master soon:)

mweststrate:
@robinfehr thanks!

PR #515, opened 2017-11-09
Named models in Redux devtools

krosk93:
This gives names in the redux devtools GUI as long as the model has a name to make it easy to track different models.


mweststrate:
Great! Thanks :)

PR #517, opened 2017-11-11
tip: turn `.Type` into interface

Bnaya:


mattiamanzati:
Just as curiosity, does interface ITodo extends Todo.Type {} works the same way? :)

Bnaya:
```
ITodo extends typeof Todo.Type {}
```
Will give you typescript error

mattiamanzati:
Nice! Thanks anyway :)

Issue #533, opened 2017-11-22
Execution order and root.identifierCache is undefined


tim-field:
I've got a pain point with MST where the execution order is causing computed columns to resolve incorrectly

An example here, my models can load related data in with a method such as this

`Foo.loadRelationship('parties')`

```JS
      const loadRelationship = flow(function*(name) {
          // make ajax api call
          const res = yield getRelation(self.type, self.id, name)
          // res.data will be an array of "snapshots", these are stored in a map
          viewer.store.set(res.data) // I'll paste this model later
          self.relationships.set(
            name,
            // If we've got an array here then this is a OneToMany relationship
            Array.isArray(res.data)
              ? OneToManyRelationship.create(res)
              : OneToOneRelationship.create(res)
          )
        })
```     

I then have a view property to expose this data in an easy to access format.

```JS
get relations() {
      return self.relationships.entries().reduce((models, [key, value]) => {
        // value.data will be an array of an object based on the relation type
        // oneToMany === Array, onToOne === Object
        const data = isObservableArray(value.data)
          ? value.data.map(relation => relation.model) // note model resolves object
          : value.data.model // note model resolves object
        models[key] = data
        return models
      }, {})
},
```

This is where I'm running into issues, model here is called from  `self.relationships.set(` in 
but at this time I'm unable to resolve to the correct instance. Here is its definition.

```JS
import { types, resolveIdentifier } from "mobx-state-tree"
import { typeToModel } from "../../utils/models"

const RelationshipData = types
  .model("RelationshipData", {
    id: types.identifier(types.number),
    type: types.string
  })
  .views(self => ({
    get model() {
    
      return resolveIdentifier(typeToModel(self.type), self, self.id)
    
      // This technique fails as well. 
      // Note that data is set in store in loadRelationship method
      // const root = getRoot(self)
      // if (!root.store) {
      //   console.error("no store!") // We get here on the inital data set call :(
      // }
      // return root.store ? root.store[self.type].get(self.id) : null
    }
  }))

export default RelationshipData

```

Throws with
`TypeError: getStateTreeNode(...).root.identifierCache is undefined`

it's source is here, and I'm at a loss how to fix this :(

https://github.com/mobxjs/mobx-state-tree/blob/981f7f7a832ea4aef2e97c4f2017e1e87ad17453/packages/mobx-state-tree/src/core/mst-operations.ts#L448

The end result is that something like this, throws with p is null

```JSX
      <ul>
[...]

Issue #513, opened 2017-11-07
Error updating to MST 1.1.0 from 1.0.1


chemitaxis:
Hi guys, we are having some issue with frozen types after updating to the last version. In our rootStore (we have one for redux dev tools), we are having this issue, but the most important thing here, is that this just happen when we build the project with CRA and upload to the server.

_RootStore.js_

```
const RootStore = types
  .model('RootStore', {
    languageStore: types.optional(LanguageStore, getSnapshot(languageStore)),
    folderStore: types.optional(FolderStore, getSnapshot(folderStore)),
    authStore: types.optional(AuthStore, getSnapshot(authStore)),
    templateStore: types.optional(TemplateStore, getSnapshot(templateStore)),
    geniallyStore: types.optional(GeniallyStore, getSnapshot(geniallyStore)),
    myBrandStore: types.optional(MyBrandStore, getSnapshot(myBrandStore)),
    inspirationStore: types.optional(
      InspirationStore,
      getSnapshot(inspirationStore)
    ),
    invoiceStore: types.optional(InvoiceStore, getSnapshot(invoiceStore)),
    modalStore: types.optional(ModalStore, getSnapshot(modalStore))
  })
  .actions(self => ({
    afterCreate() {
      self.authStore.fetchUserData();
    }
  }));
const rootStore = RootStore.create(); // THE ERROR IS HERE WITH TEMPLATE STORE (SEE BELOW)

export default rootStore;
```

The error is in this store, **templateStore**, we are using frozen on it like this:

```
const TemplateStore = types
  .model('TemplateStore', {
    templates: types.array(Template),
    status: types.enumeration('State', ['loading', 'loaded', 'error']),
    search: types.optional(types.string, ''),
    templateType: types.maybe(types.string),
    filters: types.frozen, // <-- This 
    secondaryFilters: types.frozen, // <-- This
    tags: types.array(types.string),
    templatesBuyedLength: types.maybe(types.number, 0)
  })
... actions, views and stuff
```

Create the template store:

```
const templateStore = TemplateStore.create({
  status: 'loading',
  templates: [],
  filters: {
    isNew: undefined,
    subcategoryBusiness: undefined,
    subcategoryEducation: undefined,
    subcategoryOthers: undefined
  },
  secondaryFilters: {
    premium: undefined,
    free: undefined,
    buyed: undefined
  },
  tags: []
});
export { TemplateStore };
export default templateStore;
```

[...]

PR #537, opened 2017-11-24
feat: run tests in both prod and dev mode

### LINKED ISSUE #531: Identifier cache 'undefined' in production mode
>Note: This was working perfectly fine in development, only after putting my `create-react-app` in production mode does this error occur.

Here's the error I'm seeing:
image

It is referencing an error caused on this line: https://github.com/mobxjs/mobx-state-tree/blob/cc0211ffa3d6dbf9ef14c6a72365e7ccea2700b0/packages/mobx-state-tree/src/core/identifier-cache.ts#L25

I'm taking a snapshot of a model and storing it in the parent. The two models look like the following:

```javascript
const Child = t
  .model({
    id: t.identifier(t.string),
    name: t.string,
  });

const Parent = t
  .model({
    original: t.frozen,
    children: t.map(Child),
  })
  .actions((self) => {
    function makeCopy(child) {
      self.original = getSnapshot(child);
    }
  }); 
```

The first time I run `parent.makeCopy(child)` everything works fine, but any later attempts to store a snapshot of that child will result in this error.

Could be something to do with how caching works in production builds of React or Create React App?

mattiamanzati:
Should fix #531, #533, #513 and other related to different behaviours in prod mode.
Now tests are run both in dev and prod mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,12 +4,13 @@
 		"lerna": "^2.4.0",
 		"husky": "^0.13.4",
 		"prettier": "^1.4.4",
-		"lint-staged": "^3.6.1"
+		"lint-staged": "^3.6.1",
+		"cross-env":"^5.1.1"
 	},
 	"scripts": {
 		"clean": "lerna clean",
 		"build": "lerna run build",
-		"test": "CI=true lerna run test",
+		"test": "cross-env CI=true lerna run test",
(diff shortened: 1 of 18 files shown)


mweststrate:
Awesome!

PR #553, opened 2017-12-05
WIP: Performance improvements

mweststrate:


mweststrate:
Improves performance ~20% so far

packages/mobx-state-tree/src/core/node/node-utils.ts
```diff
@@ -119,3 +62,80 @@ export function canAttachNode(value: any) {
...
+    const targetParts = splitJsonPath(target.path)
+    let common = 0
+    for (; common < baseParts.length; common++) {
+        if (baseParts[common] !== targetParts[common]) break
+    }
+    // TODO: assert that no targetParts paths are "..", "." or ""!
+    return (
+        baseParts
+            .slice(common)
+            .map(_ => "..")
```
aretecode: if this is  for a perf commit, would hoist this fn so an anon fn isn't made ~ every time it runs

mweststrate: Done, thanks for noticing. I suspect modern JVM do that themselves btw

PR #562, opened 2017-12-13
Update middleware readme

leesiongchan:
Should be `flow` not `process`


mattiamanzati:
Uh, right! Thanks for your fix! :D

PR #569, opened 2017-12-20
logo improvement, adjust readme accordingly

robinfehr:
thought - the current README header looks a bit lost too, so I tried to improve it slightly. If you don't like it I can revert it ;)


mweststrate:
Thanks!

PR #570, opened 2017-12-21
Switch link

Eclatant:
First of all, thank you very much for making such a good package.

I understand that the function of the old process was carried over by flow.

By the way, in [Using generators](https://github.com/mobxjs/mobx-state-tree/blob/master/docs/async-actions.md#using-generators) It still shows an example of the process being used, 
so I have switched link to the updated examples to reduce confusions.

Please feel free to let me know if I have made a mistake or if I need to make additional corrections.

Thank you very much.


mattiamanzati:
Good catch! :D 
Thanks for your PR!

Issue #560, opened 2017-12-10
[Feature Request] better error message for Error while converting...


chenyong:
Currently it is:

```
Error: [mobx-state-tree] Error while converting `[{"id":"537c1f4f-22c7-4f96-887a-454322aaa76b","createdAt":"...
at path "/0/properties" value `undefined` is not assignable to type: `Property[]` (Value is not an array), ...
at path "/6/properties" value `undefined` is not assignable to type: `Property[]` (Value is not an array), ...
    at fail (http://192.168.1.116:8080/index.js:22785:11)
    at typecheckPublic (http://192.168.1.116:8080/index.js:24298:9)
    at typecheck (http://192.168.1.116:8080/index.js:24283:5)
    at ArrayType../node_modules/mobx-state-tree/dist/mobx-state-tree.module.js.ArrayType.applySnapshot (http://192.168.1.116:8080/index.js:24746:9)
    at executeAction (http://192.168.1.116:8080/index.js:27375:19)
    at ArrayType.res (http://192.168.1.116:8080/index.js:27366:16)
    at http://192.168.1.116:8080/index.js:23827:31
    at executeAction (http://192.168.1.116:8080/index.js:27375:19)
    at res (http://192.168.1.116:8080/index.js:27366:16)
    at runMiddleWares (http://192.168.1.116:8080/index.js:23709:27)
```

but in reality:

![image](https://user-images.githubusercontent.com/25790987/33802941-8476a912-ddbe-11e7-8ad4-cd6f06fbc76d.png)

I think there are some features I need:

* shorten the data, maybe display it in an object, not in string
* which action triggers the error?
* which path in the tree is the data trying to apply to?
* some empty lines make paths obvious

mweststrate:
Pull requests are welcome :)

PR #578, opened 2018-01-02
change error layout to format error messages with prettied value long…

chenyong:
…er than 280 characters

----

Trying to make some changes to fix https://github.com/mobxjs/mobx-state-tree/issues/560

Changes made:

* added an extra `\n` in error to generate a new line.
* shorten value string in errors, which is longer than 280
* `console.error` to display the original value.
* add `\n` in existing tests.

There are still some concerns:

* `280` might not be the best choice, just it fits my case.
* `console.error()` is too verbose. however I can't use a string since it would be too long to display full value.

### DIFF
--- a/packages/mobx-state-tree/src/core/type/type-checker.ts
+++ b/packages/mobx-state-tree/src/core/type/type-checker.ts
@@ -35,6 +35,14 @@ export function prettyPrintValue(value: any) {
         : isStateTreeNode(value) ? `<${value}>` : `\`${safeStringify(value)}\``
 }
 
+function shortenPrintValue(valueInString: string) {
+    return valueInString.length < 280
+        ? valueInString
+        : `${valueInString.substring(0, 272)}......${valueInString.substring(
+              valueInString.length - 8
+          )}`
+}
+
 function toErrorString(error: IValidationError): string {
     const { value } = error
     const type: IType<any, any> = error.context[error.context.length - 1].type as any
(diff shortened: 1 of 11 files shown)


andrefox333:
Yay for this :)

packages/mobx-state-tree/src/core/type/type-checker.ts
```diff
@@ -116,9 +124,11 @@ export function typecheckPublic(type: IType<any, any>, value: any): void {
     const errors = type.validate(value, [{ path: "", type }])
 
     if (errors.length > 0) {
+        console.error("Failed to create `${type.name}` from:", value)
         fail(
-            `Error while converting ${prettyPrintValue(value)} to \`${type.name}\`:\n` +
-                errors.map(toErrorString).join("\n")
+            `Error while converting ${shortenPrintValue(
+                prettyPrintValue(value)
+            )} to \`${type.name}\`:\n\n` + errors.map(toErrorString).join("\n")
```
mweststrate: sure about these double linebreaks? It might confusingly separate the error from the stack?

chenyong: not a perfect solution but it would make `at path /` really stand out in the scenario https://github.com/mobxjs/mobx-state-tree/issues/560 . I'm I sure that it would help.

mweststrate: Ah got it. Maybe we should prefix each line with `\t` or `\t\t` as well to make it more stand out?

chenyong: I think an extra line break is enough in my case. `\t` is not necessary, although it might make it look better in some cases.

mweststrate: would you mind adding the `\t` as well? Then we improve in as many cases as easily feasible

chenyong: I'm considering that using 4 spaces is better than using a `\t`?

mweststrate:
Somehow forgot to merge this earlier. Done so now. Thanks @chenyong

PR #604, opened 2018-01-16
warning if snapshot undefined after postProcessSnapshot

robinfehr:
since it's very unlikely someone returns undefined on purpose from within a postProcessSnapshot action, I added this warning since it cost me quite some time today and hope it might help others.


mweststrate:
Ha this brings up an interesting issue. I highly doubt if middleware should be applied to postProcessSnapshot at all...

mweststrate:
cc @mattiamanzati

robinfehr:
make sense - this warning might be placed better generally on the return value of the middleware.

robinfehr:
obv. to be solved in a more general way.

PR #609, opened 2018-01-18
WIP: Improve custom refs

mweststrate:


mattiamanzati:
I think we need an additional method like isReferredInstance to know If "value" is a snapshot or an instance of Decimal

mweststrate:
Closed in favor of #689

PR #612, opened 2018-01-20
undoManger - withoutundo within undo

robinfehr:
so it can also handle declarative withoutUndo actions within recorded actions.
e.g.: 

```js
.actions(self => {
            return {
                yinc: () =>
                    undoManager.withoutUndo(() => {
                        self.y += 1
                    }),
                xinc: () => {
                    ;(self as any).yinc()
                    self.x += 1
                }
            }
        })
t.is(store.x, 1)
t.is(store.y, 1)
store.xinc()
t.is(store.x, 2)
t.is(store.y, 2)
t.is(undoManager.canUndo, true)
t.is(undoManager.canRedo, false)
undoManager.undo()
t.is(store.x, 1)
t.is(store.y, 2)
```

### DIFF
--- a/packages/mst-middlewares/src/undo-manager.ts
+++ b/packages/mst-middlewares/src/undo-manager.ts
@@ -35,37 +35,71 @@ const UndoManager = types
         }
     }))
     .actions(self => {
-        let skipping = false
-        let flagSkipping = false
         let targetStore: IStateTreeNode
-        let replaying = false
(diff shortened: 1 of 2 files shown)


zhDmitry:
Looks like hack, but not like a real feature.

robinfehr:
async not working properly yet. will dig into it and see if it's solvable.

PR #624, opened 2018-01-25
Update README.md

atsu85:
fix links from `asReduxStore` and `connectReduxDevtools` to example project (project location has changed)


mattiamanzati:
Good catch! Thanks again!

PR #633, opened 2018-01-31
performance actionTrackingMiddleware

robinfehr:
reduce memory allocation (delete of recording entries)
and reduce gc collecting interval (by lifting the var) - noticeable for example when zooming


mweststrate:
Looking good, feel free to merge

Issue #629, opened 2018-01-27
Typescript 2.7-rc compilation errors


xaviergonz:
Due to some new strict mode constraits with TS 2.7 compilation fails:

```
  node_modules/mobx-state-tree/dist/types/complex-types/model.d.ts:29:5 - error TS2416: Property 'props' in type 'ModelType<S, T>' is not assignable to the same property in base type 'IModelType<S, T>'.
    Type '<SP, TP>(properties: { [K in keyof TP]: IType<any, TP[K]>; } & { [K in keyof SP]: IType<SP[K], an...' is not assignable to type '<SP, TP>(props: { [K in keyof TP]: TP[K] | IType<any, TP[K]>; } & { [K in keyof SP]: SP[K] | ITyp...'.
      Types of parameters 'properties' and 'props' are incompatible.
        Type '{ [K in keyof TP]: TP[K] | IType<any, TP[K]>; } & { [K in keyof SP]: SP[K] | IType<SP[K], any>; }' is not assignable to type '{ [K
in keyof TP]: IType<any, TP[K]>; } & { [K in keyof SP]: IType<Snapshot<SP>[K], any>; }'.
          Type '{ [K in keyof TP]: TP[K] | IType<any, TP[K]>; } & { [K in keyof SP]: SP[K] | IType<SP[K], any>; }' is not assignable to type '{ [K in keyof TP]: IType<any, TP[K]>; }'.

  29     props<SP, TP>(properties: {
         ~~~~~
```

nikita9:
```> tsc

node_modules/protractor-element-extend/js/index.d.ts(11,22): error TS2415: Class 'BaseArrayFragment<T>' incorrectly extends base class 'ElementArrayFinder'.
  Types of property 'filter' are incompatible.
    Type '(filterFn: (elementFinder?: T, index?: number) => boolean | Promise<boolean> | Promise<boolean>) ...' is not assignable to type '(filterFn: (element: ElementFinder, index?: number) => boolean | Promise<boolean>) => ElementArra...'.
      Types of parameters 'filterFn' and 'filterFn' are incompatible.
        Type 'boolean | Promise<boolean>' is not assignable to type 'boolean | Promise<boolean> | Promise<boolean>'.
          Type 'Promise<boolean>' is not assignable to type 'boolean | Promise<boolean> | Promise<boolean>'.
[...]

PR #635, opened 2018-02-01
fix: enhance types for modelType in Typescript 2.7.1

duckywang1:
Today I also met the same problem #629  when I upgrade to Typescript 2.7.1
Maybe Typescript do more strict type checking to the instances.  

Actually I don't know the history or context of `ModelType`. I try to do something that fits to Typescript type checking.


mweststrate:
Closed in favor of #667

mweststrate:
Thanks for the effort nonetheless!

PR #640, opened 2018-02-06
Update README.md

chenxiaochun:


chenxiaochun:
ok. I don't understand the mean of this word and try to translate it use [google translate](https://translate.google.cn/#en/zh-CN/memoized), but it is fail.

So i think so it may be 'memoried'. 😳

PR #649, opened 2018-02-10
reflection

robinfehr:


packages/mobx-state-tree/src/core/mst-operations.ts
```diff
@@ -647,6 +647,38 @@ export function walk(target: IStateTreeNode, processor: (item: IStateTreeNode) =
     processor(node.storedValue)
 }
 
+export interface IReflected {
```
mweststrate: maybe `IModelReflectionData`?

packages/mobx-state-tree/src/core/mst-operations.ts
```diff
@@ -647,6 +647,38 @@ export function walk(target: IStateTreeNode, processor: (item: IStateTreeNode) =
...
+ * Returns a reflection of the node
+ *
+ * @export
+ * @param {IStateTreeNode} target
+ * @returns {IReflected}
+ */
+export function reflect(target: IStateTreeNode): IReflected {
+    // check all arguments
+    if (process.env.NODE_ENV !== "production") {
+        if (!isStateTreeNode(target))
```
mweststrate: I think we should also check if it is a model type specifically? (not array or map)

packages/mobx-state-tree/src/core/mst-operations.ts
```diff
@@ -647,6 +647,38 @@ export function walk(target: IStateTreeNode, processor: (item: IStateTreeNode) =
...
+            fail("expected first argument to be a mobx-state-tree node, got " + target + " instead")
+    }
+    const node: any = getStateTreeNode(target)
+    const type = node.type.subType || node.type
```
mweststrate: Remind me: why is the `subType` needed here :)?

packages/mobx-state-tree/src/core/mst-operations.ts
```diff
@@ -647,6 +647,38 @@ export function walk(target: IStateTreeNode, processor: (item: IStateTreeNode) =
...
+    const reflected: IReflected = {
+        name: type.name,
+        properties: type.properties,
+        actions: Object.keys(node.type.actions || {}),
+        volatile: Object.keys(node.type.volatile || {}),
+        views: Object.keys(node.type.views || {})
```
mweststrate: for the views, should we somehow make clear whether a view is a function or a property? Or separately expose as `viewProperties` and `viewFunctions`? (implementation wise you could check whether the property descriptor has `get` )

packages/mobx-state-tree/src/core/mst-operations.ts
robinfehr: currently it also works with a map - that's also why the subtype

packages/mobx-state-tree/src/types/complex-types/model.ts
```diff
@@ -147,17 +148,20 @@ export class ModelType<S, T> extends ComplexType<S, T> implements IModelType<S,
...
         return this.cloneAndEnhance({ initializers: [actionInitializer] })
     }
 
-    instantiateActions(self: T, actions: { [name: string]: Function }) {
+    instantiateActions(self: T, actions: { [name: string]: Function }, node: ObjectNode) {
         // check if return is correct
         if (!isPlainObject(actions))
             fail(`actions initializer should return a plain object containing actions`)
+
+        if (node.type) addReadOnlyProp(node.type, "actions", actions)
```
mweststrate: 2 questions about this approach

1. potentially several instances could expose different actions (I mean you could return actions conditionally). However I don't think that matters much, as that would be bad practice anyway
2. I think this will break for multiple chained `actions` calls? Or no, probably, it doesn't because each time a new type is produced :thinking: . Let's make sure that is verified by some test.

robinfehr: lets' verify both 👍

packages/mobx-state-tree/test/reflection.ts
```diff
@@ -0,0 +1,73 @@
...
+        }
+    })
+    .views(self => ({
+        get viewName() {
+            return 1
+        }
+    }))
+
+test("reflection model", t => {
+    const node = Model.create()
```
mweststrate: can you add a test that `model({ x: types.string}` and then assert `reflect(type).properties.x === types.string`?

packages/mobx-state-tree/src/core/mst-operations.ts
mweststrate: Never mind, you made that work as well :)

packages/mobx-state-tree/src/types/complex-types/model.ts
mweststrate: can you make sure that the fields like `actions`  are predeclared on ComplexType? (initialized as `utils.EMPTY_ARRAY`)

mweststrate:
Great stuff @robinfehr! I left a bunch of comments & questions, but overal looking great :)

packages/mobx-state-tree/src/types/complex-types/model.ts
robinfehr: about 1. i think that's good since the implementation is a representation of the created node, not the type itself.

robinfehr: we might want to change the name to indicate that a bit more. e.g. `getMembers(node)`
might be easier to understand and the sugar around maybe too

e.g. 
```js
getMembers(node, publicOnly)
getPrivateMembers(node) // if such ever would be introduced
getActions(node)
...

```

robinfehr:
@mweststrate I rewrote that so it's all on the node since with how we currently initialise the members (on the node creation) it's cleaner like that. would be glad if you could review that again.

mweststrate:
Hey @robinfehr, thanks! I like that the reflection is now correct. The risk however is that we make creating object nodes even more expensive (which is already problem). So I played with making the whole reflection api lazy, so that the pain is postponed until somebody actually needs the reflection. The reflection itself is more expensive now (one might want to cache it), but it won't slow down object construction. 

What do you think? https://github.com/mobxjs/mobx-state-tree/commit/3184f26899def6b2a48ab171c5f2483e88dd6497

robinfehr:
@mweststrate looking good. cool idea to check for isComputed etc.

mweststrate:
Merged!

PR #653, opened 2018-02-12
toggle hooks in middleware

robinfehr:
- makes it optional if the hooks are piped to the middleware.
we might wan't to fine grain it so we can choose which hooks.
the default is 'included' to not break current implementations.
```ts
addMiddleware(node, middleware: (actionDescription, next) => any, includeHooks = true)
```

- throws on omitting next() within a middleware

### DIFF
--- a/API.md
+++ b/API.md
@@ -33,6 +33,7 @@ _This reference guide lists all methods exposed by MST. Contributions like lingu
 -   [getSnapshot](#getsnapshot)
 -   [getType](#gettype)
 -   [hasParent](#hasparent)
+-   [hooks](#hooks)
 -   [Identifier](#identifier)
 -   [IdentifierCache](#identifiercache)
 -   [isAlive](#isalive)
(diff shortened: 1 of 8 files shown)


packages/mobx-state-tree/src/core/action.ts
```diff
@@ -126,41 +135,61 @@ export function decorate<T extends Function>(middleware: IMiddlewareHandler, fn:
...
-        const handler = handlers[index++]
-        if (handler) return handler(call, runNextMiddleware)
-        else return mobxAction(originalFn).apply(null, baseCall.args)
+        const middleware = middlewares[index++]
+        const handler = middleware && middleware.handler
+        const invokeHandler = () => {
+            const next = handler(call, runNextMiddleware)
+            if (!next && index < middlewares.length && process.env.NODE_ENV !== "production") {
+                const node = getStateTreeNode(call.tree)
+                fail(
```
mweststrate: I'm not sure about this check, there are cases where it is fine that `next()` isn't called, for example if you have middleware that prevents modifications, or only applies it to a copy, etc. I had a few cases before. I am aware it is pretty tricky if you forget it in your middleware, but I think that is an inherent risk about writing your own middleware and hopefully docs help preventing mistakes there :)

API.md
```diff
@@ -124,7 +124,9 @@ For more details, see the [middleware docs](docs/middleware.md)
 **Parameters**
 
 -   `target` **IStateTreeNode** 
--   `middleware`  
+-   `middleware`  **IMiddleware**
+-   `includeHooks` **([boolean](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) | any)** indicates whether the hooks should be piped to the middleware.
```
mweststrate: elobarote; `that is, should the middleware be run for built in methods like `applySnapshot`

packages/mobx-state-tree/src/core/action.ts
robinfehr: fair point about aborting the middleware queue.
what about returning false if it the abort was intended and only throw on undefined?

robinfehr: e.g. an `abort()` next to `next()`

mweststrate:
Merged!

PR #675, opened 2018-02-24
middleware callback syntax - abort()

robinfehr:
introduces `abort()`

```js
function customMiddleware1(call, next, abort) {
    console.log(call.name)
    next(call, (value) => value + 1) // no return needed, value can be manipulated
}

function customMiddleware2(call, next, abort) {
    console.log(call.name)
    abort('someValue') // 'someValue' will be returned instead of value from the actual action
}

addMiddleware(store, customMiddleware1)
addMiddleware(store, customMiddleware2)

const valueFromAction = store.someAction()  // someValue1
```

because of the syntax change - we can

- check whether at least one of the callbacks was invoked.
- check if accidentally both were invoked.
- eliminate the current problems where one missed to return / didn't call next(call)

TODO: 
- [x] add callback fn within next `next(call, (value) => value)` as a second param to overwrite the return value

TBD: 
- [ ] should abort values also be manipulated? (currently they are)

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -7,13 +7,13 @@ If a tree is protected (by default), this means that any mutation of the tree wi
 
 [SandBox example](https://codesandbox.io/s/mQrqy8j73)
 
-Middleware can be attached by using `addMiddleware(node, handler: (call, next(call) => void) => void)`
+Middleware can be attached by using `addMiddleware(node, handler: (call, next(call) => void, abort(value: any) => any) => void)`
 
 It is allowed to attach multiple middlewares. The order in which middleware is invoked is inside-out:
 local middleware is invoked before parent middleware. On the same object, earlier attached middleware is run before later attached middleware.
 
-A middleware handler receives two arguments: 1. the description of the the call, 2: a function to invoke the next middleware in the chain.
-If `next(call)` is not invoked by your middleware, the action will be aborted and not actually executed.
+A middleware handler receives three arguments: 1. the description of the the call, 2: a function to invoke the next middleware in the chain. 3: a function to abort the middleware queue and return a value.
+You must either call `next(call)` or `abort(value)` within a middleware. If none is invoked an error will be thrown.
 Before passing the call to the next middleware using `next`, feel free to (clone and) modify the call arguments. Other properties should not be modified
 
 A call description looks like:
@@ -44,9 +44,18 @@ A very simple middleware that just logs the invocation of actions will look like
 @example
 ```typescript
 const store = SomeStore.create()
-const disposer = addMiddleWare(store, (call, next) => {
+const disposer = addMiddleWare(store, (call, next, abort) => {
   console.log(`action ${call.name} was invoked`)
   return next(call) // runs the next middleware (or the implementation of the targeted action if there is no middleware to run left)
+})
+```
+
+@example
+```typescript
+const store = SomeStore.create()
+const disposer = addMiddleWare(store, (call, next, abort) => {
+  console.log(`action ${call.name} was invoked`)
+  return abort('value') // aborts running the middlewares and returns the 'value' instead. Note that the underlying action won't be invoked either.
 })
 ```
 
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -1,23 +1,71 @@
 # Middleware
+Middlewares can be used to intercept any action on a subtree.
 
-MST ships with a small set of [pre-built / example middlewares](../packages/mst-middlewares/README.md)
+It is allowed to attach multiple middlewares to a node.
+The order in which middlewares are invoked is inside-out:
+This means that the middlewares are invoked in the order you attach them.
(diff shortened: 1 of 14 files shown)


docs/middleware.md
```diff
@@ -44,12 +44,21 @@ A very simple middleware that just logs the invocation of actions will look like
...
+@example
+```typescript
+const store = SomeStore.create()
+const disposer = addMiddleWare(store, (call, next, abort) => {
+  console.log(`action ${call.name} was invoked`)
+  return abort('value') // aborts running the middlewares and returns the 'value' instead. Note that the underlying action won't be invoked either.
+})
+```
+
 _Note: for middleware, it is extremely important that `next(action)` is called for asynchronous actions (`call.type !== "action"`), otherwise the generator will remain in an unfinished state forever_
```
mweststrate: I think this note can be removed?

Also, can you add a test to verify that having an aborting middleware does properly finish the generator when using asynchronous actions?

packages/mobx-state-tree/src/core/action.ts
```diff
@@ -126,41 +136,94 @@ export function decorate<T extends Function>(middleware: IMiddlewareHandler, fn:
...
+            // overwrite the result
+            // can be manipulated through middlewares earlier in the queue using the callback fn
+            result = value
+        }
+        const invokeHandler = () => {
+            handler(call, next, abort)
+            if (process.env.NODE_ENV !== "production") {
+                if (!nextInvoked && !abortInvoked) {
+                    const node = getStateTreeNode(call.tree)
+                    fail(
```
mweststrate: I think this is a breaking change now? 
What we could do is, `middleware.length === 2 ? warn(message) /* a legacy decorator */ : fail(message)`?

packages/mobx-state-tree/src/core/action.ts
```diff
@@ -126,41 +136,94 @@ export function decorate<T extends Function>(middleware: IMiddlewareHandler, fn:
...
+                        `Neither the next() nor the abort() callback within a middleware for the action: "${call.name}" on the node: ${node
+                            .type.name} was invoked.`
+                    )
+                }
+                if (!!nextInvoked && !!abortInvoked) {
```
mweststrate: These are bools, no need for `!!`

packages/mobx-state-tree/src/core/action.ts
```diff
@@ -126,41 +136,94 @@ export function decorate<T extends Function>(middleware: IMiddlewareHandler, fn:
...
+            result = value
+        }
+        const invokeHandler = () => {
+            handler(call, next, abort)
+            if (process.env.NODE_ENV !== "production") {
+                if (!nextInvoked && !abortInvoked) {
+                    const node = getStateTreeNode(call.tree)
+                    fail(
+                        `Neither the next() nor the abort() callback within a middleware for the action: "${call.name}" on the node: ${node
+                            .type.name} was invoked.`
```
mweststrate: could include `middleware.name`, so that if middleware is named, it can be more easily found

packages/mobx-state-tree/src/core/action.ts
```diff
@@ -126,41 +136,94 @@ export function decorate<T extends Function>(middleware: IMiddlewareHandler, fn:
...
+                    )
+                }
+                if (!!nextInvoked && !!abortInvoked) {
+                    const node = getStateTreeNode(call.tree)
+                    fail(
+                        `The next() and abort() callback within a middleware for the action: "${call.name}" on the node: ${node
```
mweststrate: Neat!

packages/mst-middlewares/test/custom.ts
```diff
@@ -0,0 +1,137 @@
...
+import atomic from "../src/atomic"
+import { types, addMiddleware, getSnapshot } from "mobx-state-tree"
+
+let error: any = null
+
+function omitNextAbort(call, next) {
+    // omit next() / abort()
+}
+function nextAndAbort(call, next, abort) {
+    abort("someValue")
```
mweststrate: I think we should clearly document that `abort` and `next` should be invoked _synchronously_

packages/mst-middlewares/test/custom.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+function abortNumeric(call, next, abort) {
+    abort(5)
+}
+function nextNoAlter(call, next) {
+    next(call)
+}
+function nextAlter(call, next, abort) {
+    next(call, value => value + 1)
```
mweststrate: I don't get this form entirely yet, what is the advantage over 

`return next(call) + 1` ?

robinfehr: i guess both would be do able. added this since it separates the actual middleware logic from the value manipulation logic a bit more.

mweststrate: I think we should go for "one way of doing things"

mweststrate:
Merged!

Issue #644, opened 2018-02-07
Compose model with function as prop (types.function)


alexwasner:
I'm converting a project to state tree and have some properties that are functions. For example, an `action` property that you can pass a function to run at certain instances. I couldn't find anything in the docs related to a `types.function`.

I want to do something like this:

```
const ButtonModel = types.model({
  action:types.optional(types.function,()=>{}),
  icon:types.number
})
```

maxguzenski:
Props has to be serializable... You have to use volatile for that.

 ```
const ButtonModel = types.model({
  icon:types.number
}).volatile(self => ({
  action: () => ()
}))
```

**Comment by barbalex** [comment]
@maxguzenski where is `volatile` documented? Trying to use it like this:
```js
export default types
  .model({
    showDeleted: types.optional(types.boolean, false),
    deletionMessage: types.optional(types.string)
  })
  .volatile(self => ({
    deletionCallback: null
  }))
```

I got an error:

```
mobx-state-tree.module.js:2348 Uncaught Error: [mobx-state-tree] Error while converting `undefined` to `string`:

    value `undefined` is not assignable to type: `string` (Value is not a string).
    at fail (mobx-state-tree.module.js:2348)
    at typecheckPublic$$1 (mobx-state-tree.module.js:2106)
    at typecheck$$1 (mobx-state-tree.module.js:2091)
    at Object.optional$$1 [as optional] (mobx-state-tree.module.js:4244)
    at Object.<anonymous> (index.js:11)
    at Object../app/store/index.js (renderer.dev.js:2648)
    at __webpack_require__ (bootstrap:724)
    at fn (bootstrap:101)
    at Object.<anonymous> (index.js:10)
    at Object../app/index.js (renderer.dev.js:2122)
```

I am trying to pass a callback to the deletion modal when a user deletes a dataset. The callback would be executed after the user confirms the deletion.

barbalex:
@maxguzenski Uups, seems that I stumbled over a different error. Sorry for bugging you

Issue #641, opened 2018-02-06
Get missing reference identifier


TimHollies:
Apologies if I've missed an obvious way to do this - but I'm a bit stuck.

I have a reference in a model which links to another model which is yet to be loaded. Attempting to dereference the property (correctly) throws an error. Is there an easy way to access the identifier the reference is looking for.

E.g. Attempting to access the property throws: `Error: [mobx-state-tree] Failed to resolve reference of type EntityType: 'd584fa4e-fb7a-4875-8048-3870118fcdad'`. I want to access the value `d584fa4e-fb7a-4875-8048-3870118fcdad`.

I've had a few ideas about how to do this:
- Get a snapshot of the whole model and get the identifier out of the JSON object
- Catch the error and parse the error message string
- Create a custom reference and return the identifier as a string or throw an easy-to-parse error if the reference is not found 

These all seem a bit overkill to retrieve a string value - is there an API function I'm missing?

robinfehr:
what do you mean by yet to be loaded? (currently i'm assuming you did serialize your state and try to load the snapshot)
could you also show us how you defined your references?

TimHollies:
I meant that the model with the identifier that the reference points to is not yet in the tree. Here's a toy example:

```javascript

// setup types

const Person = types.model({
  id: types.identifier(types.string),
  name: types.string
});

const Cat = types.model({
  id: types.identifier(types.string),
  owner: types.reference(Person)
});

const Container = types.model({
  cat: Cat,
  person: Person
});

// database helper functions

const loadCat = (id) => { 
  //return Cat model 
};

const loadPerson = (id) => { 
  //return Person model 
};

// load cat from data store

/*
{
  id: "cat_id_1",
  owner: "person_id_1"
}
*/

const cat = await loadCat("cat_id_1");

// now load the owner

const person = await loadPerson(cat.owner); // throws

// create complete tree for use in program

const container = Container.create({
  cat, person
});

```

The workaround I am currently using is:

```javascript
const person = await loadPerson(getSnapshot(cat).owner);
```

What I'm hoping for is that there's something like this which I'm missing:

```javascript
const person = await loadPerson(cat.owner.reference);
// or 
const person = await loadPerson(getReference(cat.owner));
```

I would like to be able to get the string 'person_id_1' out of the 'cat' model without having to create a snapshot of the whole model. 

[...]

PR #678, opened 2018-02-27
Added 'getIdentifier' function

### LINKED ISSUE #674: expose utility function `getIdentifier(thing)` to get the identifier of an model instance that has an identifier attribute
Should address #641, #643.

TimHollies:
Exposes 'getIdentifier' function to resolve #674. 

```javascript
/**
 * Returns the identifier of the target node. If a property is supplied returns the identifier
 * of the property even if it is a reference to a missing object.
 *
 * @export
 * @param {IStateTreeNode} target
 * @param {string} property
 * @returns {(string | null)}
 */
export function getIdentifier<T extends IStateTreeNode>(target: T, property?: keyof T): string | null;
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,5 @@
 language: node_js
 sudo: required
-before_install: # if "install" is overridden
-  # we don't want the default yarn version
-  - which yarn
-  - sudo rm `which yarn`
-  # Repo for Yarn
-  - sudo apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg
-  - echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
-  - sudo apt-get update -qq
-  - sudo apt-get install -y -qq yarn=1.2.1-1
-  - which yarn
-  - yarn --version
 cache:
   yarn: true
 script: yarn run bootstrap && yarn run build && yarn run test && yarn run coverage
--- a/packages/mobx-state-tree/src/core/mst-operations.ts
+++ b/packages/mobx-state-tree/src/core/mst-operations.ts
@@ -452,18 +452,13 @@ export function resolveIdentifier(
 }
 
 /**
- * Returns the identifier of the target node. If a property is supplied returns the identifier
- * of the property even if it is a reference to a missing object.
+ * Returns the identifier of the target node.
  *
  * @export
  * @param {IStateTreeNode} target
- * @param {string} property
- * @returns {(string | null)}
+  * @returns {(string | null)}
  */
-export function getIdentifier<T extends IStateTreeNode>(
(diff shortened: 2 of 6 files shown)


packages/mobx-state-tree/src/core/mst-operations.ts
```diff
@@ -451,6 +451,47 @@ export function resolveIdentifier(
...
+ * of the property even if it is a reference to a missing object.
+ *
+ * @export
+ * @param {IStateTreeNode} target
+ * @param {string} property
+ * @returns {(string | null)}
+ */
+export function getIdentifier<T extends IStateTreeNode>(
+    target: T,
+    property?: keyof T
```
robinfehr: not so sure the property param is needed - omitting it could clean up the code and would be less error prone since string literals can't be checked on e.g. renaming a property.

was there a specific use case you had in mind?

.travis.yml
```diff
@@ -1,20 +1,9 @@
 language: node_js
 sudo: required
-before_install: # if "install" is overridden
```
robinfehr: there already is a separate PR addressing this problem :)

packages/mobx-state-tree/src/core/mst-operations.ts
TimHollies: The property param is for the use case in my original issue #641. 

The idea is that you can access the identifier that is being searched for by a reference, regardless of whether that identifier exists in the tree or not.

I want to be able to do this so I can lazily load references. This can be done using Custom References, but loading them on first access is a bit unpredictable, which I don't like. I know which references are required for each view, so I'd rather load them all at the start in one place. Currently you have to snapshot each object in order to find out the identifiers that need loading - which is not the intended usage of the snapshot feature.

I've added some tests in `packages/mobx-state-tree/test/node.ts` which cover these cases.

.travis.yml
TimHollies: 👍 I thought there might be - I just wanted to be able to check if it would pass the CI tests.

robinfehr:
quick review @TimHollies  
cc @mweststrate

- the property param that is passed to getChildNode(param) seems to be a workaround around getStateTreeNode(node) that fails on resolving the reference.
- i think the current implementation would only work for identifieRreferences, can that be verified?
- i'm not so sure if getting identifiers of non resolvable nodes is a good thing in general. 
since to support customRefs and custom types we'd would probably have to use `getSnapshot`.

for the use case: 
if i got that correct - you have e.g. cats stored and you want to lazily load the owner on `create()`, but not using a custom ref. since they wouldn't eagerly load the owner on creation.
would what i've commented on #644 help?

mweststrate:
Thanks for the PR @TimHollies! I think @robinfehr's comments were spot on, so thanks for improving :). Will merge and release!

Thanks :)

Issue #673, opened 2018-02-22
Introduce custom types


mweststrate:
Some work was already done in #609, but there was no general issue describing the feature. So here it is. 

```
types.custom<S, T>({
   fromSnapshot(value: S): T {
        // will be cached from same S
    },
    toSnapshot(value: T): S {
        // only executed if value is different from the previous one
    }
})
```

This makes it possible to create custom types for Decimals, Dates etc. 

The contract is as follows:
* `T` is considered to be immutable (like frozen)
* `T` is allowed to be non-serializable (unlike frozen). After all, the Snapshot gets serialized in snapshots, patches etc
* `S` can be non-primitive, but should be serializable and immutable (see #588)
* Unlike (custom) references, the values are not computed, tracked, and getter and setter won't receive information about their usage context (like `parent`)

mweststrate:
Example copied from #662:

```javascript
const MyBackendDate = types.custom<string, Date>({
   get(value: string): Date {
        const parsedDate = someWeirdParsingLogicFromTheString(value)
        return parseDate // a normal JS Date object
    },
    set(value: Date): string {
        const serializedForm = someWeirdUnparseLogicFromTheDate(value)
        return serializedForm // e.g. "Tue 7 Octobr '18'"
    }
})

const Appointment = types.model("appointment", {
   subject: types.string,
  date: MyBackendDate
})

const appointment = Appointment.create({
  subject: "Buy snacks",
  date:  "Tue 7 Octobr '18'"
})

console.log(appointment.date instanceof Date) // true
appointment.date = new Date() // n.p.
getSnapshot(appointment).date // some string representing the date
```

robinfehr:
would it be doable to also implement a custom getSnapshot for them?
one could for example then implement the current Date type - while e.g. within react components working with the value from the getter - using another for the snapshot
or am I missing and such would already be achievable?

mweststrate:
@robinfehr the return value from the setter / the input arg of the getter *is* the custom snapshot

mweststrate:
but indeed, I think we should change the implemetation of `types.Date` to be a custom snapshot with this change

robinfehr:
since i was quite confused here a quick summary from chat conversation:

`getter` , `setter` seem to be misleading names.
what was meant by them was:
`toSnapshot`, `fromSnapshot`

`toSnapshot`: invoked when using _getSnapshot(node)_
`fromSnapshot`: invoked when using _Appointment.create({ subject ... })_

[...]

PR #689, opened 2018-03-09
Add support for custom primitive-ish types

mweststrate:
As discussed in #673


packages/mobx-state-tree/src/types/utility-types/custom.ts
```diff
@@ -0,0 +1,145 @@
...
+        return this.name
+    }
+
+    isAssignableFrom(type: IType<any, any>): boolean {
+        return type === this
+    }
+
+    isValidSnapshot(value: any, context: IContext): IValidationResult {
+        if (this.options.isTargetType(value)) return typeCheckSuccess()
+        const typeError: string = this.options.isValidSnapshot(value)
```
robinfehr: i think transpiled to js one could return true / false => the typeError within the error message would be `Invalid value for type '${this.name}': true`

also isValidSnapshot does imply that one should return a boolean.

maybe we could make that a bit more expressive to avoid confusion.

PR #691, opened 2018-03-09
WIP: Feature: flow cancellation

### LINKED ISSUE #681: Cancel async flow action
Is there a reference to the active generator that runs the flow? I need that in order to cancel node fetches before the node is destroyed.

mweststrate:
Fixes #681


mweststrate:
WIP: await and or leverage improved `flow` implementation in Mobx-core package

mweststrate:
w00ps, that close was accidental

elite174:
Actually, I think, this should be done also for Mobx, because mobx also has flows, so this is an update for flow and not for MST.

elite174:
Please, merge this PR

Bnaya:
I'm missing something: how to cancel the actual in-flight operation (like XHR) and not just ignore the result

duro:
@Bnaya that's not really how request cancelation works (except maybe in Firefox). Most XHR cancelation implementations these days simply ignore the response and do not physically close the request. At least that's what I found in my research.

Bnaya:
@duro i find it hard to believe. how did you check? lets say i'm fetching 50mb, and after 1MB i abort the xhr (or fetch()) you say that the browser will not stop the transfer? (if i will have wireshark open i will see 49mb more passing on the wire)

mweststrate:
request cancellation is supported on libraries that build on top of
xmlhttprequest, it is just not supported by the window.fetch api. Whether
MobX / MST will cancel the request depends on whether the pending promise
exposes a `cancel()` methods. So with most libraries, you could support
request cancellation by giving the pending promise is cancel method and
wiring it up to the libs cancellation api (if applicable)

Op zo 29 apr. 2018 om 11:14 schreef Bnaya Peretz <notifications@github.com>:

Bnaya:
@mweststrate Actuall  as chrome 66, it's supported in all major browsers for fetch, using AbortController (somekind of cancellation token)
And seems that it will be the de-facto api for cancelling stuff.
Adding cancellable promise to the standard was cancelled.
You may add cancel, or any other method on Promise, but when you chain/pass to any standard promise, you will lose the cancel method

mweststrate:
Yeah I am aware of this. With this mechanism there is at least a
possibliity of suporting it, with some effort, otherwise there is none (the
abortController is afaik for window.fetch, but not generically applicable
to promises)

Op zo 29 apr. 2018 om 16:38 schreef Bnaya Peretz <notifications@github.com>:

mattiamanzati:
Uhm, instead of throwing a generic error with a text, shoul'nt we instead throw a custom error and provide an utility function like "isCancellation(error)" so users can check that inside the generator catch? :)

shahrouzz:
Is this ready to be merged, since this feature is released for MobX? Would be nice to have the CancellablePromise's in the MST flows as well since they work so well in MobX!

benjamind:
Also very interested in this. In trying to integrate mst with history pop state, it becomes clear that we need to cancel flow's when going forward/backward through history.

davidyxu:
Just wanted to check in on the status of this PR, what sort of work is remaining to get this merged in? It would be great to get have cancellable flows available in MST.

Bnaya:
Im in favor to aducate developers to use the acceptable practices of async operation cancelation,
With after fetch got its abort controller, thats the way to go.

And then the promise the flow is yield on, will be rejected, and the developer may catch it with try catch and check if its abort error etc.

Imo there is no reason to add another api for that.

vonovak:
@mweststrate can you please close the PR if the api is not going to be added to MST? (Personally, feature parity with mobx.flow would be nice but I understand if you don't want to add another api.)

az67128:
Are there any news?

mweststrate:
Not really, but if you are willing to champion the PR and move it forward
I'll happily merge (I think most of the work was actually done already, I
don't remember anymore :-P).

auvipy:
I will check it more thoroughly after I check the mobx flow more deeply.

mweststrate:
Feel free to leave the TS part open (that is put in `:any` wherever
needed). It is actually a non trivial, but pretty cool subject to look
into; as it involves learning generators :)

auvipy:
I am Python guy so new shiny ES.next stuffs are quite cool but not entirely unknown to me. I will surely give this a try.

Bnaya:
Throwing in on-going spec-level efforts to make cancelation standard:
Addin AbortController + AbortSignal to ES spec
https://github.com/tc39/proposal-observable/issues/201 

Stage 2:
https://github.com/tc39/proposal-explicit-resource-management
Symbol based mechanism 
`Symbol.dispose`

Bnaya:
IMO abort token/controller based pattern can be implemented on user-land, and can be abstracted-away easily when any spec will materialized.

wildeyes:
While this feature is still pending, any suggestions on how to do this on the outside?

DenVifer:
I wanted to add comment earlier but didn't have enough time. But I see that you are still discussing this feature, so I want to give my 5 cents.

Waiting for this feature.
I absolutely agree that request cancellation it's better in many cases, but... it's not always necessary. You may deal with fetch API requests, web-sockets, different libraries or some custom async functions. And sometimes it doesn't really matter whether the called function will be interrupted or not. You just need to cancel some deferred code execution.
As for now, I use custom utils written using flow util from mobx.
I also want to mention redux-sagas as an example. Redux saga has API for cancellation and it's quite convenient.

AndrewMorsillo:
I've run into a situation where I need this. I have state trees in a react native app handling "animated sequences". A flow representing a logical grouping of animations will `yield` several promises in sequence that each wait for an event from an RXJS subject indicating that a particular animation(s) is complete. 

That makes it trivial to write out a series of animations in a flow as though it was a synchronous function. It would otherwise be a huge pain to coordinate these animations.

Work great!... except when the user presses the Android back button or otherwise exits the sequence early. There's no good way that I can find to cancel the ongoing flow. This PR would be ideal as I simply want to stop and ignore the rest of these flows if they get interrupted by the user.

Since this has been sitting for awhile I'm guessing it's probably not in a state ready to be merged anymore. I'll try to find time to update it but for now I don't think I'll be able to get to it very soon (or maybe at all). 

Separately (but interestingly IMO) I think the[ new TS 4.1 features ](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/#recursive-conditional-types) will let us have accurate types for generators/flows which is neat.

jamonholmgren:
On reviewing this PR, it looks like Michel got this to a pretty good place and it's a desirable feature, but it will require a bit more thought, resolving conflicts, some documentation, some manual testing, and then for the core team to give the 👍 .

Going to leave this for now, but revisit later.

AndrewMorsillo:
I forked this and got it all up and running. I'm using it in my app and it seems to work great so far. I'll send a PR as soon as I have the time.

Bnaya:
AbortController & signal are now also part of node, which make it de-factor way to cancel things in js.
abort signal is added to more and more browser & node apis 

Cancellable promises are pulled away and are no longer considered to be part of ecma spec
Instead of adding none-spec extension to the returned Promise, we can just tell users to make their async code cancelable and keep the API of the library more focused.

jamonholmgren:
@Bnaya Good point. We should provide ample examples of how to do this.

airhorns:
Some comments as a lurker who wants to cancel stuff:

> AbortController & signal are now also part of node, which make it de-factor way to cancel things in js.

100%, I agree that this should be the way that developers work with cancellable flows in MST instead of non-spec extensions to promises. I think the reason cancellable promises were themselves cancelled all apply to MST just teh same. 

> keep the API of the library more focused

I think that aborts got added to the Node standard library and to the browsers because they are genuinely needed super often. Timeouts and whatnot are a really commonly invoked tool for any two systems that need to talk and are forced to do so asynchronously cause of the way JS works. I agree that all else being equal smaller, focused APIs are better, but in this case, this is a really common need and IMO worth adding some facility to MST to support.

I also think it's important because mobx-state-tree takes away some of the control away from developers for async processes because of the switch from `async` to `flow()`. Threading an abort signal down an entire chain of `flow` calls through all the `yields` seems kind of annoying, and MST takes away the ability to use something like `AsyncHooks` to make it a bit more ergonomic. If an `AbortController` is the main way to cancel flows, is endorsed by the browser vendors and by node core, and all async work done in MST is done via the one `flow` entrypoint, it seems like a really good opportunity to help developers do the right thing from the start by supporting it right in MST (or mobx proper).

Bnaya:
The huge foot-gun with returning spacial promise is:

const runningOperation = runFlow();

// works
runningOperation.cancel()

const runningOperation2 = runFlow().then(() => {
// some other async stuff
});

// ERROR!! calling then will return native promise
runningOperation2.cancel()

And we really don't want to patch `then` and all of rest

airhorns:
> The huge foot-gun with returning spacial promise is:
> And we really don't want to patch then and all of rest

Just to be clear, I agree! I am just advocating more for coupling to the AbortController API from MST and making the controller accessible to developers easily instead of having each user of the library have to reinvent some convention for cancelling controllers.

AndrewMorsillo:
Apologies that I haven't had more time to move this forward yet. I'm not very familiar with the AbortController API. Can somebody provide an example of how the API would be used in the context of cancelling MST flows?

Here's how I'm using the cancellable promise API, so far this works great
```
    .actions((self) => {
        let pendingPromises: CancellablePromise<FlowReturn<any>>[] = []

        /**
         * Based on https://github.com/mobxjs/mobx-state-tree/issues/681
         * Added a .cancel() function to flows so they can be stopped if they're waiting for a yield
         *
         * This wraps a flow and keeps a reference to its promise so it can be cancelled
         * @param generator
         */
        function trackedFlow<R, Args extends any[]>(
            generator: (...args: Args) => Generator<Promise<any>, R, any>
        ): (...args: Args) => Promise<FlowReturn<R>> {
            //Make flow function
            const theFlow = flow(generator)
            //Return a wrapper function that keeps track of when it's called and cleans up after it
            return function wrappedFlow(...args: Args): Promise<FlowReturn<R>> {
                //Call the flow
                const p = theFlow(...args)
                //Record the promise
                pendingPromises.push(p)
                //When promise is resolved or rejected remove the flow from tracked
                p.finally(function removeCompletedFlow() {
                    _.remove(pendingPromises, (pending) => pending === p)
                })
                //Return the promise to caller
                return p
            }
        }


      const myAsyncAction = trackedFlow(function* myAsyncAction( ) {
            .....
            yield when(() => self.somethingThatMightOrMightNotHappenBeforeDestroy)
            .....
      })

     function beforeDestroy() {
            //Cancel all outstanding flows when the model is torn down
            pendingPromises.forEach((p) => p.cancel())
            pendingPromises = []
        }


    return {
        beforeDestroy,
       myAsyncAction
     }
```

I'm doing a bit more than this in practice by cancelling some flows based on conditions using a `reaction`, but this example illustrates how cancellation can be practically used. The `.cancel()` API works just fine for these purposes but as others have pointed out it may not be ideal for other situations. I'm interested to learn more about other use cases and how AbortController might be used to replicate the above behavior.

jamonholmgren:
@AndrewMorsillo Thanks for the usage example! I'm also interested to see other examples where your approach doesn't quite work.

DonnyVerduijn:
For a project i'm currently working on, i've taken a fork of MST and exposed a cancel method by letting flow return a tuple , but the implementation right now only throws inside the generator. It would be nice to also reject the promise returned from the flow, but somehow i couldn't catch the error from the promise, even though catch handlers lower down the stack were triggered.

I understand that we want to be careful here. How about creating an additional package in the monorepo to provide more functionality which could stay experimental until a final decision has been made?

jamonholmgren:
That makes a ton of sense @DonnyVerduijn. (Sorry, didn't see this until now!)

I'd be happy to review a PR to that effect.

runjuu:
👀 If anyone wants to use more "reactive" ways to control the side-effect, like **cancel previous request** or **wait until previous request finishes**, they could take a look at **[mst-effect](https://github.com/runjuu/mst-effect)**, which allows you to use [RxJS](https://github.com/ReactiveX/rxjs) along with mobx-state-tree. It should be super easy to integrate mst-effect with your existing code base.

packages/mobx-state-tree/test/async.ts
```diff
@@ -280,3 +280,114 @@ function filterRelevantStuff(stuff) {
...
+        err => {
+            fail()
+        }
+    )
+    promise.cancel()
+})
+
+test("flows can be cancelled - 3 - rethrow cancellation", done => {
+    let steps = 0
+    const M = types.model({}).actions(self => ({
```
trinadhkoya, line 339: If we are not using self, you can remove it for now. Just replace it with anonymous function

packages/mobx-state-tree/test/async.ts
```diff
@@ -280,3 +280,114 @@ function filterRelevantStuff(stuff) {
...
+            expect("" + err).toBe("Error: FLOW_CANCELLED")
+            done()
+        }
+    )
+    promise.cancel()
+})
+
+test("flows can be cancelled - 3 - pending Promise will be ignored", done => {
+    let steps = 0
+    const M = types.model({}).actions(self => ({
```
trinadhkoya, line 368: If we are not using self, you can remove it for now. Just replace it with anonymous function

coolsoftwaretyler:
@trinadhkoya - thanks for giving this PR some review. Do we know if @mweststrate is still interested in moving this PR forward? I know he's pretty busy these days. I just want to make sure that if we're moving this PR ahead, we've got someone responsible for managing it overall.

chakrihacker:
Can we suggest this 3rd party library https://www.npmjs.com/package/mst-async-task or merge this PR?

I would lean toward the library, as it also provides some good to have functionality

coolsoftwaretyler:
@chakrihacker - I think the third-party approach makes sense, and since it helps avoid breakage, I'm guessing that's why [Jamon liked the idea a few years ago](https://github.com/mobxjs/mobx-state-tree/pull/691#issuecomment-1054591610).

I'd also like to close or move forward with this PR to clean it up, but I don't want us to lose the context quite yet. I will bring this up at the next maintainer meeting.

coolsoftwaretyler:
Hey folks, we are going to close this issue. Not because we don't want to include it, but because: 

1. The PR is very stale as it is
2. There is an existing third party solution that solves the problem, so this isn't the only existing path forward
3. We are going to be making a big formatting diff in the coming weeks that will make rebasing old PRs very challenging

I think we're definitely interested in adding something like this, but going to hold off now and close this out to clean up the GitHub a little.

If anyone really wants this merged in the next 2-3 weeks, feel free to open a new PR with a similar implementation.

Thanks!

PR #692, opened 2018-03-09
Improve resolving paths

### LINKED ISSUE #686: tryResolve does not work as advertised
I have a model with a reference to another model where not all ids can be resolved to an actual instance of the referenced model.

Firstly, I would expect a reference under types.maybe() to fail silently should it fail to reconcile an identifier to an actual instance, but that is just my opinion.

Be that as it may, I have tried using tryResolve in order to avoid throwing an exception, and while it definitely succeeds when an identifier is reconciled; when it is not, it still throws the same error as when trying to access the reference property directly, thus completely missing the point of tryResolve according to documentation.

mweststrate:
Should fix #686

### DIFF
--- a/packages/mobx-state-tree/src/core/node/node-utils.ts
+++ b/packages/mobx-state-tree/src/core/node/node-utils.ts
@@ -1,4 +1,4 @@
-import { IType, fail, ObjectNode, splitJsonPath, joinJsonPath } from "../../internal"
+import { IType, fail, ObjectNode, splitJsonPath, joinJsonPath, ScalarNode } from "../../internal"
 
 export enum NodeLifeCycle {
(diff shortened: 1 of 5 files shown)


packages/mobx-state-tree/src/core/node/node-utils.ts
```diff
@@ -118,32 +118,54 @@ export function resolveNodeByPathParts(
     pathParts: string[],
     failIfResolveFails: boolean = true
 ): INode | undefined {
+    debugger
```
mslenc: .. :)

mweststrate: will fix!

PR #711, opened 2018-03-17
fix codesandbox wihtin middleware docs

### LINKED ISSUE #496: [Bug] Middleware SandBox example is actually not working
The [SandBox example](https://codesandbox.io/s/mQrqy8j73) provided on the [Middleware page](https://github.com/mobxjs/mobx-state-tree/blob/feature/separate-middleware/docs/middleware.md) is not working properly. New Todos should be created with an uppercased title but they are not.
Even though arguments are modified before passing the call to the next middleware, 
'addTodo' action is still called with original arguments not the modified ones.

robinfehr:
fixes #496

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -11,7 +11,7 @@ MST ships with a small set of [pre-built / example middlewares](../packages/mst-
 
 
 Custom middleware example:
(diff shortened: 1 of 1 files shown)


mweststrate:
Yay

PR #721, opened 2018-03-21
Do not initialize store twice

ezrubaal:
In the MST example TodoMVC the store was initialized twice, once here and once in the `createTodoStore` function.


mweststrate:
Nice catch :)

PR #724, opened 2018-03-22
Export missing API functions joinJsonPath and splitJsonPath.

jjrv:
These functions are mentioned in the readme but not actually exported.


mweststrate:
Thanks!


Next time run build-docs to generate the api docs again and make the above tests succeed as well. Will merge now and fix afterwards

PR #745, opened 2018-04-02
getSnapshot without postProcess

robinfehr:
introduce a flag to optionally disable the postProcessSnapshot.

```
getSnapshot(node, applyPostProcess)
```


robinfehr:
we might run into problems with the `onSnapshot` since the snapshot might change. needs to be tested.

robinfehr:
comment above verified. code needs to be changed so that we have a raw computed snapshot.

mweststrate:
Looking good, merging!

PR #753, opened 2018-04-04
WIP: disposer fuction returned from addDisposer

robinfehr:
```js
const disposer = addDisposer(reaction(
  addDisposer(self, autoSaveDisposer)
    () => getSnapshot(self),
    snapshot => sendSnapshotToServerSomehow(snapshot)
))
```

makes reactions abort-able within view components (e.g. unmount) and on node destruction at the same time.

### DIFF
--- a/packages/mobx-state-tree/src/core/mst-operations.ts
+++ b/packages/mobx-state-tree/src/core/mst-operations.ts
@@ -266,6 +266,7 @@ export function getSnapshot<S = any>(target: ISnapshottable<S>): S
 /**
  * Calculates a snapshot from the given model instance. The snapshot will always reflect the latest state but use
  * structural sharing where possible. Doesn't require MobX transactions to be completed.
+ * By default postProcessSnapshot hooks are applied to the snapshot.
  *
  * @export
  * @param {Object} target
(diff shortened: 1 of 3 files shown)


robinfehr:
back to WIP - might wan't to add a fn so that multiple nodes can be added as target so also an autorun / reaction reading multiple nodes,  could be used

mweststrate:
@robinfehr I didn't get the gist of this PR entirely, but sounds like it can be closed ?

robinfehr:
I'm actually not sure the following is correct for the latest version - will update soon and check.

`reading the comment above, I can't figure out myself what I wanted to describe 👍😆`
 
but the idea had been to be able to remove the addDisposer because there are situations we need that.

i.e. since for the second param of the addDisposer we could add a reaction that would be disposed by the addDisposer if the node dies. If the addDisposer is setup within e.g. a react component tough we'd want to dispose the addDisposer too if the component unmounts since the reaction could be needed for this component only.

e.g. 
```js
componentDidMount() {
    const reaction1 = reaction(
        () => {
            const { view } = someNode;
            const { width, height } = view;
            return { width, height };
        },
        ({width, height}) => {
            this.doSmth(width, height);
        }, { name:'view' }
    );
    this.reactions.push(reaction1);

    const disposerOfAddDisposer = addDisposer(someNode, reaction1);
    this.disposers.push(disposerOfAddDisposer);
}

componentWillUnmount() {
    // remove all disposers
    this.disposers.forEach(d => d());
    // remove all the reactions
    this.reactions.forEach(d => d());
}
```

we might be able to solve that more elegantly. let's see.

xaviergonz:
Closing since returning the disposer was done in #1059 (without the disposing from the react component since mobx-react now has it's own personal disposeOnUnmount)

Issue #477, opened 2017-10-19
getParentOfType(node, parentType)


xaviergonz:
Will try to get the first parent found of a given node that matches a given type, or undefined if none found.

Basically I'm using getParent with different depths to achieve the same effect at the moment, but I'm starting to loose track :) (plus that implementation would be more resiliant to model changes)

mweststrate:
Sounds useful Interested in attempting a PR? `mst-operations.ts` should be
a fine place to add it to

Op do 19 okt. 2017 om 20:36 schreef Javier Gonzalez <
notifications@github.com>:

xaviergonz:
Right now I'm using the following code:
```js
export function getParentOfType<S, M>(target: IStateTreeNode, type: IType<S, M>): M | undefined {
  let currentNode = target;
  while (hasParent(currentNode)) {
    currentNode = getParent(currentNode);
    if (getType(currentNode) === type) {
      return currentNode as M;
    }
  }
  return undefined;
}
```

But I have a few doubts. What does getParent do? the docs say:
 * Returns the immediate parent of this object, or null.
but then the actual method does:
 return fail(`Failed to find the parent of ${getStateTreeNode(target)} at depth ${depth}`)

so, should getParent be changed so it returns null? should its docs be changed?

should the new method return null if not found? undefined? throw?
if throw, should there be a "hasParentOfType" equivalent?

the-noob:
Just hit this issue with something like 
```JavaScript
const Model = types.model('Model', {
  name: types.string,
})
  .view((self) => ({
    get builder() {
       // return getParent(self); - this returns the `models` Map
       return getParent(getParent(self)); // - returns the Builder model
    }
  ));

const Builder = types.model('Builder', {
  models: types.map(Model)
});
```
Indeed the Map is the first parent but not exactly what I've expected.

xaviergonz:
```js
export function getParentOfType<S, M>(target: IStateTreeNode, type: IType<S, M>): M | undefined {
  let currentNode = target;
  while (hasParent(currentNode)) {
    currentNode = getParent(currentNode);
    if (getType(currentNode) === type) {
      return currentNode as M;
    }
  }
  return undefined;
}
```

returns undefined if not found or the node if found. I'm just waiting for @mweststrate or @mattiamanzati to get me a hint on the desired semantics to adapt it and make a pull request

mweststrate:
[...]

PR #767, opened 2018-04-13
Get parent of type

k-g-a:
Implements [#477](https://github.com/mobxjs/mobx-state-tree/issues/477)
Unfortunately I couldn't generate API.md properly. It converts 
```
-   [addDisposer](#adddisposer)
```
into 
```
-   [addDisposer][1]
```
And adds 
```
[1]: #adddisposer
```
at the end of API.md

This leads to failure of two api tests.


mweststrate:
@k-g-a thanks! no worries about the docs, we can look into that. The tool is a bit fragile indeed :-/

PR #784, opened 2018-04-25
Use Array.from with .values() in React iteration

luisherranz:
I've also fixed and updated the codesandbox to the latest version of React, Mobx and Mobx-State-Tree.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -294,7 +294,7 @@ More details can be found on the mobx-react package documentation, but keep in m
 ```javascript
 const App = observer(props => <div>
     <button onClick={e => props.store.addTodo(randomId(), 'New Task')}>Add Task</button>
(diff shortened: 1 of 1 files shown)


robinfehr:
maybe we should directly introduce `values(x)` what do you think?

luisherranz:
So... `import { values } from 'mobx'` or `import { values } from 'mobx-state-tree'`?

robinfehr:
what I've meant was more if we already should show the newcomers to the `values(x)` from mobx instead of making them use the `Array.from` option. 

adding the same to mst that is in mobx already...idk, I think the users have to get used to use both anyway.

mweststrate:
Just from mobx

Op do 26 apr. 2018 12:14 schreef Rob <notifications@github.com>:

luisherranz:
done!

mweststrate:
Thanks!

Issue #781, opened 2018-04-24
Disable console.errors in production mode when typechecking


nathanbirrell:
**Scenario**

* Use `typecheck` to validate a model at runtime (with `NODE_ENV=production`)
* Cause an exception in that check using invalid data

**Expected**

Error is thrown by MST, so that I can catch it a level above, but **no console.error gets logged** in production.

**Actual**

Error is thrown by MST, console.error gets logged.

Do we really want to log console errors in production?

mweststrate:
I think that is a bug, can you create a test case showing where a type
error gets logged even when catched?

Op di 24 apr. 2018 08:49 schreef Nathan Birrell <notifications@github.com>:

nathanbirrell:
Sure @mweststrate - take a look at https://github.com/mobxjs/mobx-state-tree/pull/785

mweststrate:
This has been released a while ago

PR #785, opened 2018-04-25
Add a test to check that console errors don't get logged when error is caught

nathanbirrell:
As per request in #781


mweststrate:
Thanks! Will investigate

PR #789, opened 2018-05-01
Fix error arg name in async code example

johnrees:


mweststrate:
Thanks for noticing!

johnrees:
Thanks for mobx!

PR #797, opened 2018-05-04
fix issue on map with number identifier that must be cast as string

knalinne:
Hello,

When using map with a model that has a number identifier, the identifier must be cast as string when using 'get' function.

As identifiers can be number, map should accept number too.

In this PR, the key type is changed from string to string | number.
The get function is also changed to force a string comparison when getting a value.
I've also added a test for this case.

### DIFF
--- a/packages/mobx-state-tree/src/types/complex-types/map.ts
+++ b/packages/mobx-state-tree/src/types/complex-types/map.ts
@@ -39,7 +39,7 @@ interface IMapFactoryConfig {
     isMapFactory: true
 }
 
-export interface IExtendedObservableMap<T> extends ObservableMap<string, T> {
+export interface IExtendedObservableMap<T> extends ObservableMap<string | number, T> {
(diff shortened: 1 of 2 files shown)


mweststrate:
Thanks for the PR. I have to investigate this a bit more, but I think number based maps will always be troublesome until #808 is implemented

mweststrate:
#904 will make the behavior consistent, although different from what proposed; identifiers can still be strings or numbers, and are now consistenly serialized as is. However, map keys, snapshot indices, json path parts etc are now always strings, and a lot of inconsistencies in the normalization have been removed.

mweststrate:
Merged into #904, as normalizing on read as well makes sense, since we do it on write. 

Didn't add the typings, as this is now how maps are supposed to be used. Might remove again in the future if the performance overhead is too big

mweststrate:
Extended on this a bit, `has`, `set` and `delete` will now also normalize keys for consistent behavior

PR #821, opened 2018-05-16
Expose Snapshot type in package type declaration.

dsabanin:
Right now if for some reason you want to type output of types.model(...), you can't because it returns IModelType, but Snapshot is unavailable so this can't work:

```typescript
const myModel = IModelType<Snapshot<T>, T> = types.model<T>('MyModel', {})
```


mweststrate:
Fix released as 2.1.0

PR #842, opened 2018-06-01
#634 fix optional frozen to return default value if undefined

### LINKED ISSUE #634: frozen and applySnapshot [bug]
When use  applySnapshot it set a optional t.frozen to undefined...
Is this expected ?

```javascript
const Model = types.model('model', {
  field: t.optional(t.frozen, {})
})

const m = Model.create({})
m.field == {} 

applySnapshot(m, {})
m.field == undefined
```

SirbyAlive:
Fixes #634, optional frozen were undefined after applySnapshot()

Issue was the undefined is a valid type for frozen so was accepted as it by optional type and not falling back to the default value.


mweststrate:
Thanks!

mweststrate:
Released in 2.1.0. Thanks!

SirbyAlive:
Awesome, you are most welcome, thank YOU!

Issue #810, opened 2018-05-10
RFC: initialize instances upon usage


mattiamanzati:
Right now MST can have performance issues when re-hydrating huge arrays or large trees.
This happens because MST converts snapshots to their observable counterpart upon snapshot application, and that may take a lot of resources.

So why not we create first the internal Node from the initial snapshot and eventually create the observable instance only when the value is first referenced?

e.g. If we have a huge list of Todos, the observable array instance will be created, then the react-virtualized component will access the first 20 items, so only that 20 todo instances will be accessed and their observable counterpart will be created, next we scroll the view and the other instances will be created while scrolling.

Only change would be the afterCreate hook firing upon value conversion (so afterAttach may fire first).
But we can alternatively keep the current execution order and value will be converted only if the self is actually used by the afterCreate hook.

This change would require "Node" to store the initialSnapshot and return that into "@computed get snapshot" if the value is not initialized yet.

This would also solve situations where validate runs twice.

Looking forward comments and wild ideas! :)

Bnaya:
Will the data validate upfront, or also lazy?
Because if you will get data validation error lazy, its very unexpected behavior

mattiamanzati:
No, data validation will be performed synchronusly upon Node creation :)

mweststrate:
..so, lazily ;-)?

Op do 10 mei 2018 om 12:11 schreef Mattia Manzati <notifications@github.com

mattiamanzati:
Whops, I was not clear enough :)
The entire node tree will be validated and then created (as happens right now), only the observable instance will be delayed to the first access :)

mattiamanzati:
Here's some pseudocode, basically you end up creating just the entire Node tree instead of the whole observable tree :)
```typescript
class Node{

  constructor(public readonly type: IType<any, any>, public readonly initialSnapshot: any, public parent: Node){
    typecheck(type, initialSnapshot)

    // create child nodes here
    childNodes = type.initializeChild(initialSnapshot)
  }

  get snapshot(){
    return this.hasInitialized ? this.type.getSnapshot(this) : this.initialSnapshot
  }

  get value(){
    if(!this.hasInitialized) this.storedValue = this.createNewInstance()
    return this.storedValue
  }
}
```

Bnaya:
So the validation is not the main CPU intensive task? do you have benchmarks on it?

mattiamanzati:
[...]

Issue #801, opened 2018-05-07
Action events (from actions) don't provide object. Bug?


belozer:
```js
import { spy } from 'mobx';
import { types } from 'mobx-state-tree';

const Todo = types.model("Todo", {
  title: types.string,
  done: false
}).actions(self => ({
  toggle() {
    self.done = !self.done
  }
}))

const Store = types.model("Store", {
  todos: types.array(Todo)
}).views(self => ({
  completed() { return self.todos.filter(todo => todo.done) }
})).actions(self => ({
  addTodo(title) {
    self.todos.push({ title })
  }
}))

// create an instance from a snapshot
const store = Store.create({
  todos: [{
    title: "Get coffee"
  }]
})

spy((e) => {
  if(e.type === 'action') console.log(e);
});

store.addTodo('Test todo');
store.todos[0].toggle();
```

Log prints
```
Object {type: "action", name: "create", object: ModelType$$1, arguments: Array[1], spyReportStart: true}
Object {type: "action", name: "<unnamed action>", object: null, arguments: Array[1], spyReportStart: true}
Object {type: "action", name: "applySnapshot", object: ArrayType$$1, arguments: Array[2], spyReportStart: true}
Object {type: "action", name: "addTodo", object: null, arguments: Array[1], spyReportStart: true}
Object {type: "action", name: "addTodo", object: null, arguments: Array[1], spyReportStart: true}
Object {type: "action", name: "Reaction@57", object: undefined, arguments: Array[2], spyReportStart: true}
Object {type: "action", name: "Reaction@57", object: undefined, arguments: Array[2], spyReportStart: true}
Object {type: "action", name: "Reaction@55", object: undefined, arguments: Array[2], spyReportStart: true}
Object {type: "action", name: "Reaction@55", object: undefined, arguments: Array[2], spyReportStart: true}
Object {type: "action", name: "toggle", object: null, arguments: Array[0], spyReportStart: true}
Object {type: "action", name: "toggle", object: null, arguments: Array[0], spyReportStart: true}
Object {type: "action", name: "Reaction@57", object: undefined, arguments: Array[2], spyReportStart: true}
Object {type: "action", name: "Reaction@57", object: undefined, arguments: Array[2], spyReportStart: true}
Object {type: "action", name: "Reaction@55", object: undefined, arguments: Array[2], spyReportStart: true}
Object {type: "action", name: "Reaction@55", object: undefined, arguments: Array[2], spyReportStart: true}
[...]

PR #845, opened 2018-06-04
Initialize instances upon usage (issue #810)

k-g-a:
Good day!
This pull request implements [#801](https://github.com/mobxjs/mobx-state-tree/issues/810)

I'll try to hightlight all the implementation details in this comment, but any questions and suggestions are greatly anticipated.

So we begin with **object-node.ts**:
- added private non-observable property ```observableInstanceCreated```
- added private ```@action createObservableInstance``` method, which is in most cases called from ```get value()```; this one is responsible for creating observable instance of a node and also changes ```observableInstanceCreated``` in process
- ```get value``` was split into simple getter and ```@somputed get _value()```

Firts of all, ```observableInstanceCreated``` is non-observable because ```createObservableInstance``` (which changes it) may be called from ```@compued``` and we can not control this. An example is _"it should compose factories with computed properties"_ test: ```views.area``` tries to get ```props``` (which is ```types.map``` and was not yet initialized during ```ComposedFactory.create```). 
Second, by the same reason, ```get value``` is not computed either. For example in _"Parent property has value during child's afterAttach() event"_ we'll get same "Computed values are not allowed to cause side effects by changing observables" error: ```instantiate``` -> ```createNode(...).value``` -> ```observableInstanceCreated === false``` -> ```createObservableInstance()``` -> ```finalizeCreation()``` -> ```fireHook("afterAttach")``` -> user code modifies any property -> error

- ```applyPatches```/```applySnapshot``` are now defined on object node from the beginning as:
```typescript
    applyPatches(patches: IJsonPatch[]): void {
        if (!this._observableInstanceCreated) this.createObservableInstance()
        this.applyPatches(patches)
    }
    applySnapshot(snapshot: any): void {
        if (!this._observableInstanceCreated) this.createObservableInstance()
        this.applySnapshot(snapshot)
    }
```
Becuase there is one case in which we may try to applySnapshot on not-yet-initialized-node: root model's ```applySnapshot()``` -> ```forAllProps()``` -> ```willChange()``` -> ```node.getChildNode(change.name)``` (as ```current```) -> ```type.reconcile(current, ...)```  -> ```current.applySnapshot(newValue)```. As ```current``` was obtained via ```getChildNode``` - it wasn't unboxed, thus ```get value()``` was not acessed and ```createObservableInstance()``` was not called => ```current.applySnapshot``` was not yet defined (as it riles on storedValue which has not even been created).
The other solution would be to make ```observableInstanceCreated```/```createObservableInstance()``` public and check/call from within ```type.reconcile```. Actually, this was my first solution. But it seems to me that keeping all the "lazy-logic" isolated in one place is better idea and worths defining two little methods on ObjectNode's prototype. To be discussed.

- ```createObservableInstance``` uses ```getAtom```/```trackAndCompute``` for snapshot to invalidate it

This is neccessary, because ```observableInstanceCreated``` is not ```@observable```, but we must invalidate ```@computed snapshot``` after initializetion. Otherwise we'll continue to get ```initialSnapshot```. This is prooved by many test, for example _"it should apply patches"_. 
Also this does not incure any performance penalty as ```snapshot``` is accessed anyway by the ```snapshotDisposer``` reaction two lines of code below :)

- added private ```_childNodes``` which are filled in ```constructor``` and used later on for observable instance creation.

This is quite straightforward. The only thing to mention here is that type's ```createNewInstance()```/```finalizeNewinstance()``` now recieve childNodes instead of snapshot.

- ```identifierAttribute``` is now filled from the very beginning (in ```constructor```)
- ```get identifier()``` now reads ```identifierAttribute``` from ```initialSnapshot``` as it can not be changed during lifecycle of node by convention

I'll elaborate on those two points later.

Next important changes consider **model.ts**: 
- ```postProcessSnapshot``` now must be declared as type's method and is removed from hooks; warning added.

This is required because user may demand snapshot of non-initialized node and it must be postprocessed. Alternatively, we could initialize node at this point, but it seems pointless, adds complexity and reduces runtime performance. Moreover ```postProcessSnapshot``` is a counterpart of ```preProcessSnapshot``` and it's logical for them to be defined on the same level. 
Composition during ```cloneAndEnhance``` is preserved.
The only thing that embarasses me is that at ```object-node._getInitialSnapshot()``` I implicitly check for ```this.type instanceof ModelType```. I'd prefer object-node to avoid having any specific knowledge about underlying ```type```. It could be changed to ```typeof (this.type as any).applySnapshotPostProcessor === "function"```. But I actually perform ```instanceof``` check second time: for ```identifierAttribute``` as it can only present on ModelType's ```$treenode```. To be discussed.

- ```ModelType.constructor``` now collects it's ```OptionalValue``` children, as those are used to enrich initialSnapshot with defaults further
- as it already has to process all the properties, we use this moment to find ```identifierAttribute``` among them

```constructor``` seems to be best place, as properties can not be changed during lifecycle of type so there is no need to recollect them on every instantiation. 
Also it seems usefull to determine ```identifierAttribute``` upfront, as it simplifies ```map.processIdentifier```/```map.put``` and guarantees that we know identifier on lazy nodes.
Those two changes were not intended in initial proposal by @mattiamanzati, so we should discuss whether to include them or not. 

- ```initializeChildNodes()``` method was added, which actually instantiates child nodes (those are lazy in turn)
- ```finalizeNewInstance()``` now recives previously created nodes and just copies those to instance

Second point slightly inconsistent with ```array```/```map``` types. 
Those get child nodes at ```createNewInstance()``` and use them directly during ```instance = observable.array(childNodes, ...)```/```instance = observable.map(childNodes, ...)```.  Later on, during ```finalizeNewInstance()``` only ```_interceptReads```/```intercept```/```observe``` are invoked on instance.
But model is a little more complicated: intercepts reads of every property, initializes views/actions etc. - so I decided to keep those in one place and not to spread across two methods. 
To my mind we can actually unite those two in one for all types. To be discussed.

Next step is **array.ts**/**map.ts**:

- the most important part here is that ```finalizeNewInstance``` does not use ```applySnapshot``` to fullfill self anymore

The case here is about actions: ```ObjectNode.applySnapshot``` is actually an action. Previously this action was always invoked during creation and could not be intercepted by any middleware (as we can attach middleware only to instance). But now ```finalizeNewInstance``` may be called far behind this point in time and those changes will be detected by middleware. Such behaviour was never intended. 
Using ```this.applySnapshot``` (instead of ```node.applySnapshot```) resolves this problem, but other bugs occur: mainly during reconcilation, as those nodes are craeted and we must just setparent, but their _parent is already the same node, which is ```fail()```, but we could compare ```subpaths```, but it's complex and so on, and so forth... After digging too deep into this codepath I made decision to keep things simple and use existing ```childNodes``` right during ```createNewInstance()```, before any interceptions/observers are attached. This seems to do it's work well.

- ```map.ts``` has one additional important change comparing to ```array.ts```: it now can and will ```_determineIdentifierMode()``` upfront

Ideal path would be to make it once in ```constructor```, but unfortunatelly ```LateType``` may cause an issue here, which is quite tricky.
_"it should apply deep patches to objects"_ test failed with ```ReferenceError``` as ```NodeObject``` is not defined during types construction. So for ```LateType``` i used try-catch and left ```identifierMode = UNKNOWN```. Later on, in ```Map.instantiate()``` I check, whether  ```identifierMode``` is still ```UNKNOWN``` and run ```_determineIdentifierMode()``` once again, if needed. This was not ideally obvious, but works as expected. 
But I was really surprised, when test passing in webstorm (which uses jest@20.0.4 from main package.json) failed in console (which, in turn, uses jest@22.4.2 from packages/mobx-state-tree/package.json). It turned out that in es5 environments (where ```const``` converted to ```var```) LateType.subType.definition() returned ```undefined``` instead of throwing ReferenceError, so I had to handle this manually (we can not be certain about user's env):
```typescript
    get subType(): IType<S, T> {
        if (this._subType === null) {
            // If called too early in es5 environment, this won't throw, but return undefined
            // see "it should apply deep patches to maps" test for example
            const definition = this.definition()
            if (typeof definition !== "object") fail("Late type definition should return object")
            this._subType = definition
        }
        return this._subType
    }
```
All of this is quite complex and counterintuitive, so i'm looking forward to any suggestions. Nevertheless doing this upfront, as early as possible, and as less times as possible seems like a goood idea to mee, so i'd like to keep those changes.

Next thing is **union.ts**:
- i've changed subtypes order for the most common case of two types without ```dispatcher``` to direct instead of reversed

For now all the descriptions and errors are showing correct order of subtypes, same as it was defined by user. Inverse order did especially noticable matter for ```map``` error about "unions should have same identifierAttribute". Had to inverse types for ```maybe``` and correct error messages for some tests.

Finally, some **tests** had to be corrected:
- actions.ts _"middleware events are correct"_: as less actions are run now upon cration, id/parentId/rootId have changed
- hooks.ts _"it should trigger lifecycle hooks"_: much less hooks are called now as not all child nodes are really instantiated
- map.ts _"map expects regular identifiers"_: "The objects in a map should all have the same identifier attribute" error is now thrown right upon declaration, which is **breaking change**, but seems right as it would happen later in runtime; i could not think of a case where one desire to declare union of two maps with different identifiers - this is error-prone aproach
- map.ts _"map can resolve late identifiers"_ test added
- optimizations.ts _"it should avoid processing patch if is exactly the current one in reconcile"_ test rewritten to ensure that reconciliation works on initial values and initializing instance does not cause snapshot invalidation
- parent-properties.ts _"Parent property does not have value during child's afterCreate() event"_: changed to "Parent property have value during child's afterCreate() event" as this is new reality (and **breaking change**)
- changed hooks.ts and mst-middlewares tests to define ```postProcessSnapshot``` directly on model type as this is new reality (and **breaking change**)

I hope I covered all the complex changes and tried to explain reasons behind them. Changes not mentioned here seem to be self-explanatory, but in case something left unclear, I'll be happy to answer any questions.


k-g-a:
Ops, forgot to mention few unrelated to original issue things: 
- ```middlewares```/```snapshotSubscribers```/```patchSubscribers``` are now ```null``` upon ObjectNode's creation and get initialized to arrays at first access; idea taken from [this comment](https://github.com/mobxjs/mobx-state-tree/issues/440#issuecomment-336137340) by @mweststrate
- all packages besides mobx-state-tree itself and mst-middlewares now have "tests": "exit 0" in their package.json to avoid "No test script specified" error locally. This was an issue at least for yarn@1.3.2 on windows.

k-g-a:
More news :) 
I have another branch on top of this one, which focuses on removing unnecessary closures. Changes mainly consider removing ```.bind()``` and making willChange/didChange able to live on prototype + removing redundant arrow-function wrappers around them. 
Those chanegs do not alter any user experience, but provide closures amoun reduction of about 10%. This in turn results into 3-6% less memory footprint, depending on model's count and their structure.
This is a drop in the sea, but little step is better than none :)

mattiamanzati:
@k-g-a I want to personally thank you for all this great work!!

dnakov:
I’m running this in production now and haven’t had any issues. Load time went down from 11s to 2.5s!
Well done @k-g-a

mweststrate:
Awesome work @k-g-a! With MobX 5 out of the way I want to focus a lot more
on MST from next week an and merge and push a lot of MR's out of the door
:).

And also think about setting this up more scaleable so that I don't become
the bottleneck of publishing awesome improvements like these!

Op do 7 jun. 2018 om 15:18 schreef Daniel Nakov <notifications@github.com>:

packages/mobx-state-tree/src/core/node/object-node.ts
```diff
@@ -20,73 +19,107 @@ import {
...
+        this.applyPatches(patches)
+    }
+    applySnapshot(snapshot: any): void {
+        if (!this._observableInstanceCreated) this._createObservableInstance()
+        this.applySnapshot(snapshot)
+    }
 
-    applySnapshot: (snapshot: any) => void
+    private _observableInstanceCreated: boolean = false
+    private readonly _childNodes: IChildNodesMap | null = null
```
mweststrate: I was wondering about this one, if the instance is not yet created, that means that all children are still (in serialized form probably) in that initial snapshot value, so none of the children need or should be constructed either?

k-g-a: If I understood the question correctly, this was the original idea to "end up creating just the **entire Node tree** instead of the whole observable tree" ([comment 1](https://github.com/mobxjs/mobx-state-tree/issues/810#issuecomment-388020137), [comment 2 ](https://github.com/mobxjs/mobx-state-tree/issues/810#issuecomment-390393224));

Initializing children as ObjectNodes (without observable counterpart) is needed to:
- fill identifierCache - otherwise references won't work;
- getting deeply nested children for ```snapshot```: root is fully initialized => ```.snapshot``` calls ```getActualSnapshot()``` => it goes to ```type.getSnapshot()``` => for any complex type it goes through ```this.getChildNode(node, name)``` => if child is not yet fully initiazed, it goes to ```this._childNodes![subpath]```
- finalizeCreation() uses getChildren() to fire hooks, which is expected to return nodes, not just objects

I could forget a point or two, becuase I had no chance to peek into the code.

mweststrate: You are correct, thanks!

k-g-a:
As I'm quite new to opensource contributions: should I resolve conflicts by github means, just merge current master to my branch or try to rebase it over current master?

mweststrate:
With github is tricky, merging current master or rebase is both fine,
whatever suits you best. Merge is most probably easier.

Op di 19 jun. 2018 om 07:43 schreef Gleb <notifications@github.com>:

mweststrate:
@k-g-a let me know when rebasing / merging succeeded or if you did run into trouble. I want to start the MST 3 branch based / with this one :)

k-g-a:
@mweststrate haven't had time to merge. I'll try to make it today.

k-g-a:
Should I also merge changes mentioned at [this comment](https://github.com/mobxjs/mobx-state-tree/pull/845#issuecomment-394952700)? Those are at separate branch now.

mweststrate:
@k-g-a  Yeah, feel free to! Or latter on as additional branch, whatever is more easy to manage for you

mweststrate:
P.S.: I just fixed the master branch, so if you that is merged again in this branch, the build will become green :)

k-g-a:
@mweststrate, great news! merged to all my PRs

bourquep:
I'd like to test this branch, but I am unsure of how to add a reference to it in my `package.json` because the repo is a yarn workspace.

This: 

```
"mobx-state-tree": "git+https://github.com/k-g-a/mobx-state-tree.git#deferred-observable-creation",
```

Leads to this:

```
"Can't add \"mobx-state-tree\": invalid package version undefined.
```

bourquep:
Got it working by cloning the `k-g-a` repo, building, and referencing that build on my local machine in package.json.

k-g-a:
More news about laziness. 
Currently every ObjectNode creates following ```reaction``` upon observable instance initialization:
```javascript
    const snapshotDisposer = reaction(
            () => this.snapshot,
            snapshot => this.emitSnapshot(snapshot),
            snapshotReactionOptions
        )
        this.addDisposer(snapshotDisposer)
```
But it seems like those are needed only in two cases:
1) when we explicitly call onSnapshot(myInstance, myHandler) - to emit those snapshots
2) for root node, as something must observe ```@computed``` ```snapshot``` for computatuins caching to work properly (unobserverd computeds are recalculated on every access) and this reaction does so.

So I made following changes to ObjectNode:
```typescript
    private _hasSnapshotReaction = false
    ...
    private _addSnapshotReaction() {
        if (!this._hasSnapshotReaction) {
            const snapshotDisposer = reaction(
                () => this.snapshot,
                snapshot => this.emitSnapshot(snapshot),
                snapshotReactionOptions
            )
            this.addDisposer(snapshotDisposer)
            this._hasSnapshotReaction = true
        }
    }
    @action
    private _createObservableInstance() {
        ...
        invalidateComputed(this, "snapshot")
        if (this.isRoot) this._addSnapshotReaction()
        this.finalizeCreation()
        ...
    }

    public onSnapshot(onChange: (snapshot: any) => void): IDisposer {
        this._addSnapshotReaction()
        if (!this._snapshotSubscribers) this._snapshotSubscribers = []
        return registerEventHandler(this._snapshotSubscribers, onChange)
    }        
```

I made it as a quick experiment right on top of this branch, but it actually can be done in master, as these changes come into play after observable instance creation. 
Here is comparison for my favorite example of 10 parents * 100 children * 10 grandchildren (prod):
1) before
01-before
2) after
02-after

It's 10.7Mb out of 70.1Mb for my particualr example which is close to 15%. Of course diff percentage will be smaller for larger models (my models only have 3-4 primiteive types and couple of actions).

All tests are passing. I've also tested manually that adding ```onSnapshot``` for instance's child nodes works.

@mweststrate, @mattiamanzati two main questions:
1) did I miss something? 
2) should I commit it here or create separate PR over master? I can prodive comparison for master + add tests about onSnapshot working for cases like ```onSnapshot(myInstance.very.deep.node, myHandler)```

packages/mobx-state-tree/src/types/complex-types/model.ts
```diff
@@ -138,18 +150,49 @@ export class ModelType<S, T> extends ComplexType<S, T> implements IModelType<S,
         // ensures that any default value gets converted to its related type
         this.properties = toPropertiesObject(this.properties)
         freeze(this.properties) // make sure nobody messes with it
-    }
+        this.propertyNames = Object.keys(this.properties)
+        this._preBootModel()
+    }
+
+    private _preBootModel() {
+        let optionalFound = false
```
mweststrate: would it matter to just store all properties in the optionalMap? (and by that not having optional map at all but just look at the properties if something from the snapshot?)

packages/mobx-state-tree/src/utils.ts
```diff
@@ -176,6 +176,11 @@ export function argsToArray(args: IArguments): any[] {
     return res
 }
 
+export function invalidateComputed(target: any, propName: string) {
+    const atom = getAtom(target, propName) as any
+    atom.trackAndCompute()
```
mweststrate: I think this would ignore normal mobx transactions? Would there be potential risks in that?

packages/mobx-state-tree/src/types/complex-types/map.ts
```diff
@@ -102,23 +125,55 @@ export class MapType<S, T> extends ComplexType<{ [key: string]: S }, IExtendedOb
         )
     }
 
+    private _determineIdentifierMode() {
```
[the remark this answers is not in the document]
mweststrate: Ah, that seems to be covered in the tests below.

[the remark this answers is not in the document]
k-g-a: Yep, recursive late types will be resolved once - on first instantiation. Other cases can handle it right during construction.

mweststrate:
Merged! See #899

packages/mobx-state-tree/src/types/complex-types/model.ts
k-g-a: If I ubderstand correctly, you suggest to replace 
```typescript
        if (processed && this.optionalChildren) {
            const optionalChildren = this.optionalChildren
            Object.keys(optionalChildren).forEach(name => {
                if (typeof processed[name] === "undefined") {
                    processed[name] = optionalChildren[name].getDefaultValueSnapshot()
                }
            })
        }
```
to
```typescript
        if (processed) {
            this.forAllProps((name, type) => {
                if (typeof processed[name] === "undefined") {
                    const optional = tryGetOptional(type)
                    if (optional) {
                        processed[name] = optional.getDefaultValueSnapshot()
                    }
                }
            })
        }
``` 
at ```applySnapshotPreProcessor```. By doing so we can avoid introducing ```this.optionalChildren
```.

It will be a little bit more cpu-intensive, as ```_preBootModel``` happens once for type and ```applySnapshotPreProcessor``` happens on every snapshot application. 
But on the other hand code will be clearer and couple of ```typeof``` checks do not seem heavy. Revisiting this code also made me think that ```types.late(() => types.optional(...))) case is not covered. It's quite strange and rare, but possible.
I will add changes within couple of hours ;)

mweststrate:
Yeah exactly, since the typechecking already happened, and checking if a
field is undefined is pretty cheap, I think this can simplify the code
without really affecting performance. (P.s. do you have a twitter handle?)

Op do 28 jun. 2018 om 09:39 schreef Gleb <notifications@github.com>:

mweststrate:
Also, please branch from (or push to) the mst3 branch :)

Op do 28 jun. 2018 om 09:42 schreef Michel Weststrate <mweststrate@gmail.com

packages/mobx-state-tree/src/utils.ts
k-g-a: To be honest, this part is terra-incognita for me. As I wrote before "```observableInstanceCreated``` is not ```@observable```, but we must invalidate ```@computed``` ```snapshot``` after initializetion".
I did not find any documented way to do so and was stuck around this problem for a while. 
I ended up exploring mobx TS type definitions for something like ```myComputed.forceRecompute()``` and found such solution (almost by accident).
I used it to invalidate ```snapshot``` at this branch and ```path``` at [#872](https://github.com/mobxjs/mobx-state-tree/pull/872) successfully. 
If you give me a clue about what to check, I can explore further ;)

mweststrate:
Ok, no problem, I'll test with it and see if I run into any trouble :)

Op do 28 jun. 2018 om 10:08 schreef Gleb <notifications@github.com>:

mweststrate:
For all clarity; this PR was merged into the mst3 branch :). Due to my lack of github skills this PR appears as being closed, but it ain't!

mweststrate:
Yes, it's merged into the mst3 branch 😊

Op do 28 jun. 2018 22:26 schreef Chema <notifications@github.com>:

PR #847, opened 2018-06-05
fix(type): parameter 'fn' in undoManager.stopGroup should be optional

abruzzihraig:
As the usage described in https://github.com/mobxjs/mobx-state-tree/blame/master/packages/mst-middlewares/README.md#L369 and the condition in https://github.com/mobxjs/mobx-state-tree/blob/878d7f312d03276304d20af9dc055837666dbc6f/packages/mst-middlewares/src/undo-manager.ts#L180, the type of the parameter 'fn' here should be optional.


abruzzihraig:
It is closed due to code conflicts.

Issue #843, opened 2018-06-01
Change names: amountOfChildren


AjaxSolutions:
I don't want to be a smart guy (MST is a great library), but I just wanted to point out a minor grammatical error.

The word "amount" should be used with nouns that cannot be counted, but children, trees or people can be counted.

I suggest changing the word amount to number or count in the following:

"There is no limit to the amount of MST trees ..."
"amountOfChildren"
"amountOfPeopleOlderThan"

mattruby:
@AjaxSolutions please cut a PR with your suggested changes.

PR #848, opened 2018-06-05
Fix grammatical error, rename amount to number

AjaxSolutions:
Rename 'amount' to 'number' to fix a grammatical error.

https://github.com/mobxjs/mobx-state-tree/issues/843

https://www.ecenglish.com/learnenglish/lessons/whats-difference-between-amount-and-number


mattruby:
Thank you for your thoughtful recommendations!

If you notice any other grammatical errors, please don't hesitate to create another PR.

Thanks again!

Issue #623, opened 2018-01-25
decorate passTrough values


robinfehr:
### Edit

we should support both ways `decorate` and `addMiddleware(node...)` the pass trough option described later might therefore be way to improve that.

---  

.as I think the demand is increasing 
e.g.  #621 ~, @tim-field  on gitter, undoManger

for that, we can get the attached middlewares of an action within a middleware or within flows (e.g. if an action passed as a parameter). Also, it would solve the problem that lifecycle hook actions are passed to middlewares.

That would mean that we kind of should be able to attach flows or any other custom middleware declaratively and therefore "tag" the actions before the invocation. this though would mean that we'd have to attach the middleware to an action itself / multiple actions and find a syntax for that so that does not get too verbose.

example (bad example just to illustrate the idea):
```js
actions(‘groupName’, self => {
        setName({ name }) {
            self.name = name
        },
        setAge(age) {
            self.age = age
        }
}).middlewares(self => ({
        // of course addMiddleware would be hidden
        addMiddleware(self.groupName, undoManger.withoutUndo)
})
```

a downside/ upside (not sure yet) of this could be that we'd also have to declaratively add the opponent e.g. using 
withoutUndo for one group
withUndo for the other group
unless we'd keep the node middleware too as a default

mattiamanzati:
Uhm, is'nt that addMiddleware the same as decorate? :) https://github.com/mobxjs/mobx-state-tree/blob/master/API.md#decorate

robinfehr:
should be the same yes - a bit too verbose though if we'd like to setup all the actions like that

robinfehr:
first reaction of @mweststrate 
decorate(middleware, actionsmap). Then one could do 

```js
actions(self => ({
    ...decorate(middleware1, { some actions }),
    ...decorate(middleware2, { some other actions })
}))
```

robinfehr:
hm - next to allowing multiple actions - not thought through - but why not making the decoration also more general and allowing adding some props to the call instead of just middlewares - kind of tags that could be reused within middlewares.

robinfehr:
might also be solvable using a factory with a closure instead of a param. 
POC needed.

zuhorer:
issue moved to : https://github.com/coolsoftwaretyler/mst-middlewares/issues/13

coolsoftwaretyler:
[...]

PR #852, opened 2018-06-07
feat(undo-manager): implement methods debounceGroup and throttleGroup

abruzzihraig:
It's the feature that I want to import to UndoManager.

The method `startGroup` and `stopGroup` are helpful when we are trying to store state batches. They are significant in some cases like dragging on sliders but only track the value when mouse up.

Same to track batched states manually, `debounceGroup` and `throttleGroup` will give us a chance to automatically group states based on time. It is helpful in some cases especially when we want to track editing changes in interactive components(input, text area, editor, etc.)

### DIFF
--- a/packages/mst-middlewares/README.md
+++ b/packages/mst-middlewares/README.md
@@ -204,6 +204,8 @@ API:
 * `withoutUndoFlow(fn*)` patches the fn* are not recorded.
 * `startGroup(() => fn)` can be used to start a group, all patches within a group are saved as one history entry.
 * `stopGroup()` can be used to stop the recording of patches for the grouped history entry.
+* `debounceGroup((val?: T) => fn:(val?: T) => any, wait)` can be used to stop the recording of patches by a debounce function.
+* `throttleGroup((val?: T) => fn:(val?: T) => any, wait)` can be used to stop the recording of patches by a throttle function.
 
 Setup and API usage examples:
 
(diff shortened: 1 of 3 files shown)


robinfehr:
could you elaborate a bit on the usage? i'm a bit confused why this became a feature a in the undo manager and not more generally a middleware for actions. 

intuitively i'd have expected that to be e.g. in a separate middleware that could be put in front of the undo manager middleware.

abruzzihraig:
@robinfehr Yes, that sounds better. I haven't thought about the middleware at the beginning.

Let me explain to you why did I import these functions to undo manager. I made a website builder which uses MST, and I manually implemented the feature time traveling based on Snapshot. Recently, I realized that using JSON patch instead of recording snapshot would be better, so I tried to refactor it with the official middleware undo manager.

The most significant part of undo manager is that we can group patches. It resolves the issue that recording states based on continuous interactive events are usually useless.

So I used the method `startGroup` and `stopGroup` everywhere to reduce the useless recorded states, especially for the components which need continuous interaction like Slider. We can just start a group when mouse down, then stop the group when mouse up.

Everything is fine until I realized that I could not just use such simple way to define the 'group' when recording text content change. I cannot just use focus or blur to group it as it's so rough. Then I thought that I might need to group states based on time, like throttle or debounce.

I tried to wrap the method `startGroup` and `stopGroup` out of MST, and I asked myself, why don't just put these methods into the undo manager? There must be some people need this feature if they actually use the undo manager and if they need to track changes in text content. You can easily find the usage anywhere if a website needs users to input something and give them a possibility to undo. That's why I put these changes to the undo manager.

I'm not sure if it makes sense for the undo manager, maybe put them somewhere else would be better? But it is indeed helpful when using the undo manager in lots of real-world cases.

robinfehr:
i like the feature a lot but i think we should at least try to make it more general and put it into an own middleware that then can be chained with the undoManager. like that we could use it even for use cases where the undo manager is not involved. 

to do that properly i think we should have a look at the `decorate` and make use of it / extend it. it would be cool if we could declaratively add the middlewares to actions too (but not only declaratively I guess) and then chain multiple middlewares for that action. 

having it declaratively would for example allow your timing middleware to have the timer stored outside the middleware e.g. in the scope of the `.actions` where another action could abort the timer.

not sure we can achieve that nicely - the chaining (syntax wise for the user)...

to get into the topic a bit deeper, have a look at https://github.com/mobxjs/mobx-state-tree/issues/623
i think we should keep that issue in mind while trying to tangle the current problem where we'd have multiple middlewares to chain instead of multiple actions to one milddleware.

I hope it makes sense. else just hit me up on spectrum and we can talk forth and back about it abit.

mweststrate:
Closing PR for inactivity

Issue #439, opened 2017-10-09
applyAction with async result


mikecann:
Given an async action (borrowed from https://github.com/mobxjs/mobx-state-tree/blob/master/docs/async-actions.md#asynchronous-actions-and-middleware)

```typescript
const Store = types.model({
        githubProjects: types.array(types.frozen),
        state: types.enumeration("State", ["pending", "done", "error"])
    })
    .actions(self => ({
        fetchProjects: process(function* fetchProjects() { // <- note the star, this a generator function!
            self.githubProjects = []
            self.state = "pending"
            try {
                // ... yield can be used in async/await style
                self.githubProjects = yield fetchGithubProjectsSomehow()
                self.state = "done"
            } catch (e) {
                // ... including try/catch error handling
                console.error("Failed to fetch projects", error)
                self.state = "error"
            }
        })
    }))

const store = Store.create({})
// async actions will always return a promise resolving to the returned value
store.fetchProjects().then(() => {
    console.log("done")
})
```
How would one apply this action using `applyAction()` and await the result? The signature for apply action is:

```typescript
export declare function applyAction(target: IStateTreeNode, actions: ISerializedActionCall | ISerializedActionCall[]): void;
```

It would be nice if it could be made async:

```typescript
export declare function applyAction<T>(target: IStateTreeNode, actions: ISerializedActionCall | ISerializedActionCall[]): Promise<T>;
```

Then we can call:

```typescript
applyAction(store, {name: "fetchProjects"}).then(() => {
    console.log("done")
})
```

Obviously this is a contrived example, but if you have a deeply nested tree it becomes useful.

mweststrate:
I think it is a pretty neat idea and is feasible.

robinfehr:
closing in favour of #719

zuhorer:
issue moved to : https://github.com/coolsoftwaretyler/mst-middlewares/issues/14

coolsoftwaretyler:
^ Whoops, that's my bad. I gave unclear instructions and this got wrapped up in a project to extract mst middlewares. We are going to keep this open here.

coolsoftwaretyler:
Hey folks, it looks like we had some effort in this regard, but closed a stale PR: https://github.com/mobxjs/mobx-state-tree/pull/857

I'd love to review a new PR against current MST if anyone's interested. But if we don't hear back in a while, I will close this out as well and let it resurface if folks need it.

PR #857, opened 2018-06-08
feat(applyAction): returns Promise from the result of applyAction

abruzzihraig:
Implemented #439 
There are two things may need us to be careful:
* `IActionRecorder.replay` will return Promise<any[]>(same as the result of applyAction) instead of void/undefined.
* Some docs or code that keep in other packages(e.g. mst-middlewares) haven't been updated yet, because they use mst from node modules.

### DIFF
--- a/packages/mobx-state-tree/src/middlewares/on-action.ts
+++ b/packages/mobx-state-tree/src/middlewares/on-action.ts
@@ -31,7 +31,7 @@ export type ISerializedActionCall = {
 export interface IActionRecorder {
     actions: ReadonlyArray<ISerializedActionCall>
     stop(): any
-    replay(target: IStateTreeNode): any
+    replay(target: IStateTreeNode): Promise<any[]>
 }
 
 function serializeArgument(node: INode, actionName: string, index: number, arg: any): any {
@@ -144,7 +144,7 @@ function baseApplyAction(target: IStateTreeNode, action: ISerializedActionCall):
  *      // stop recording actions
  *      stop(): any
  *      // apply all the recorded actions on the given object
- *      replay(target: IStateTreeNode): any
+ *      replay(target: IStateTreeNode): Promise<any[]>
  * }
  *
  * @export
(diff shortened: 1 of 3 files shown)


packages/mobx-state-tree/src/middlewares/on-action.ts
```diff
@@ -81,16 +81,34 @@ function serializeTheUnserializable(baseType: string) {
...
         if (!isStateTreeNode(target))
             fail("expected first argument to be a mobx-state-tree node, got " + target + " instead")
         if (typeof actions !== "object")
             fail("expected second argument to be an object or array, got " + actions + " instead")
     }
-    runInAction(() => {
-        asArray(actions).forEach(action => baseApplyAction(target, action))
+
+    return new Promise((res, rej) => {
+        runInAction(() => {
```
mweststrate: I am wondering whether async could be opt-in here; e.g. run synchronosly until there is an action that returns a promise, and then start chaining them together?

abruzzihraig: @mweststrate Did you mean:
1. if each action that passed in `applyAction` is synchronous, each action should be executed synchronously and finally return one promise.
2. if any action that passed in `applyAction` is asynchronous, just wait for those async actions fulfilled, then chain their results and the results from sync actions that were synchronously executed. After that, returns one promise.

The point is that execute synchronous actions at the first tick.

Is this what you want?

jamonholmgren:
More discussion needed, but seems like a cool idea. Perhaps we could expose it as `applyActionAsync` to avoid a breaking change.

jesse-savary:
If `applyActionAsync` is a viable path forward I'd be willing to clean this PR up @jamonholmgren

ryskin:
This is a really good Idea I need this!

ryskin:
When this will be merged???

jamonholmgren:
@jesse-savary Sorry, I didn't see this until now! I don't want to break any existing APIs right now, but I'd be more than happy to merge a PR with `applyActionAsync` if you want to tackle it.

coolsoftwaretyler:
Hey folks - this PR has gone pretty stale, and we are going to be making some big diffs to the repo pretty soon.

I'm going to close this out. It's certainly a feasible and good option, and if anyone who was involved would like to implement the secondary `applyActionAsync`, I'd be happy to review a new PR to that effect. But in the absence of someone who is up for it, I'm going to close this out to clean things up.

Thanks all!

PR #865, opened 2018-06-12
Fix todomvc example

O4epegb:
Mobx was throwing some strange errors there, like `mobx-react requires mobx to be available`.
Updated to last `mobx-react` version and last `mobx@4` version


mweststrate:
Nice catch, thanks!

PR #868, opened 2018-06-13
WIP: Mobx 5 support

jeffberry:


mweststrate:
I see why this fail, and check if I can fix it :)

PR #875, opened 2018-06-15
Fix undo-manager not working on child trees

marcofugaro:
I've been using the undo-manager middleware, I found out it throws an error when the `targetStore` is not the root store. 

This happens because the patch is saved with the path from the root store, and applied to the targetStore instead, if those are not the same, it throws an error.

This fixes it.


mweststrate:
Nice catch, thanks!

PR #883, opened 2018-06-18
Do not preProcess snapshot if it is already a node (fixes #876)

### LINKED ISSUE #876: map.put throws error, works fine with map.set(id, value)
I'm getting the following error when using `put` to add to a map. 
`Cannot add an object to a state tree if it is already part of the same or another state tree. Tried to assign an object to '<somepath>', but it lives already at '<anotherpath>'`

This seems to be the same issue described in #383.
I was able to reproduce in CodeSandbox and it seems to only happen if I use `preProcessSnapshot`. The problem doesn't occur if i use `map.set(id, value)`

Here's a link to the sandbox showing the error.

https://codesandbox.io/s/ox6v8jrl9y

Reference line 36 of alertStore.js

k-g-a:
Fixes [#876](https://github.com/mobxjs/mobx-state-tree/issues/876).
Current test are passing, but I had no time to write explicit test for this particular issue.


k-g-a:
@mweststrate, i'll try to add a test within two days :)

mweststrate:
Merged, will release soon :)

Issue #873, opened 2018-06-15
Cannot compose more than 3 models. Expected 2-4 arguments, but got 5.


pravinbashyal:
I want to try following: 

```javascript
export const Shipment = types
  .compose('ModelName', model1, model2, model3, mode4) // Expected 2-4 arguments, but got 5.
```
but it gives `expected 2-4 arguments warning`.

The [source](https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/types/complex-types/model.ts#L538): 
```javascript
export function compose(...args: any[]): IModelType<any, any> {
    // TODO: just join the base type names if no name is provided
    const typeName: string = typeof args[0] === "string" ? args.shift() : "AnonymousModel"
    // check all parameters
    if (process.env.NODE_ENV !== "production") {
        args.forEach(type => {
            if (!isType(type)) fail("expected a mobx-state-tree type, got " + type + " instead")
        })
    }
    return (args as ModelType<any, any>[])
        .reduce((prev, cur) =>
            prev.cloneAndEnhance({
                name: prev.name + "_" + cur.name,
                properties: cur.properties,
                initializers: cur.initializers
            })
        )
        .named(typeName)
}
```

and [documentation](https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/types/complex-types/model.ts#L530): 

 - Composes a new model from one or more existing model types.
 - This method can be invoked in two forms:
 - *Given 2 or more model types, the types are composed into a new Type*

does not seem to limit the number of models to compose from. but somehow its limited in [types](https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/types/complex-types/model.ts#L519): 
```javascript
export declare function compose<T1, S1, T2, S2, T3, S3>(t1: IModelType<T1, S1>, t2: IModelType<T2, S2>, t3?: IModelType<T3, S3>): IModelType<T1 & T2 & T3, S1 & S2 & S3>;
export declare function compose<T1, S1, A1, T2, S2, A2, T3, S3, A3>(name: string, t1: IModelType<T1, S1>, t2: IModelType<T2, S2>, t3?: IModelType<T3, S3>): IModelType<T1 & T2 & T3, S1 & S2 & S3>;
```
I solved it using following: 
```javascript
types.compose(model1, types.compose( model2, types.compose(model3, ...)))
```

[...]

PR #892, opened 2018-06-21
Update compose function overload type definition

t49tran:
This PR purpose is to update `compose` overload type definition to cover scenario when more than 4 parameters are passed to the function as raised in the issue:  #873.

### DIFF
--- a/packages/mobx-state-tree/src/types/complex-types/model.ts
+++ b/packages/mobx-state-tree/src/types/complex-types/model.ts
@@ -532,10 +532,23 @@ export function compose<T1, S1, A1, T2, S2, A2, T3, S3, A3>(
     t2: IModelType<T2, S2>,
     t3?: IModelType<T3, S3>
 ): IModelType<T1 & T2 & T3, S1 & S2 & S3> // ...and so forth...
+export function compose(
+    t1: IModelType<any, any>,
+    t2: IModelType<any, any>,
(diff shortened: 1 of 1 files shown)


packages/mobx-state-tree/src/types/complex-types/model.ts
```diff
@@ -532,10 +532,23 @@ export function compose<T1, S1, A1, T2, S2, A2, T3, S3, A3>(
     t2: IModelType<T2, S2>,
     t3?: IModelType<T3, S3>
 ): IModelType<T1 & T2 & T3, S1 & S2 & S3> // ...and so forth...
+export function compose(
+    t1: IModelType<any, any>,
+    t2: IModelType<any, any>,
+    ...types: Array<IModelType<any, any>>
```
mweststrate: I don't think it is useful to add a variadic argument here, as it looses all typings. Can you instead follow the pattern of the first overload, but then we with more generic arguments. (See for example how that is done for unions or enumerations)

t49tran: @mweststrate, thanks for the review I will have a look

t49tran:
Hi @mweststrate , I have updated the overload functions following example in `enumeration`

mweststrate:
Merged, thanks!

Issue #889, opened 2018-06-19
[Question] UndoManager - Extend the Entry model and callbacks in undo/redo actions


chemitaxis:
Hi as we have been talking in Twitter, I need to extend the Entry Model to add a new prop.

In my case, I have something like this:

AddUndoState:

```
 addUndoState(recorder) {
        if (recorder.patches.length === 0) {
          return;
        }

        if (replaying) {
          // skip recording if this state was caused by undo / redo
          return;
        }
        self.history.splice(self.undoIdx);
        self.history.push({
          zoom: $('#canva').css('transform'),
          patches: recorder.patches,
          inversePatches: recorder.inversePatches
        });

        self.undoIdx = self.history.length;
      }
```

I need to save the css transform in the history, because if the action was in a different "environment" (zoomed, for example), if I revert the change, I have an issue with my selector. We are not recording nothing about the zoom.

I need to extend the entry model to something like this:

```
const Entry = mst.types.model('UndoManagerEntry', {
  zoom: mst.types.frozen,
  patches: mst.types.frozen,
  inversePatches: mst.types.frozen
});

```

In the other hand, to solve my problem, I can pass a callback to the undo/redo action, and check if the "environment" is different and process in the correct way:

```
undo(callback) {
        replaying = true;
        self.undoIdx -= 1;
        const { zoom } = self.history[self.undoIdx];
        // n.b: reverse patches back to forth
        // TODO: add error handling when patching fails? E.g. make the operation atomic?
        mst.applyPatch(
          targetStore,
          self.history[self.undoIdx].inversePatches.slice().reverse()
        );
        replaying = false;

        callback(zoom);
      }
```

My callback function:

```
function example_callback(zoom){
          if (
            !!rootStore.canvaStore.selector &&
            zoom !== $('#canva').css('transform')
          ) {
            // DO CUSTOM STUFF HERE
          }
}
```

I have this working, but I have copied the code of the undoManager and I have added to my project, but I don't want to do it... any ideas? Thanks @mweststrate & @robinfehr

robinfehr:
I'm still confused :) what prevents you from storing the zoom on a canvasModel/ viewModel?

mweststrate:
Closing for inactivity

PR #895, opened 2018-06-25
add(undoManager): add timestamp to the object recorder

abruzzihraig:
Just add a timestamp for recorder objects like what I suggested on https://spectrum.chat/?t=d5c82ac2-8259-45ce-a72c-a4503c591604

Could anyone can help me with it? No matter what discussion or code review on this requirement. Thanks!

### DIFF
--- a/packages/mst-middlewares/src/undo-manager.ts
+++ b/packages/mst-middlewares/src/undo-manager.ts
@@ -16,6 +16,10 @@ import {
 } from "mobx-state-tree"
 import { IObservableArray } from "mobx"
 
+interface IPatchRecord extends IPatchRecorder {
+    timestamp: Readonly<number>
+}
+
 const Entry = types.model("UndoManagerEntry", {
     patches: types.frozen,
     inversePatches: types.frozen,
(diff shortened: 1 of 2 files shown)


abruzzihraig:
I just found the issue #889 seems to have a similar request with me. If there could be any way to extend the recorder object will be better.

robinfehr:
the timestamp might really be something that is useful and makes sense for this particular middleware. 

I'm not entirely happy that one has to store the timestamp but i guess it wouldn't be that memory / storage (if persisted) consumptive either that it should be optional .
@mweststrate what do you think?


about the extending: I can't yet understand where it could be useful and fit the undoManager middleware topic - middleware for time travelling.

abruzzihraig:
@robinfehr Thanks for your reply.

For the extending, yes I don't know where it could be useful as well. I was just thinking if there would be something like timestamp and we are going to add them into the recorder object in the future, which might be better to make the recorder object extendable. I'm not sure about that.

I agree with you on the timestamp. It should be optional. The code I committed it just basically explains the how did I think about the timestamp. You can even see there is a temporary interface extends IPatchRecorder, which should be modified in some formal way. And that's why I need your help and recommendations.

robinfehr:
hm I guess that's ok since it's extended for this middleware but i'll go over the code in more detail and leave notes if necessary. thanks.

abruzzihraig:
@robinfehr when you check the code, please at lease notice these two points:
1. I'm not sure if the `Date.now()` I used follows the browser compatibility that MST requests
2. I remember that this code change might influence the method `addUndoState` because it can be invoked directly from outside by developers, but they might not know that they can pass a timestamp for the recorder object.

mweststrate:
Closing this PR for the lack of a champion / activity / recent need

PR #899, opened 2018-06-28
Mobx-state-tree 3

mweststrate:
See: https://github.com/mobxjs/mobx-state-tree/projects/2


dnakov:
so many great things here! thanks everyone!!!

Bnaya:
Is there's any nightly/tag published to npm with this WIP?

mweststrate:
Nope, but clone, build, link should be the king. The current status of the
branch should be very stable, it is just not complete yet :) (and the
changelog is not too elaborate yet)

Op vr 29 jun. 2018 om 11:55 schreef Bnaya Peretz <notifications@github.com>:

mattiamanzati:
@mweststrate  what about keeping maybe nullable and introduce `option` as name for `Type | undefined`?
Maybe in programming languages is usually `Type | null`.
Also this would be less breaking change

https://gcanti.github.io/fp-ts/Option.html

PR #901, opened 2018-06-28
Mst3 optimizations

mweststrate:
@k-g-a while testing the optimizations, I discovered that paths didn't always correctly propagate there changes (see the added unit test in node.ts). 

So I reverted that to being an observable, but created a lowlevel atom instead, which is a bit cheaper. Also found out that the observables used in scalar-node only slowed things down. (Scalar nodes are not reconciled, so their own subpath / direct parent will never change)

Numbers on the speedtest:

```
Master:      24.6 sec / 352 MB
MST base:    10.8 sec   / 179 MB
this branch: 10.6 sec / 182 MB
```

Would you mind reviewing? I think the numbers are very acceptable and this doesn't include the snapshot disposer optimization yet

### DIFF
--- a/packages/mobx-state-tree/.vscode/launch.json
+++ b/packages/mobx-state-tree/.vscode/launch.json
@@ -5,7 +5,7 @@
             "type": "node",
             "request": "launch",
             "name": "debug unit test",
-            "program": "${workspaceRoot}/node_modules/jest/bin/jest.js",
+            "program": "${workspaceRoot}/node_modules/.bin/jest",
             "args": ["-i", "${file}"],
         }
     ]
--- a/packages/mobx-state-tree/package.json
+++ b/packages/mobx-state-tree/package.json
@@ -10,7 +10,7 @@
     "build": "tsc && cpr lib dist --delete-first --filter=\\\\.js && yarn rollup",
     "rollup": "rollup -c",
     "test": "cross-env NODE_ENV=development jest --ci && cross-env NODE_ENV=production jest --ci && yarn run test-cyclic && yarn run test-mobx4",
-    "speedtest": "yarn build && node --expose-gc test/perf/report.js",
+    "speedtest": "yarn build && /usr/bin/time node --expose-gc test/perf/report.js",
     "test-cyclic": "yarn run build && node -e \"require('.')\"",
     "test-mobx4": "yarn add -D mobx@4.3.1 && yarn build && jest --ci && git checkout package.json ../../yarn.lock && yarn install",
     "watch": "jest --watch",
--- a/packages/mobx-state-tree/src/core/node/object-node.ts
+++ b/packages/mobx-state-tree/src/core/node/object-node.ts
@@ -49,6 +49,7 @@ export class ObjectNode implements INode {
 
     subpathAtom = createAtom(`path`)
     subpath: string = ""
+    escapedSubpath: string
     parent: ObjectNode | null = null
     state = NodeLifeCycle.INITIALIZING
     storedValue: any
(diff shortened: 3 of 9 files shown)


k-g-a:
I'll try to look today or tomorrow. 
By the way, a question came to my mind. Path computation seems quite cheap: couple of replaces plus one concatination - so do we really want it to be ```@computed``` at all? 
It seems like we can even make ```this.subpath = escapeJsonPath(subpath)``` at constructror/setParent and make it even cheaper?

mweststrate:
Good one, will add that


Op do 28 jun. 2018 om 13:21 schreef Gleb <notifications@github.com>:

k-g-a:
I peeked into code, it looks celver! Should we use same approach for ```snapshot``` instead of my tricky ```trackAndCompute()```?

mweststrate:
I was tempted to do that, but on the other hand, although I didn't entirely
understand why it works, current approach seems to work flawless :) But
yeah, maybe we should. I'll update it later

Op do 28 jun. 2018 om 14:15 schreef Gleb <notifications@github.com>:

mweststrate:
@k-g-a, tried to use atoms, but I noticed that it emits more snapshot. That might be correct, but it fails two tests so figured to skip it for now, until current approach causes actual problems. 

The diff was:

```diff
diff --git a/packages/mobx-state-tree/src/core/node/object-node.ts b/packages/mobx-state-tree/src/core/node/object-node.ts
index 774318b..ed1f688 100644
--- a/packages/mobx-state-tree/src/core/node/object-node.ts
+++ b/packages/mobx-state-tree/src/core/node/object-node.ts
@@ -25,7 +25,6 @@ import {
     resolveNodeByPathParts,
     convertChildNodesToArray,
     ModelType,
-    invalidateComputed,
     IAnyType
 } from "../../internal"
 
@@ -77,6 +76,8 @@ export class ObjectNode implements INode {
     private _snapshotSubscribers: ((snapshot: any) => void)[] | null = null
 
     private _observableInstanceCreated: boolean = false
+    private _observableInstanceCreatedAtom = createAtom("hasObservableInstance")
+
     private _childNodes: IChildNodesMap | null
     private _initialSnapshot: any
     private _createNewInstance: ((initialValue: any) => any) | null
@@ -145,9 +146,7 @@ export class ObjectNode implements INode {
                 this.state = NodeLifeCycle.DEAD
             }
         }
-        // NOTE: we need to touch snapshot, because non-observable
-        // "observableInstanceCreated" field was touched
-        invalidateComputed(this, "snapshot")
+        this._observableInstanceCreatedAtom.reportChanged()
 
         const snapshotDisposer = reaction(
             () => this.snapshot,
@@ -248,7 +247,7 @@ export class ObjectNode implements INode {
         if (!this.isAlive) return undefined
         const snapshot = this._observableInstanceCreated
             ? this._getActualSnapshot()
-            : this._getInitialSnapshot()
+            : (this._observableInstanceCreatedAtom.reportObserved(), this._getInitialSnapshot())
         return freeze(snapshot)
     }
 
diff --git a/packages/mobx-state-tree/src/utils.ts b/packages/mobx-state-tree/src/utils.ts
index 41d4f11..5ae0909 100644
--- a/packages/mobx-state-tree/src/utils.ts
+++ b/packages/mobx-state-tree/src/utils.ts
@@ -176,11 +176,6 @@ export function argsToArray(args: IArguments): any[] {
     return res
 }
 
-export function invalidateComputed(target: any, propName: string) {
-    const atom = getAtom(target, propName) as any
-    atom.trackAndCompute()
-}
-
 export type DeprecatedFunction = Function & { ids?: { [id: string]: true } }
 let deprecated: DeprecatedFunction = function() {}
 deprecated = function(id: string, message: string): void {
```

k-g-a:
I think this reportObserved() at line 251:
```(this._observableInstanceCreatedAtom.reportObserved(), this._getInitialSnapshot())```
caused unnecessary new snapshots. It seems excessive at this point.

mweststrate:
Possibly, yet reportObserved() needs to be called somewhere, to have MobX
establish the dependency of snapshot on that atom

Op vr 29 jun. 2018 om 15:01 schreef Gleb <notifications@github.com>:

PR #902, opened 2018-06-28
PR 845 improvements

k-g-a:
1) review fixes
2) changes from [this comment](https://github.com/mobxjs/mobx-state-tree/pull/845#issuecomment-400721626) about reactions


mweststrate:
P.S. @k-g-a not that you should be able to push directly to branches on the mst repository, that might simplify things as you don't have to keep your fork in sync :)

Issue #804, opened 2018-05-09
Idea: introduce eager union


mweststrate:
See also: https://spectrum.chat/?t=2abe14ea-d0bc-4300-9f03-8953078b4c77

mattruby:
I'd be happy to help with this one in any way.

mweststrate:
Interested in PR with unit tests? (implementation is welcome as well of
course, but writing tests might be less intimidating and is at least as
helpful 😊)

Op wo 9  2018 16:19 schreef Matt Ruby <notifications@github.com>:

mweststrate:
Eager should probably become the default, which has nice performance benefits

mattruby:
I've been working on this: https://github.com/mobxjs/mobx-state-tree/compare/master...mattruby:feature/eager-union

Could someone help me out with how to take a varaible number of parameters first and an options object on the tail in TS?

The biggest change was here: https://github.com/mobxjs/mobx-state-tree/compare/master...mattruby:feature/eager-union#diff-4b29a6b67d0347e1e37b69c96cd65b7eR330

In doing this, I now properly sort the union types.  Before, the first types was appended to the end.

mweststrate:
@mattruby awesome! I think the union order swap is correct indeed. @k-g-a noticed the same. 

I think there is a utility `isType`, so if you check `isObject(arguments[arguments.length -1]) && !isType(arguments[arguments.length -1]))` you can `.pop()` it of to get the options

mweststrate:
Hey @mattruby! What is the status on this one? Need help somewhere? Feel free to open a PR already and mark it as `WIP:`

mattruby:
Sorry for the delay. Let me see where I can get this morning.

mweststrate:
Reviewed the branch, and merged it with some little changes! Saw the the trouble with that last argument indeed, so instead changed the setup to use `{ dispatcher: fn, eager: boolean }`. 

Thanks a lot @mattruby!

mattruby:
THANKS!!
O man!  Moving the dispatcher function is a great decision.

mweststrate:
Thank you! It was just the last push ;-)

mweststrate:
Closing, will be part of MST 3

PR #907, opened 2018-07-03
WIP - Feature/eager union

mattruby:
This change enables eager unions by default.  One of the most impactful changes is the proper sorting on union types.  Currently Union will take the first param and check if it's a dispatch function.  If it's a Type it will be appended to the end of the types array.  I've moved it back to the front. 

I'm having a hard time pop-ing the options off the array of types.  In addition, I don't have a clue if I'm approaching the UnionOptions TS the correct way.  I tried to follow reference.ts as an example.

### DIFF
--- a/packages/mobx-state-tree/src/types/utility-types/union.ts
+++ b/packages/mobx-state-tree/src/types/utility-types/union.ts
@@ -14,9 +14,12 @@ import {
 
 export type ITypeDispatcher = (snapshot: any) => IType<any, any>
 
+export type UnionOptions<T> = {
+    eager: Boolean
+}
 export class Union extends Type<any, any> {
     readonly dispatcher: ITypeDispatcher | null = null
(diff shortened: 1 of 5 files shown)


mweststrate:
Implements #804

packages/mobx-state-tree/src/types/utility-types/union.ts
```diff
@@ -62,7 +72,11 @@ export class Union extends Type<any, any> {
 
         // find the most accomodating type
         const applicableTypes = this.types.filter(type => type.is(value))
-        if (applicableTypes.length > 1)
+        if (applicableTypes.length > 1) {
+            debugger
```
praxxis: this should probably be removed before release :)

mattruby: Yikes!  Thanks!

PR #931, opened 2018-07-19
add Typeflags to exports

benjaminjackman:
Some of the code being generated for .d.ts seems to be using (for example)
```typescript
flags: TypeFlags.Optional;
```
but it's not being exported so that's causing some errors

### DIFF
--- a/packages/mobx-state-tree/src/index.ts
+++ b/packages/mobx-state-tree/src/index.ts
@@ -93,7 +93,8 @@ export {
     recordActions,
     createActionTrackingMiddleware,
     setLivelynessChecking,
-    LivelynessMode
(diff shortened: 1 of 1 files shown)


benjaminjackman:
Looks like maybe i need to add: 
TypeFlags
to: 
https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/test/api.ts

xaviergonz:
I'm also exporting TypeFlags on the PR here https://github.com/mobxjs/mobx-state-tree/pull/937

benjaminjackman:
ok i'll close mine then

benjaminjackman:
dupe #937

PR #934, opened 2018-07-19
Fix markdown syntax in Api overview section

hnq90:
Before: [`setLivelynessChecking("warn" | "ignore" | "error")`]
After: [`setLivelynessChecking("warn" \| "ignore" \| "error")`]


mweststrate:
Thanks!

Issue #932, opened 2018-07-19
(Typescript) frozen fields without a default always include `undefined` as a possible type


fruitraccoon:
I have:

* [x] I think something is not working as it should.
  * [x] I've checked documentation and searched for existing issues
  * [x] I've made sure your project is based on the latest MST version
  * [x] Fork [this](https://codesandbox.io/s/jlnmjqr7xv) code sandbox or another minimal reproduction. 
  * [x] Describe expected behavior
  * [x] Describe observed behavior

**Expected Behaviour:**
For a model like this:

    const MyModel = types.model('MyModel', {
      myField: types.array(types.frozen<MyInterface>())
    });

I would expect the type of `myField` to effectively be `Array<MyFieldType>`.

**Observed Behaviour:**
The type of `myField` is effectively `Array<MyFieldType | undefined>`.  

I expected to have to use something like `myField: types.array(types.frozen<MyInterface | undefined>())` or wrap in a maybe/optional to get this.

**Detail**
I'm looking to update from version 2.x to version 3 to take advantage of improvements, and have found this a little inconsistent:
* For MST model fields defined using other MST models, the field must be wrapped in a maybe/optional in order for `undefined` to be a valid value of the field.
* For the new typed `frozen` fields, the defaulting functionality appears to be already included (as an optional constructor) rather than using maybe/optional.

This means that when wanting a field defined as an array of frozen objects, one of the following trade-offs must be used:
* declare it like `field: types.array(types.frozen<MyItemType>()),` and then always have to check if individual items are `undefined` (and pushing `undefined` is allowed)
* declare it like `field: types.array(types.frozen<MyItemType>(myDefaultItemInstance)),` which has the correct typing, but the default instance is never used and can be difficult to come up with for complicated types (think large hierarchical business entities read from a server).

**Questions:**
* Are there any other approaches that I haven't thought of to get the typing I'm after?
* If not, is there an additional overload that could be added (or an additional type) to use to be able to explicitly define the frozen type used?  (Looking for guidance on the preferred approach in this case - I could have a go at a PR if it's not too out-there!)

Example codesandbox available here: https://codesandbox.io/s/5xj57x0ym4 - please see `src/models/Frozen.ts` and the `compareFrozenFields` action.

Thanks very much!

**Edit:**
I've also found that when a frozen field should be optionally `undefined`, I still have to use `types.maybe` otherwise when I try to use that type, typescript insists that the field must be present.

[...]

Issue #930, opened 2018-07-18
node_modules/mst-middlewares/dist/undo-manager.d.ts:17:16 - error TS2503: Cannot find namespace 'TypeFlags'.


xaviergonz:
I have:

* [x] I think something is not working as it should.
  * [x] I've checked documentation and searched for existing issues
  * [x] I've made sure your project is based on the latest MST version
  * [ ] Fork [this](https://codesandbox.io/s/jlnmjqr7xv) code sandbox or another minimal reproduction. 
  * [x] Describe expected behavior
  * [x] Describe observed behavior

```
node_modules/mst-middlewares/dist/undo-manager.d.ts:17:16 - error TS2503: Cannot find namespace 'TypeFlags'.
```

latest mobx, mobx-state-tree and mst-middlewares
ts 2.9.2
all strict options set to true

fruitraccoon:
To work around this issue, I downgraded `mst-middlewares` from version `3.0.0` to `2.2.0`.

mweststrate:
Released as 3.0.2

Issue #922, opened 2018-07-16
Error with optional and compose models. Cannot read property 'find' of undefined


jchamb:
I have:
* [x] I think something is not working as it should.
  * [x] I've checked documentation and searched for existing issues
  * [x] I've made sure your project is based on the latest MST version

Updating my code to v3.0 ran into an issue after removing types.optional around my array types. I get the above error anytime I use a composed model. Reproduction of the issue below  

https://codesandbox.io/s/71kz23n6pj

mweststrate:
Confirmed, that looks like a bug indeed

jchamb:
error is also thrown with optional around enumeration.
`types.optional(types.enumeration("state", ["init", "pending", "done", "error"]), "init")`

mweststrate:
Confirmed, fixed by #937. Fix will be released later today

mweststrate:
Released as 3.0.2

PR #937, opened 2018-07-20
fixes for TS types and added types to unit tests

xaviergonz:
basically a (kinda big) pull request to fix lots of TS typing issues

btw, shouldn't custom type actually use <C, S, T> rather than <S|T,S,T> for more flexibility?

### DIFF
--- a/packages/mobx-state-tree/src/core/mst-operations.ts
+++ b/packages/mobx-state-tree/src/core/mst-operations.ts
@@ -367,18 +367,16 @@ export function hasParentOfType(target: IStateTreeNode, type: IAnyType): boolean
     return false
 }
 
-export function getParentOfType(target: IStateTreeNode, type: IAnyType): any & IStateTreeNode
-export function getParentOfType<S, T>(target: IStateTreeNode, type: IAnyType): T & IStateTreeNode
 /**
  * Returns the target's parent of a given type, or throws.
  *
  *
  * @export
- * @param {Object} target
- * @param {IAnyType} type
- * @returns {*}
+ * @param {IStateTreeNode} target
+ * @param {IType<any, any, T>} type
+ * @returns {T}
  */
-export function getParentOfType<S, T>(target: IStateTreeNode, type: IAnyType): T {
+export function getParentOfType<T>(target: IStateTreeNode, type: IType<any, any, T>): T {
     // check all arguments
     if (process.env.NODE_ENV !== "production") {
         if (!isStateTreeNode(target))
@@ -395,8 +393,8 @@ export function getParentOfType<S, T>(target: IStateTreeNode, type: IAnyType): T
     return fail(`Failed to find the parent of ${getStateTreeNode(target)} of a given type`)
 }
 
-export function getRoot(target: IStateTreeNode): any & IStateTreeNode
 export function getRoot<T>(target: IStateTreeNode): T & IStateTreeNode
+export function getRoot(target: IStateTreeNode): any & IStateTreeNode
 /**
  * Given an object in a model tree, returns the root object of that tree
  *
--- a/packages/mobx-state-tree/src/types/complex-types/array.ts
+++ b/packages/mobx-state-tree/src/types/complex-types/array.ts
@@ -256,12 +256,15 @@ export class ArrayType<C, S, T> extends ComplexType<C[], S[], IObservableArray<T
  */
 export function array<C, S, T>(
     subtype: IType<C, S, T>
-): IComplexType<ReadonlyArray<C>, ReadonlyArray<S>, IObservableArray<T>> {
+): IComplexType<ReadonlyArray<C> | undefined, ReadonlyArray<S>, IObservableArray<T>> & {
+    flags: TypeFlags.Optional
+} {
     if (process.env.NODE_ENV !== "production") {
         if (!isType(subtype))
             fail("expected a mobx-state-tree type as first argument, got " + subtype + " instead")
     }
-    return new ArrayType<C, S, T>(subtype.name + "[]", subtype)
+    const ret = new ArrayType<C, S, T>(subtype.name + "[]", subtype)
+    return ret as typeof ret & { flags: TypeFlags.Optional }
 }
 
 function reconcileArrayChildren<T>(
--- a/packages/mobx-state-tree/src/types/complex-types/map.ts
+++ b/packages/mobx-state-tree/src/types/complex-types/map.ts
@@ -403,8 +403,10 @@ export class MapType<C, S, T> extends ComplexType<
 }
 
 export function map<C, S, T>(
-    subtype: IType<C, S, T>
-): IComplexType<IKeyValueMap<C>, IKeyValueMap<S>, IMSTMap<C, S, T>>
+    subtype: IComplexType<C, S, T>
+): IComplexType<IKeyValueMap<C> | undefined, IKeyValueMap<S>, IMSTMap<C, S, T>> & {
+    flags: TypeFlags.Optional
+}
 /**
  * Creates a key based collection type who's children are all of a uniform declared type.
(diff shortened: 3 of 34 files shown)


xaviergonz:
There's one unit test failing:
- it should emit snapshots for children
but it also fails in master
```
 it should emit snapshots for children

    expect(received).toEqual(expected)

    Expected value to equal:
      {"files": [{"name": "01-arrival"}, {"name": "Photo2"}], "name": "Vacation photos"}
    Received:
      {"name": "01-arrival"}

    Difference:

    - Expected
    + Received

      Object {
    -   "files": Array [
    -     Object {
        "name": "01-arrival",
    -     },
    -     Object {
    -       "name": "Photo2",
    -     },
    -   ],
    -   "name": "Vacation photos",
      }

      139 |     onSnapshot(folder.files[0], snapshot => snapshots.push(snapshot))
      140 |     folder.files[0].rename("01-arrival")
    > 141 |     expect(snapshots[1]).toEqual({
          |                          ^
      142 |         name: "Vacation photos",
      143 |         files: [{ name: "01-arrival" }, { name: "Photo2" }]
      144 |     })
```

xaviergonz:
Apparently the error was an accidental change to mobx-state-tree package-json that changed mobx version to 4.3.1 from 5.0.3.
I restored it and now everything seems to be fine.

packages/mobx-state-tree/package.json
```diff
@@ -41,7 +41,7 @@
     "cross-env": "^5.1.1",
     "documentation": "^5.2.2",
     "jest": "^23.2.0",
-    "mobx": "5.0.3",
+    "mobx": "4.3.1",
```
benjaminjackman: is that intentional?

xaviergonz: accidental, already reverted

xaviergonz:
besides making TypeFlags public and exported also added TypeFlags imports to undo-manager and the other mst-middleware to fix https://github.com/mobxjs/mobx-state-tree/issues/930

xaviergonz:
Also fixes https://github.com/mobxjs/mobx-state-tree/issues/932

xaviergonz:
just added a fix (with unit test) for optional enums inside a model crashing
in other words, the following will crash:
```
test("should support optional enums inside a model", () => {
    const TrafficLight = types.model({
        color: types.optional(types.enumeration(["Red", "Orange", "Green"]), "Orange")
    })
    const l = TrafficLight.create({})
    expect(l.color).toBe("Orange")
})
```

because types.optional would inherit as flag TypeFlags.Union, but not "types" and therefore crash here
```
function tryGetOptional(type: any): OptionalValue<any, any, any> | undefined {
    if (!type) return undefined
    if (type.flags & TypeFlags.Union) return type.types.find(tryGetOptional) // THIS LINE
    if (type.flags & TypeFlags.Late) return tryGetOptional(type.subType)
    if (type.flags & TypeFlags.Optional) return type
    return undefined
}
```

so I fixed it like this
```
function tryGetOptional(type: any): OptionalValue<any, any, any> | undefined {
    if (!type) return undefined
    // we need to check for type.types since an optional union doesn't have direct subtypes
    if (type.flags & TypeFlags.Union && type.types) return type.types.find(tryGetOptional)
    if (type.flags & TypeFlags.Late && type.subType) return tryGetOptional(type.subType)
    if (type.flags & TypeFlags.Optional) return type
    return undefined
}
```

another option would be to make optional just return the optional flag, but I'm not sure if that's better

mweststrate:
Thanks for the investigation! I think the solution is correct, will double
check next week!

Op vr 20 jul. 2018 om 20:01 schreef Javier Gonzalez <
notifications@github.com>:

xaviergonz:
Maybe fixes #922

xaviergonz:
Ok, basically I took the opportunity to make a revamp of the typescript typing to further improve it.
Notable changes is that ISnapshottable is no longer there and its role has been taken by IStateTreeNode<S>

With these changes now getSnapshot is properly typed without the need to use its generic argument, references and identifiers work properly and lots of other little fixes.

xaviergonz:
I made this available in npm as ```@xaviergonz/mobx-state-tree @xaviergonz/mst-middlewares``` for anybody that might want to give it a try btw

xaviergonz:
I had to remove DeepImmutable. While cool (made snapshots from getSnapshot unmodifiable, etc) I have a large project that uses MST and it was getting out of ram errors while compiling because of it.

That being said, I'm quite happy with the current state of the typings 👍 so I don't think I'll make further changes.

xaviergonz:
@mweststrate my pleasure :)

mweststrate:
Took a close look and these are great improvements! Really awesome work. I've just send you a maintainers invite, that will simplify the workflow for any future contributions a lot, as you can directly push, branch etc on this repo :-)!

dpnolte:
@xaviergonz  thanks this fixes my issue 👍 
I hope this get published soon on npm.

marvwhere:
cool thx - tried it out (v3.0.2) 

looks like it's causing some problems with fork-ts-checker-webpack-plugin - "Starting type checking and linting service..." will never finish and creating a node instance with >100% cpu

went back to 2.2.0 (working)
updated to 3.0.0 (working)
updated to 3.0.1 (not working)

Typescript 2.9
tslint 5.11.0

mweststrate:
and if you disable tslint?


Op do 26 jul. 2018 om 16:04 schreef Marvin <notifications@github.com>:

xaviergonz:
I have seen that happen from time to time when there are lots of compilation errors. Basically the default node memory of node processes is around 512mb and complex type checking takes lots of ram and then crash.

Things I'd try include:
- run tsc by itself on the command line: ```npx tsc -p .```
- give it more ram if it dies because of out of ram: ```NODE_OPTIONS=--max-old-space-size=4096 && tsc -p .```
- make use of the speed up trick (interface instead of typeof)
```
type CreationType = typeof SomeModel.CreationType
type SnapshotType = typeof SomeModel.SnapshotType
type Type = typeof SomeModel.Type
export interface ISomeModelCreation extends CreationType {}
export interface ISomeModelSnapshot extends SnapshotType {}
export interface ISomeModel extends Type {}
```
then using ISomeModelCreation, ISomeModelSnapshot, ISomeModel 

or I guess TS could get their act together and optimize that :)

Bnaya:
```ts
import { types, getSnapshot } from "mobx-state-tree";

const Square = types
    .model(
        "Square",
        {
            width: types.number
        }
    )

const a = getSnapshot(Square)
```
I'm getting this error for the getSnapshot call
`
Type 'IModelType<ModelPropertiesDeclarationToProperties<{ width: ISimpleType<number>; }>, {}, ModelCrea...' has no properties in common with type 'IStateTreeNode<any, {}>'.`

I'm probably missing something :)

mweststrate:
Yes, you are snapshoitting a type, not an instance

Op do 26 jul. 2018 20:38 schreef Bnaya Peretz <notifications@github.com>:

Bnaya:
Of Course😂 thanks

marvwhere:
@xaviergonz I tried around, but still not helping. The strange thing is that it is working with 3.0.0. I used that version to remove all lint errors which popped up through the breaking changes. then went back to 3.0.2 - still not working (i mean it is compiling, but the linter is not working)

I try to play around with it a little bit and try to find out which commit breaks it.

mweststrate:
Sounds more like an issue to report at the linter library?

Op vr 27 jul. 2018 11:26 schreef Javier Gonzalez <notifications@github.com>:

marvwhere:
kind of yes =/ 

`NODE_OPTIONS=--max-old-space-size=4096 && npx tsc -p .` command 

working fine in 3.0.0 and does nothing in 3.0.1

Issue #820, opened 2018-05-16
make `getEnv` fail when root has not been initialized with context


mweststrate:
give a clear error with what root (type) it didn't find an environment for. This will make getEnv fail if it is used to early, for example in `afterCreate` instead of `afterAttach`

PMExtra:
I have the same problem. I want to refresh some data when the store created. But the api in environment has not initialed.

PR #938, opened 2018-07-20
throw error when env not found

t49tran:
This PR address the issue: #820 .

--- Throw an error from `getEnv` when there is no env found.
--- Update the test to work with the changes.


mweststrate:
@t49tran please check the build, this seems to break several middlewares

t49tran:
@mweststrate , I fixed the problem with UndoManger, can you review it now ?

packages/mst-middlewares/src/undo-manager.ts
```diff
@@ -135,7 +135,11 @@ const UndoManager = types
                 self.undoIdx = self.history.length
             },
             afterCreate() {
-                targetStore = getEnv(self).targetStore ? getEnv(self).targetStore : getRoot(self)
+                try {
+                    targetStore = getEnv(self).targetStore
+                } catch {
+                    targetStore = getRoot(self)
```
mweststrate: This is an expensive and ugly way to check for the existence of an environment. I think we should introde `hasEnv(self)` to accompany this change

packages/mobx-state-tree/src/core/mst-operations.ts
```diff
@@ -710,8 +710,8 @@ export function getEnv<T = any>(target: IAnyStateTreeNode): T {
     }
     const node = getStateTreeNode(target)
     const env = node.root._environment
-    if (!!!env) return EMPTY_OBJECT as T
-    return env
+
+    return env || fail(`Failed to find the environment of ${getStateTreeNode(target)}`)
```
mweststrate: Maybe show the path in the error as well?

t49tran: I added the `path` in, do you think that's sufficient ?

packages/mst-middlewares/src/undo-manager.ts
t49tran: I introduced `hasEnv`, do you want to have a look at that function ?

t49tran:
Hi @xaviergonz, @mweststrate I just merge master to this PR and resolve all the conflicts, is throwing an error when `environment` is not available inside `getEnv` and add a `hasEnv` function still something we want to have ?

mweststrate:
Personally I think it is fine, the only thing is that this will probably be open for a while, as no new major is planned at this point, and this one is too small to push for a new major?

xaviergonz:
we could have a config function and make this false by default on v3 then true on v4

xaviergonz:
That being said I'd love to have a new version for all the TypeScript 3 improvements/fixes, such as the flow one

jamonholmgren:
@t49tran Is this something you're still interested in getting merged? If so, would it make sense to resolve conflicts or open a new PR and re-implement on top of the current state of the code base?

jamonholmgren:
Looking for a volunteer to push this one forward. What needs to happen:

1. Pull out relevant changes in this PR and open a new one off of the latest `master`
2. Test everything, make sure it still works
3. Let me know (probably [on Twitter](https://twitter.com/jamonholmgren)) that it's ready to merge.

coolsoftwaretyler:
I'm gonna assign this one to myself. Tryna get my head around the codebase by moving some of these old PRs forward.

coolsoftwaretyler:
I'm going to close this in favor of https://github.com/mobxjs/mobx-state-tree/pull/2039, but thank you to @t49tran for the initial work! I merely took the work and re-applied it on top of a more recent version of the repo.

PR #945, opened 2018-07-25
use interfaces wherever possible, they compile/parse faster

xaviergonz:
this small pull request is to change some types to interfaces wherever possible, since they are compiled/parsed faster by TS


xaviergonz:
while at it also converted the last tests files that were in js to ts

mweststrate:
@xaviergonz ok, let me know when you are comfortable with the changes!

xaviergonz:
that should be it 👌

mweststrate:
Thanks!

PR #946, opened 2018-07-26
Removed unused code from the example

ematipico:
`when` is not used here, I thought it can be removed


mweststrate:
Yep, thanks!

PR #950, opened 2018-07-27
create is smarter, reference snapshots fixed

### LINKED ISSUE #920: create with/without parameters is allowed in TS in cases it shouldn't
* [x] I think something is not working as it should.
  * [x] Describe expected behavior
  * [x] Describe observed behavior

https://codesandbox.io/s/yw8jn0zw51

#### Expected behavior

The TS type should be the same as with the commented line

#### Observed behavior

Properties becomes optional

xaviergonz:
basically create is now smart enough to know if it requires a snapshot or not
also fixed ref types so their snapshot type is number or string
some other little type fixes

Fixes #920

Btw, the way for create to detect if the snapshot is required or not is kinda "fragile"
I have the feeling it would be way easier to do if create worked like
```create(type, snapshot?)```

since then create could just have two overrides, one with snapshot being optional if the type includes the optional flag, other with a mandatory snapshot if not

as a side note I have a plan to remove those humongous union/compose overrides once ts3.0 comes out if it is ok to make ts3 the min version :)
^ **that is actually now done on #952, which includes this PR inside**

### CODE REVIEW

File: packages/mobx-state-tree/src/core/type/type.ts [hunk 1]
```diff
@@ -109,10 +129,11 @@ export abstract class ComplexType<C, S, T> implements IType<C, S, T> {
     }
 
     @action
-    create(snapshot: C = this.getDefaultSnapshot(), environment?: any) {
+    create = ((snapshot: C = this.getDefaultSnapshot(), environment?: any) => {
```
xaviergonz: is there a way to cast a class method to any? I had to convert it to a class property just for this reason

packages/mobx-state-tree/src/types/utility-types/frozen.ts
```diff
@@ -52,7 +52,7 @@ export function frozen<C>(subType: IType<C, any, any>): IType<C, C, C>
 export function frozen<T>(
     defaultValue: T
 ): IType<T | undefined | null, T, T> & { flags: TypeFlags.Optional }
-export function frozen<T = any>(): IType<T, T, T> // do not assume undefined by default, let the user specify it if needed
+export function frozen<T>(): IType<T, T, T> // do not assume undefined by default, let the user specify it if needed
```
xaviergonz: breaking-ish change, since create can only be smart if a type is given (although it also works with any), but I think it is better to make the user make an informed decision

xaviergonz: solved with ts3 in the pull request about ts3

packages/mobx-state-tree/src/core/type/type.ts
xaviergonz: solved with ts3 in the pull request about ts3

xaviergonz:
oh and fixes a bug I saw about isPrimitiveType

mweststrate:
Closed for now in favor of #952

PR #952, opened 2018-07-28
[WIP] updated deps to latest versions, fixed jest on windows, better typings using TS3.0

xaviergonz:
This pull request:
- updates all deps to the latest versions (including dev ones)
- makes linting work once more, updated rules that skip rules that are covered by prettier
- adds a prettierrc file to the root
- tests are now divided into __tests__/core and __tests__/perf, fixed jest tests on windows (after the last update tests were working on CI but not on my windows machine)
- [breaking change?] using TS 3.0
- enumeration is type shorter, same for union and compose
- create function is a bit more robust

Also includes everything from pull request #950

It is a WIP basically because TS 3.0 is still on RC phase for a few more days


Bnaya:
You replace `any` with `unknown` where's makes sense? nice :)

xaviergonz:
**Update: No longer a WIP since now it uses the final version of TS3** 👍

xaviergonz:
@mweststrate thoughts?

mweststrate:
Hey Xavier, sounds great! Im currently on holiday,  so hope to check them
out next week. Or if @kga or mattia approves that's fine by me as well!

Op di 31 jul. 2018 20:34 schreef Javier Gonzalez <notifications@github.com>:

xaviergonz:
ok, just wanted to make sure you were ok with the requirement of ts 3 :)

xaviergonz:
I was wondering, should this be a major version increase?
If so, should this go into some kind of beta branch and be released as mobx-state-tree@4.0.0-beta ?

xaviergonz:
Note to self: This PR needs to be revised/updated after the last changes

mweststrate:
I think the community is quite divided on the questions whether requiring a TS upgrade is majore breakage or not. It kinda is, but on the other hand, I've seen packages in the past doing it as minor as well. I could live with it being a minor, but probably MobX and MobX-react should support TS 3 first officially as well (and this will most probably be done in minors to avoid a version hell). In MST it would be less of a problem to have it as major as it is not that often an dependency of other packages.

Issue #926, opened 2018-07-17
getSnapshot() doesn't traverse all nodes in the tree


bourquep:
I have:

* [ ] A conceptual question. 
  * [ ] I've checked documentation and searched for existing issues
  * [ ] I tried the [spectrum channel](https://spectrum.chat/?t=dad48299-3dfc-4e10-b6da-9af1e39498a3) first
* [x] I think something is not working as it should.
  * [x] I've checked documentation and searched for existing issues
  * [x] I've made sure your project is based on the latest MST version
  * [x] Fork [this](https://codesandbox.io/s/jlnmjqr7xv) code sandbox or another minimal reproduction. 
  * [x] Describe expected behavior
  * [x] Describe observed behavior
* [ ] Feature request
   * [ ] Describe the feature and why you feel your feature needs to be generically solved inside MST
   * [ ] Are you willing to (attempt) a PR?
   
_Not following the above template might result in your issue being closed without further notice_

---

It seems like `getSnapshot()` is broken in v3.0.0. 

Running this repro code:

```javascript
const { types, getSnapshot } = require("mobx-state-tree");

const Nested = types.model("Nested", {
    foo: "bar",
    name: "",
}).postProcessSnapshot(snapshot => ({ ...snapshot, "fooz": "barz" }));

const Root = types.model("Root", {
    foo: "bar",
    name: "",
    nested: types.array(Nested)
});

const nested = Nested.create({});
const root = Root.create({nested: [nested]});

console.log("Snapshot of nested node includes 'fooz' attribute added in post-process:");
const nestedSnap = getSnapshot(nested);
console.dir(nestedSnap);

console.log();

console.log("But snapshot of root node does not include post-processed attribute:");
const rootSnap = getSnapshot(root);
console.dir(rootSnap.nested[0]);
```

Yields this output:

```
Snapshot of nested node includes 'fooz' attribute added in post-process:
{ foo: 'bar', name: '', fooz: 'barz' }

But snapshot of root node does not include post-processed attribute:
{ foo: [Getter/Setter], name: [Getter/Setter] }
```

Running the same code against MST v2.2.0 (after moving `postProcessSnapshot()` to an action) yields this output:

```
Snapshot of nested node includes 'fooz' attribute added in post-process:
{ foo: 'bar', name: '', fooz: 'barz' }

But snapshot of root node does not include post-processed attribute:
{ foo: 'bar', name: '', fooz: 'barz' }
```

[...]

Issue #947, opened 2018-07-26
Feature - types.snapshotProcessor(subModel, { preProcessor, postProcessor})


mweststrate:
These features leads often to confusion:

* preProcessors are run multiple times, and expensive for all the object allocations
* these hooks don't fit nicely in the rest of the MST model, as they are unlike all other model things defined 
statically
* it kills optimizations like `getSnapshot` just returning the original snapshot when the node was never touched (see #926 for an example)

What would be the impact? What are typical usage patterns and what are the alternatives? For example: introduce views that use `getSnapshot` as replacement for postProcessing

mattruby:
I agree that preProcessors are strange and cause issues.  I do most of my data manipulation outside of MST.  

My current common use case is renaming ref id fields over to something more ref like:
`snapshot.mediaList = snapshot.mediaKeys`
I know others are doing the same.  
I've tried to make a model that could handle some of that work but we can't compose preProcessors.

mweststrate:
Alternative, introduce a type that handles this: `types.snapshotProcessor(subModel, { preProcessor, postProcessor})`, that could probably mix this better in the lifecycle and of building trees and avoid needing static analysis of the subtypes used? cc @k-g-a

mattiamanzati:
@mweststrate definetely agree for the latter one, would be neat to do the same for any lifecycle event; as those can be applied to any type; not just models

k-g-a:
I love the dedicated type approach and could even implement it in mid August! For now there is a mess with nested pre/post processors. That’s the reason why I personally prefer modifying API data (used as snapshot) in a tiny custom “layer” outside the MST.

bourquep:
I've been able to get rid of our snapshot pre/post processing requirements, and it indeed makes things a bit clearer/more predictable.

ryedin:
We rely very heavily on these functions to support API endpoints that can be fat. We use these hooks for normalizing those related data into their top level stores and replacing the fat child props with references. We also use them for the obvious use cases of handling date conversions (as well as parsing floating point values to/from strings).

Any change here would completely break our architecture, so please make sure the alternative has full parity (please!)

k-g-a:
> We also use them for the obvious use cases of handling date conversions (as well as parsing floating point values to/from strings).

It looks like those cases are better covered by [custom types](https://github.com/mobxjs/mobx-state-tree/blob/master/API.md#typescustom).

ryedin:
[...]

PR #959, opened 2018-08-07
PostProcess non-initialized children

k-g-a:
Fixes [#926](https://github.com/mobxjs/mobx-state-tree/issues/926)

### DIFF
--- a/packages/mobx-state-tree/package.json
+++ b/packages/mobx-state-tree/package.json
@@ -9,13 +9,14 @@
   "scripts": {
     "build": "tsc && cpr lib dist --delete-first --filter=\\\\.js && yarn rollup",
     "rollup": "rollup -c",
-    "test": "cross-env NODE_ENV=development jest --ci && cross-env NODE_ENV=production jest --ci && yarn run test:cyclic && yarn run test:mobx4",
-    "test:perf": "yarn build && yarn jest --testRegex 'test/perf/fixture*' && TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\"}' /usr/bin/time node --expose-gc --require ts-node/register test/perf/report.ts",
+    "jest": "jest --projects='test'",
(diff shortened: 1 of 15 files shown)


k-g-a:
This one now also supersedes [PR #963](https://github.com/mobxjs/mobx-state-tree/pull/963).
Those two combined give us an opportunity to enrich ```initialSnapshot``` with optional values only when it's required (it had been done during construction previously).
I wanted to measue and impact, but as soon as I hit the devtools i saw that it's actually x2 degradation in terms of CPU time and memory. 
The reason is that we always get actual snapshot of root, which in turn collects ```_initialSnapshots``` of it's children. This used to be a simple object return, but was converted to the loop with ```childNode.snapshot``` call for the purpose of this PR. ```snapshot``` is computed, so we get much of new arrays, tons of concatinated strings and so on and so forth...
I'll try to improve this part within this week.

k-g-a:
Commit message outlines the most part, but I'll copy it for convinience:
- Node.```getSnapshot()```: uncached version of ```@computed get snapshot()``` - by now used to get initial snapshots without creating ```@computed```'s overhead
- ComplexType.processInitialSnapshot: used to delegate all type-specific code from ObjectNode to concrete type - all the ```this.type instanceof ModelType``` (and similar) are now removed from ObjectNode
- Initial performance restored: creation times are similar, little consumed memory raise
- removed ```cloneSnapshotIfNeeded``` (part of PR #963) from type.create as it's not needed anymore
- initialSnapshot is frozen: to prevent furhter accidental mutations as it happened with optional props

By now I'd prefer this PR over #963 as it combines both + cleanup. 
I'd also perfer this one over #962 , as those will conflict, but this one solves issues.

mweststrate:
Thanks @k-g-a! Will merge this one to fix the current problems. I think however that we should still keep #947 open to solve this more elegantly in the next major, as the whole snapshot code paths are becoming insanely complex now :)

PR #962, opened 2018-08-08
[Internal] Cleanup 'createNode' and related codepaths

k-g-a:
This PR is purely internal. 
~~Basically it replaces ```createNewInstance``` + ```finalizeNewInstance``` combo with just one ```initializeInstance``` call.~~
Another step to make code cleaner was to remove ```createNewInstance```/```finalizeNewInstance``` from ```createNode```'s arguments, as those were taken from type anyway (inside ```type.instantiate```). As we have type inside ```ObjectNode```/```ScalarNode``` there is not need to pass those two methods 3 calls deep separately.
As a side effect this makes ```createNode``` and Node's constructor's calls less polymorphic and slightly reduces subsequent calls time.

### DIFF
--- a/packages/mobx-state-tree/src/core/node/create-node.ts
+++ b/packages/mobx-state-tree/src/core/node/create-node.ts
@@ -1,56 +1,28 @@
-import {
-    isStateTreeNode,
-    INode,
-    identity,
-    noop,
-    fail,
-    ObjectNode,
-    ScalarNode,
-    IType
-} from "../../internal"
+import { INode, fail, ObjectNode, ScalarNode, IType, getStateTreeNodeSafe } from "../../internal"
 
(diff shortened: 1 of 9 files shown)


k-g-a:
Ok, had to split create/finalize back, because ```ModelType```'s finalize calls hooks, which in turn want ```$treenode``` to be attached to ```self```... Well, it got too complex and 
This PR is basically a cleanup for ```createNode``` and related codepaths now.

k-g-a:
I advice merging #959 first - it will conflict with this one.

mweststrate:
@k-g-a could you rebase this branch and see what remains?

k-g-a:
@mweststrate, sure. I’ll have a look at Monday

Issue #961, opened 2018-08-08
Bug: applying snapshot getted by getSnapshot to union of models with different optional properties throws error


codewylle:
Applying snapshot getted by getSnapshot to union of models with different optional properties throws error:

```typescript
import { types, getSnapshot, applySnapshot } from "../src"

test("apply snapshot to union should not throw when union keeps models with different properties and snapshot getted by getSnapshot", () => {
    const Foo = types.model({ foo: 1 })
    const Bar = types.model({ bar: 1 })
    const U = types.union(Foo, Bar)

    const u = U.create({ foo: 1 })
    applySnapshot(u, getSnapshot(Bar.create())) // THROW error
})
```

When union type validates snapshot, it validates all subtypes until one of them fits to snapshot. But validation of model type can add property to snapshot if value of optional model property is absent in snapshot (is it normal?):

```typescript
import { types } from "../src"

test("model creating should not change snapshot", () => {
    const M = types.model({ foo: 1 })
    const o = {}

    M.create(o)
    expect(o).toEqual({}) // o is { foo: 1 }
})
```

getSnapshot returns frozen object. When mst tries to add property to frozen object in validation of model it throws error `Cannot add property *, object is not extensible`

k-g-a:
I'd say this is a bug. Will make PR today.

mweststrate:
@k-g-a thanks! Fix has been released as 3.1.0

PR #963, opened 2018-08-08
Do not spoil snapshots during 'create'

k-g-a:
FIxes [#961](https://github.com/mobxjs/mobx-state-tree/issues/961)
- Clone mutable snapshots in .create()
- Apply optional values only for initialSnapshot (also happened during typecheck)


mweststrate:
Closed in favor of #959 as suggested for now

Issue #968, opened 2018-08-15
3.2.0 object-node.d.ts subpathAtom error


garygentry:
With a clean npm init project:
    "mobx": "5.0.3",
    "mobx-state-tree": "3.2.0"

> node_modules/mobx-state-tree/dist/core/node/object-node.d.ts
> Object.subpathAtom

 has this value:

```
subpathAtom: import("../../../../../../../../VSProjects/github/mobx-state-tree/node_modules/mobx/lib/core/atom").IAtom;
```
installed 3.1.1 and it looks like this:
```
subpathAtom: import("mobx/lib/core/atom").IAtom;
```

![image](https://user-images.githubusercontent.com/12444042/44133086-b4ee335a-a02b-11e8-9ab7-7600b638b4fc.png)

Is this a problem with the build?

k-g-a:
Actually, there are no explicit imports of ```IAtom``` at 3.1.1 either. 
Nevertheless, adding it back really changes ```subpathAtom``` import from relative back to normal.

xaviergonz:
Should be fixed in 3.2.1 now, sorry about that!

xaviergonz:
@RainerAtSpirit what kind of issues are you running into with TS3.0? just curiours

garygentry:
3.2.1 seems to work.  I was in the process of upgrading to typescript 3.0.1 (from 2.7.2).  So I'm clear, any issues with mst and typescript 3.0.1?

RainerAtSpirit:
@xaviergonz TS3.0.1 produces a bunch of Typing errors in our projects. Haven't found the root issue yet, so we're down to TS2.8.4. Probably related to the fact that we're creating common things like stores as separate private npm packages and then consume those in our projects. d.ts output changed between 2.8.4/2.9.2 and 3.0.1 and there are some issues related to that on the typescript repo.

PR #969, opened 2018-08-15
Explicit IAtom import added, marked as exception for IDE's

k-g-a:
Seems to fix [#968](https://github.com/mobxjs/mobx-state-tree/issues/968).


xaviergonz:
looks good, seems like it was my fault since I made the build with windows and ts 2.9.2 with windows seems to generate wrong relative imports from time to time. I'll add it and release a 3.2.1

xaviergonz:
In 3.2.1, thanks @k-g-a !

PR #974, opened 2018-08-17
Cast second use case docs

xaviergonz:
some docs for the cast where cast can be used to use instances inside snapshots


xaviergonz:
I will just merge it since it is just a few line of docs being added

PR #975, opened 2018-08-17
updated deps to latest versions, fixed jest on windows, better typings for enum, compose, union, literal

xaviergonz:
This pull request is like https://github.com/mobxjs/mobx-state-tree/pull/952 except it doesn't require TS3 and doesn't have the improved create function:

- updates all deps to the latest versions (including dev ones)
- makes linting work once more, updated rules so that linting skips rules that are covered by prettier
- adds a prettierrc file to the root, removed them from the vscode config
- tests are now divided into tests/core and tests/perf, fixed jest tests on windows (after the last update tests were working on CI but not on my windows machine)
- enumeration type declaration is shorter and better
- better union and compose types + fixed generation scripts
- improved literal typings


.travis.yml
```diff
@@ -18,4 +18,4 @@ script:
 after_success:
   - cat ./packages/mobx-state-tree/coverage/lcov.info|./packages/mobx-state-tree/node_modules/coveralls/bin/coveralls.js
 node_js:
-  - 9.3.0
+  - "node"
```
mweststrate: what does this mean?

xaviergonz: latest lts version of node

mweststrate:
@xaviergonz could you check the merge conflict? Beyond that I think it is ready to merge :)

.travis.yml
xaviergonz: I mean make Travis use the latest lts version of node, a VM with node 10.x and npm 6 installed in this case

mweststrate:
I'll cut a release tomorrow, or if you feel for it you can do it today, but I think the changelog is not complete yet.

xaviergonz:
@mweststrate I can do it, but only for mobx-state-tree, I don't have publish permissions for mst-middlewares

xaviergonz:
would it be a minor or a patch version btw?

mweststrate:
I think patch will do

Op ma 20 aug. 2018 om 17:27 schreef Javier Gonzalez <
notifications@github.com>:

mweststrate:
Oh I think I added mst-middleware permissions in the mean time, so that
should work out now :)

Op ma 20 aug. 2018 om 17:28 schreef Michel Weststrate <mweststrate@gmail.com

xaviergonz:
@mweststrate v3.2.2 is now out for both packages 👍

PR #978, opened 2018-08-21
better enum typings for ts2

xaviergonz:
Improved the enumeration typings for TS2, making them shorter and now support up to an infinite number of arguments

The only place where the typings default to a generic string is when the options array is given from a variable rather than directly, in this case it will default to a generic string (which was already the case anyway)

This is
```
types.enumeration(["a", "b"]) // type is "a" | "b"
const arr = ["a", "b"]
types.enumeration(arr) // type is string since the type of arr is inferred to be string[]
```


packages/mobx-state-tree/src/types/utility-types/enumeration.ts
```diff
@@ -140,7 +28,7 @@ export function enumeration<E extends string>(name: string, options: E[]): ISimp
  * @param {string[]} options possible values this enumeration can have
  * @returns {ISimpleType<string>}
  */
-export function enumeration(name: string | string[], options?: any): ISimpleType<any> {
+export function enumeration(name: string | string[], options?: any): ISimpleType<string> {
```
mweststrate: Just double checking; `type.enumeration(["a","b"]).create("c")` will still result in a compile error?

xaviergonz: yes it will be covered by the first overrides
this override will only trigger in this case
```
const arr = ["a", "b"]
types.enumeration(arr).create("c") // won't complain since arr is infered to be string[]
```

which could be worked around using this, but it is kind of pointless
```
const arr: ["a", "b"] = ["a", "b"]
types.enumeration(arr).create("c") // in this case the compiler will complain
```

The only way to fix it (which would require TS3) is to make enumeration take all the options as flat parameters rather an array like union / compose does

Issue #982, opened 2018-08-22
model extendClass and extendDecorate idea


xaviergonz:
I have:

* [x] Feature request
   * [x] Describe the feature and why you feel your feature needs to be generically solved inside MST
   * [x] Are you willing to (attempt) a PR?
   
What about a extendClass model method that works like this?
```ts
    const M = types.model({ x: 5 }).extendClass(self => {
        class E {
            @modelState
            localState = 3

            @modelView
            get x2() {
                return self.x * 2
            }

            @modelView
            xBy(by: number) {
                return self.x * by
            }

            @modelAction
            setX(x: number) {
                self.x = x
            }

            @modelAction
            setXBy(x: number) {
                this.setX(this.xBy(x))
            }

            @modelAction
            setLocalState(x: number) {
                this.localState = x
            }
        }
        return E
    })
```

**pros:**
- basically this allows interactions between views/actions/state in an easier way
- doesn't break any compatibility, since it is a new method
- might make typings faster since it uses a class?
- PR ready :D

**cons:**
- classes need to have given a name (TS doesn't support (yet?) decorators inside anonymous classes), so this won't work
```ts
.extendClass(self => (
  class {
    ...
  })
)
```
- bound methods give problems since they bind to the class instance (which is not self)
  - non arrow function methods get bound to self properly though

**future:**
- if TS ever adds decorator support for any object properties then a similar extendObject method could be created (like extendDecorate but with actual decorators)

Bnaya:
It would also improve the performance of server!

xaviergonz:
how come?

Bnaya:
read https://github.com/mobxjs/mobx-state-tree/issues/705#issuecomment-385341525 and comments afterward

xaviergonz:
would the decorate version help as well?

xaviergonz:
PR for extendDecorate there as well, I think I prefer that one because while uglier it doesn't allow the programmer to make as many mistakes

xaviergonz:
@Bnaya  I just read the thread. I think that while extendClass might help with the actions/views/state it wouldn't help with the props part

Either way I think that both (extendClass and extendDecorate) would make types a tiny bit faster since:
a) usually only use one of these would be used for each model
b) rather than an union of actions & views & state it just uses the whole object interface

xaviergonz:
That being said, maybe the props part typing could be speed up by using something like
model(new Props({ x: 5; ... })) ?

[...]

PR #983, opened 2018-08-22
extendClass [docs missing]

xaviergonz:
first idea of https://github.com/mobxjs/mobx-state-tree/issues/982


xaviergonz:
Guess I'll close this and the PRs now that TS type completion got faster and "this" can be safely used inside views/actions (and it is not like views will ever use actions, so :) )

PR #985, opened 2018-08-22
Fix for exception when destroying children of types.maybe

dsabanin:
Since ```types.maybe``` type in MST 3.0 no longer includes ```null```, trying to assign parent's property that was containing the child that is being removed would throw ```value `null` is not assignable to type:```.

```typescript
const A = types.model({id: types.reference, myChild: types.maybe(B)});
const B = types.model({id: types.reference, name: types.string});
destroy(B) // throws: value `null` is not assignable to type: `(B | undefined)`.
```

Setting parent's property to ```undefined``` fixes the issue, and should not cause any problems since both ```types.maybe``` and ```types.maybeNull``` include ```undefined```.

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/object.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/object.test.ts
@@ -605,10 +605,10 @@ if (process.env.NODE_ENV !== "production") {
         unprotect(b)
         expect(() => {
             detach(b.a)
-        }).toThrowError(/Error while converting `null` to `A`/)
+        }).toThrowError(/Error while converting `undefined` to `A`/)
         expect(() => {
             destroy(b.a)
-        }).toThrowError(/Error while converting `null` to `A`/)
+        }).toThrowError(/Error while converting `undefined` to `A`/)
     })
 }
 test("it should be possible to share states between views and actions using enhance", () => {
--- a/packages/mobx-state-tree/__tests__/core/object.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/object.test.ts
@@ -620,6 +620,8 @@ if (process.env.NODE_ENV !== "production") {
             detach(a)
             destroy(a)
         }).not.toThrow()
+        expect(b.a).toBe(undefined)
+        expect(getSnapshot(b).a).toBe(undefined)
     })
     test("it should be possible to remove a node from a parent if it is defined as type maybeNull ", () => {
         const A = types.model("A", { x: 3 })
@@ -631,6 +633,8 @@ if (process.env.NODE_ENV !== "production") {
             detach(a)
             destroy(a)
         }).not.toThrow()
+        expect(b.a).toBe(null)
+        expect(getSnapshot(b).a).toBe(null)
     })
 }
 test("it should be possible to share states between views and actions using enhance", () => {
--- a/packages/mobx-state-tree/__tests__/core/object.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/object.test.ts
@@ -610,6 +610,30 @@ if (process.env.NODE_ENV !== "production") {
             destroy(b.a)
         }).toThrowError(/Error while converting `undefined` to `A`/)
     })
+    test("it should be possible to remove a node from a parent if it is defined as type maybe ", () => {
+        const A = types.model("A", { x: 3 })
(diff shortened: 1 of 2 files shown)


xaviergonz:
what I mean is could you include both cases in the unit test?

dsabanin:
Yep, done. Can you check it out? Thanks!

packages/mobx-state-tree/__tests__/core/object.test.ts
```diff
@@ -605,10 +605,34 @@ if (process.env.NODE_ENV !== "production") {
...
+        }).toThrowError(/Error while converting `undefined` to `A`/)
+    })
+    test("it should be possible to remove a node from a parent if it is defined as type maybe ", () => {
+        const A = types.model("A", { x: 3 })
+        const B = types.model("B", { a: types.maybe(A) })
+        const b = B.create({ a: { x: 7 } })
+        unprotect(b)
+        expect(() => {
+            detach(b.a!)
+        }).not.toThrowError(/Error while converting `undefined` to `A`/)
```
xaviergonz: I think in this case it is better to use `not.toThrowError()` without the string argument to make sure it doesn't throw any kind of errors, not just an error with that error message

also feel free to put detach and destroy inside the same expect block in this case
also I think just destroy is good enough since it detachs it before destroying it IIRC

same for the other one

dsabanin: Makes sense, thanks!

dsabanin:
Thanks Javier!

xaviergonz:
thanks to you! :D

dsabanin:
If the fix is acceptable, would it be possible to cut out a patch release with it?

xaviergonz:
most probably, yeah

k-g-a:
I’ll check on Monday, but how come that “maybeNull” can be “undefined”? Seems strange...

xaviergonz:
guess that's because maybeNull is optional, which means that when undefined is assigned from the snapshot it will default to null on the instance?

packages/mobx-state-tree/__tests__/core/object.test.ts
```diff
@@ -605,10 +605,32 @@ if (process.env.NODE_ENV !== "production") {
...
+    test("it should be possible to remove a node from a parent if it is defined as type maybe ", () => {
+        const A = types.model("A", { x: 3 })
+        const B = types.model("B", { a: types.maybe(A) })
+        const b = B.create({ a: { x: 7 } })
+        unprotect(b)
+        expect(() => {
+            const a = b.a!
+            detach(a)
+            destroy(a)
+        }).not.toThrow()
```
mweststrate: assert `b.a === undefined` and `getSnapshot(b).a === undefined`?

packages/mobx-state-tree/__tests__/core/object.test.ts
```diff
@@ -605,10 +605,32 @@ if (process.env.NODE_ENV !== "production") {
...
+    test("it should be possible to remove a node from a parent if it is defined as type maybeNull ", () => {
+        const A = types.model("A", { x: 3 })
+        const B = types.model("B", { a: types.maybeNull(A) })
+        const b = B.create({ a: { x: 7 } })
+        unprotect(b)
+        expect(() => {
+            const a = b.a!
+            detach(a)
+            destroy(a)
+        }).not.toThrow()
```
mweststrate: assert `b.a === null` and `getSnapshot(b).a === null`?

mweststrate:
Added suggestions for further assertions, beyond that LGTM. Because `maybeNull` is always optional, it can be instantiated _from_ `undefined` in a snapshot (otherwise we would have to distinguish the absence of a property in a snapshot versus the property being present but `undefined`, but that get's real hairy really quickly).

However, converting to a snapshot should always yield the property be present and `null`. In otherwords, `maybeNull<X>` should behave according the type signature `IType<null | undefined | Snapshot<X>, null | Snapshot<X>, null | Instance<X>>`

xaviergonz:
Added requested checks to the unit tests, since they are passing merged, thanks!

dsabanin:
Thank you everybody!

dsabanin:
Is it possible to push a patch release for that? We're stuck with our forked packages for now and I'd love to switch back.

PR #990, opened 2018-08-25
Smaller .d.ts and fixing doc build

xaviergonz:
- reduces and simplifies .d.ts files by a lot by using the `@internal` jsdoc decorator to avoid exposing internals in TS types (hopefully this will also help reducing times when emit declarations is used)
- fixes doc building with the latest version of document by using the `@private` jsdoc decorator and importing/exporting public API in index.ts
- adds a lint-docs command to mobx-state-tree package
- fixes build-docs command on windows
- fixes some warnings related to using the latest version of rollup with outdated config
- renamed typecheckPublic to typecheck and typecheck to typecheckInternal
- typing system (TS) now relies on `interface OptionalProperty { optional: true }` rather than TypeFlags.Optional
- updates some deps to the latest versions
- added docs for isXType family of functions


k-g-a:
Sounds great!
BTW, how did you fix docs on windows?

xaviergonz:
just used shx to run the rm command

packages/mobx-state-tree/src/core/type/type.ts
```diff
@@ -45,35 +49,116 @@ export enum TypeFlags {
 
 export interface IType<C, S, T> {
     name: string
-    flags: TypeFlags
```
mweststrate: `validate`, `is`, `describe` can remain public, they are very useful to just run validation

mweststrate:
Few comments, further LGTM :) Hope it helps with the typecheck performance!

packages/mobx-state-tree/src/core/type/type.ts
xaviergonz: done!

xaviergonz:
fixed the few comments.
while I'm not really sure it will make it much faster, one good side-effect is that user's won't see any internals of the library and therefore auto-completion should be safer to use

mweststrate:
Cool! Feel free to merge

Op zo 26 aug. 2018 17:58 schreef Javier Gonzalez <notifications@github.com>:

Issue #994, opened 2018-08-29
3.2.4 - TypeScript error on types.reference


birkir:
I have:

mobx-state-tree: 3.2.4
typescript: 3.0.1

* [x] I think something is not working as it should.
  * [x] I've checked documentation and searched for existing issues
  * [x] I've made sure your project is based on the latest MST version
  * [x] Describe observed behavior
 
---

This is my case:

```ts
import { types } from 'mobx-state-tree';
import Cinemas from './Cinemas';

export const Cinema = types.model('Cinema', {
  id: types.identifier,
  name: types.maybe(types.string),
});

//               Here is a typescript error:   ↓
//                                             ↓
export const CinemaReference = types.reference(Cinema, {
  get(identifier: string) {
    return Cinemas.getOrLoadById(identifier) as typeof Cinema.Type || [];
  },
  set(value: typeof Cinema.Type) {
    return value.id;
  },
});
```

This is the error I get:
```java
Argument of type 'IModelType<ModelPropertiesDeclarationToProperties<{ id: ISimpleType<string>; name: IComplexType<string | undefined, string | undefined, string | undefined> & OptionalProperty; }>, {}, ModelCreationType<...>, ModelSnapshotType<...>, ModelInstanceType<...>>' is not assignable to parameter of type 'IAnyModelType'.
  Types of property 'views' are incompatible.
    Type '<V extends Object>(fn: (self: ModelInstanceType<ModelPropertiesDeclarationToProperties<{ id: ISimpleType<string>; name: IComplexType<string | undefined, string | undefined, string | undefined> & OptionalProperty; }>, {}, ModelCreationType<...>, ModelSnapshotType<...>>) => V) => IModelType<...>' is not assignable to type '<V extends Object>(fn: (self: any) => V) => IModelType<any, any, ModelCreationType<any>, ModelSnapshotType<any>, any>'.
      Type 'IModelType<ModelPropertiesDeclarationToProperties<{ id: ISimpleType<string>; name: IComplexType<string | undefined, string | undefined, string | undefined> & OptionalProperty; }>, any, ModelCreationType<...>, ModelSnapshotType<...>, any>' is not assignable to type 'IModelType<any, any, ModelCreationType<any>, ModelSnapshotType<any>, any>'.
        Types of property 'preProcessSnapshot' are incompatible.
[...]

PR #995, opened 2018-08-29
fix for reference type issue in ts3 (#994)

xaviergonz:
also fixes IType sometimes being inferred as IComplexType


xaviergonz:
PS I can cut a new minor version also if desired

packages/mobx-state-tree/src/core/type/type.ts
```diff
@@ -156,6 +156,9 @@ export type Primitives = ModelPrimitive | null | undefined
 export type TAndInterface<T, I> = (Exclude<T, Primitives> & I) | Extract<T, Primitives>
 
 export interface IComplexType<C, S, T> extends IType<C, S, T> {
+    // fake, only used for tying
```
mweststrate: typo

xaviergonz:
typo fixed 👍 
@mweststrate should I cut a new minor version?

Issue #993, opened 2018-08-27
Performing Actions on recently created objects on a reference does not trigger `onSnapshot` for the store.


NickMele:
I have:

* [x] I think something is not working as it should.
  * [x] I've checked documentation and searched for existing issues
  * [x] I've made sure your project is based on the latest MST version
  * [x] Fork [this](https://codesandbox.io/s/jlnmjqr7xv) code sandbox or another minimal reproduction. 
  * [x] Describe expected behavior
  * [x] Describe observed behavior
   
I am running into an issue with performing an action on a reference not triggering the "onSnapshot" call in our store.

I am able to reproduce this in a simple store example: https://codesandbox.io/s/rrk8wojnp4

**Observed behavior:**
When loading the initial store, I want to perform an action on one selected todo. this selected todo is stored as a reference to a todo in the todos array. When I perform the `toggle` action using the reference you can see that there is no snapshot logged. Just a second later if I perform that same action directly on the item instead of via the reference, that action gets logged and can be stored.

**Expected behavior:**
Performing an action on the referenced todo should trigger a snapshot, just the same as performing that action directly on the item.

I am still pretty new to this library so I might be misunderstanding how references are supposed to work or be used.

xaviergonz:
I'd say that's a bug, since strictly speaking "something" inside the store changed

xaviergonz:
makes me wonder also if actions are recorded as well when doing an action over a reference

NickMele:
@xaviergonz In my local project I am seeing the actions recorded in reactotron, and I can also get the patch when using `onPatch`. The only piece that does not seem to be working in `onSnapshot`

xaviergonz:
yep, just tested the same with https://codesandbox.io/s/qv6k9l89jj

xaviergonz:
A funny thing, if the object is modified a first time using the direct child *before* changing it using the reference then it works ok
https://codesandbox.io/s/qv6k9l89jj

@k-g-a might be something related to initialSnapshot?

xaviergonz:
summarizing:
create -> change referenced model (no snapshot) -> change using child model (snapshot)
create -> change using child model (snapshot) -> change referenced model (snapshot) -> change using child model (snapshot)

xaviergonz:
unit test
```ts
test("#993", () => {
    const Todo = types
        .model("Todo", {
            id: types.identifier,
            finished: false
        })
        .actions(self => ({
            toggle() {
[...]

PR #997, opened 2018-08-29
Fix the parent of a reference being undefined when the parent has never been accessed before

xaviergonz:
Fix for #993

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/object.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/object.test.ts
@@ -11,7 +11,8 @@ import {
     unprotect,
     types,
     setLivelynessChecking,
-    getParent
+    getParent,
+    hasParent
 } from "../../src"
 
 import { autorun, reaction, observable } from "mobx"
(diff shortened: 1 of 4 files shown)


xaviergonz:
Please review, I'm not sure if this is the best approach

xaviergonz:
actually, should I call. value from parent to child to ensure the creation order is done that way?

xaviergonz:
or maybe I just need to make getParent and family use `.value` before `. storedValue` to force initialization, hmm 🤔

k-g-a:
Hi! The fix seems efficient and elegant to me.
I only have doubt about parent.storedValue check in while. Could that be the case, when one of the intermediate parents is already initialized, while another parent upper the tree is not?
Maybe we should do:
```javascript
              if (matched) {
                    let parent = matched.parent
                    while (parent) {
                         if(!parent.storedValue) {
                          // accessing the value will ensure the node is initialized
                          // tslint:disable-next-line:no-unused-expression
                           parent.value
                         }
                         parent = parent.parent
                    }
                }
```

Another suggestion is to make ```_observableInstanceCreated```/```_createObservableInstance``` public and use them? So we make code self-documented and avoid need of comments like "accessing the value will ensure the node is initialized".

xaviergonz:
@k-g-a I put the check in place to avoid traversing the tree up to the root everytime a reference is used, so I'd like to keep that, but of course only if that covers all cases.

so about the cases, you think there might be cases where a parent might be initialized but it's parent won't? (skipping the current bug of course)

also about afterCreation/afterAttach order, should the node initialization be done from root to child or from child to root? or it doesn't matter.

another thing I'd like to try is fixing instead getParent so it initializes the node if not done already (including the whole chain to the root) rather than at reference time but the same doubt about initialization remains

xaviergonz:
ok, I just made the sure the order of the afterCreate/afterAttach events are the same when using a reference and the standard way

I still didn't remove the optimization though since I can't think of a way to access a node other than drilling through the store (l1.l2.l3.l4) or using a reference, (l1.selectedL4, which in turn initializes with the fix l2.l3.l4), but if you have any other ideas of when that could happen I will make sure to remove the optimization

packages/mobx-state-tree/__tests__/core/object.test.ts
```diff
@@ -706,3 +707,115 @@ if (process.env.NODE_ENV === "development")
...
+
+        l1.selected.toggle()
+        l1.l2.l3.l4.toggle()
+        expect(events).toEqual(expectedEvents)
+    }
+
+    // test 3, reference get parent should be available from the beginning
+    {
+        const l1 = createL1()
+        expect(getParent(l1.selected)).toBeTruthy()
```
mweststrate: Note that this returns the parent of the selected's target, not (so L4's parent)

mweststrate:
I am now wondering, can this issue also occur when using `getParent` in after attach?

packages/mobx-state-tree/__tests__/core/object.test.ts
xaviergonz: yep, but if I change it to test against l3, then that reference will initialize it already, so the unit test wouldn't do much

xaviergonz: I'm not sure if I follow about the after attach and getParent thing...

xaviergonz:
@mweststrate improved the unit test for getParent and added a unit test for after attach to check if the parent is there when it should be

mweststrate:
@xaviergonz looking good! Approved

Issue #1002, opened 2018-08-31
Throw/warn/snapshot-itize instances passed as argument to types.optional


xaviergonz:
I have:

* [ ] A conceptual question. 
  * [ ] I've checked documentation and searched for existing issues
  * [ ] I tried the [spectrum channel](https://spectrum.chat/?t=dad48299-3dfc-4e10-b6da-9af1e39498a3) first
* [ ] I think something is not working as it should.
  * [ ] I've checked documentation and searched for existing issues
  * [ ] I've made sure your project is based on the latest MST version
  * [ ] Fork [this](https://codesandbox.io/s/jlnmjqr7xv) code sandbox or another minimal reproduction. 
  * [ ] Describe expected behavior
  * [ ] Describe observed behavior
* [x] Feature request
   * [x] Describe the feature and why you feel your feature needs to be generically solved inside MST
   * [x] Are you willing to (attempt) a PR?
   
_Not following the above template might result in your issue being closed without further notice_

Today a coworker was telling me that it took them a while to find a bug related to passing an instance as argument to types.optional, since as soon as a second node of the same type was created it would crash because it was attached to the first node parent.

```
const someDefault = TModel1.create({...})

const TModel2 = types.create({ m1: types.optional(TModel1, someDefault) })
const a = TModel2.create({}) // ok
const b = TModel2.create({}) // boom!
```

My question is, should that particular case be prevented?
If so, what would the best course of action?
- A. snapshot-ize the instance - mostly non-breaking, but might loose the environment (not a big deal, since anyway children and parents have to have the same environment anyway)
- B. throw an error - breaking-ish change though?
- C. (my least favorite) just show a console warning

mweststrate:
B sounds the most clear to me

Op vr 31 aug. 2018 22:31 schreef Javier Gonzalez <notifications@github.com>:

xaviergonz:
@mweststrate upon further investigation, it seems like the current behavior is A, just that it failed sometimes when the instance was created from a function rather than passed as a direct parameter

I fixed what made it fail in https://github.com/mobxjs/mobx-state-tree/pull/1003
I added there some thoughts there as well, with a question on how to proceed

xaviergonz:
Released in v3.3.0

PR #1003, opened 2018-09-01
throw on instance for types.optional, fix for typechecking of instances generated by function

xaviergonz:
- Throws if an instance is passed as default value for types.optional (not if a function is passed and an instance is generated there though) - implements https://github.com/mobxjs/mobx-state-tree/issues/1002 (read the last comment on thoughts about this though)

- Improves typescript types.optional typing
- Fixed an issue where when a function was used to create an instance inside types.optional it would typecheck the instance (and in some cases fail) rather than the instance snapshot
- Reconciliation algorithm for default values will use the snapshot rather than the instance

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # 3.2.5
 
 -   Fix for TS3 issues with reference type [#994](https://github.com/mobxjs/mobx-state-tree/issues/994) through [#995](https://github.com/mobxjs/mobx-state-tree/pull/995) by [@xaviergonz](https://github.com/xaviergonz)
+-   types.optional will now throw if an instance is passed as default value [#1002](https://github.com/mobxjs/mobx-state-tree/issues/1002), fixes related to functions that generate instances inside types.optional incorrectly throwing through [#1003](https://github.com/mobxjs/mobx-state-tree/pull/1003) by [@xaviergonz](https://github.com/xaviergonz)
 
 # 3.2.4
 
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,7 @@
 # 3.2.5
 
 -   Fix for TS3 issues with reference type [#994](https://github.com/mobxjs/mobx-state-tree/issues/994) through [#995](https://github.com/mobxjs/mobx-state-tree/pull/995) by [@xaviergonz](https://github.com/xaviergonz)
--   types.optional will now throw if an instance is passed as default value [#1002](https://github.com/mobxjs/mobx-state-tree/issues/1002), fixes related to functions that generate instances inside types.optional incorrectly throwing through [#1003](https://github.com/mobxjs/mobx-state-tree/pull/1003) by [@xaviergonz](https://github.com/xaviergonz)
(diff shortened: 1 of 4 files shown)


xaviergonz:
Btw, the fix is for
```
    const Factory = types.model({
        rows: types.optional(types.array(Row), () => types.array(Row).create())
    })
```
since the two arrays are of different "type" (calling types.array two different times) it would be ok as a direct parameter but break when generated by a function, since for direct parameters it would check the snapshot was of the correct type, but for the function it would check if the instance is of the correct type.

In other words, before:
```
// ok, since the snapshot of the instance matches the type
rows: types.optional(types.array(Row), types.array(Row).create())
// throws since the instance does not match the type (typecheck fails)
rows: types.optional(types.array(Row), () => types.array(Row).create())
```
after this patch:
```
// throws since an instance is passed instead of an snapshot
rows: types.optional(types.array(Row), types.array(Row).create())
// ok, since the instance snapshot matches the type
rows: types.optional(types.array(Row), () => types.array(Row).create())
```

I was thinking that since the old behavior was to snapshot-itize the instance even for direct parameters (which I actually thought that was broken) and actually what was broken was the function parameter, maybe I should change the PR to accept direct instances once more not to create a breaking change? (although I'd say passing an instance as a default value doesn't really make much sense)

Thoughts?

mweststrate:
@xaviergonz 

`A`, and I think it should be possible to directly return an instance, without having it snapshotted; the reason is that you might be using an environment in that initializer function, or relying on some hooks or volatile state which would otherwise run twice. 

So I think 
```
// throws, instance is directly type checked and does not match the equivalent type
rows: types.optional(types.array(Row), () => types.array(Row).create())
```

Is the correct behavior, (maybe we should see if the error is clear enough or clarify that the types are not pointer equal, or document this as pitfal etc). There are many places in MST where using structural similary, but not literally the same types has this very same consequence (and this is the same as classes in almost any type system). 

You will run into the same issue with references etc. It is clear why this can be confusing in cases, but I think clear constraints are in the end easier to grasp as making some exceptions where we try to work around those constraints for the user

mweststrate:
Can you add a test for the above case as well btw? I don't think we have a test for that now

mweststrate:
Further looking good, thanks a lot!

xaviergonz:
@mweststrate I made some changes based on what you said, hopefully that's what you meant
(throw if instance is passed without function, allow instance if passed with function, use instance rather than snapshot whenever possible for reconcile/instantiation)

Another change I did is that typechecking for generator functions are only done once the generator function is actually called
This is to have the same behaviour both on production and non production envs, this is:

before:
- production env would only call the generator function once needed to create the node
- non-prod env would call it when types.optional is used and then again when needed to create the node

after the change:
- both behave like in the prod environment

mweststrate:
@xaviergonz yes, looking good! Nice catch on the type checking of the generator function btw!

xaviergonz:
@mweststrate is it officially approved then? also should I cut a new version with both fixes?

mweststrate:
Yes, approved! Feel free to cut a release, or if you lack the time I can do
it later today :-) I think this can best be a minor one, since we slightly
changed the behavior of optional?

Op do 6 sep. 2018 om 15:48 schreef Javier Gonzalez <notifications@github.com

Issue #906, opened 2018-06-29
Make array and map types optional by default


mweststrate:

mweststrate:
array and map types will be optional by default when used _directly_ as model properties. 

Tried making child objects optional as well, but this became very confusing in combination with `maybe` and such (since `undefined` would then be an acceptable value, leaving a field in a snapshot would be inferred to `{}` instead of `null`, etc..

mweststrate:
Scheduled for 3.0, closing

jlgrall:
How to make them mandatory when I need it ?

Seems to me the only way would be adding a `refinement()` over the current implicit `optional()` wrapping:
```javascript
var MyList = types.model({
	list: types.refinement(types.array, v => v !== undefined);
});
```

Would this refined type still return `true` with a call to `isOptionalType(value)` ?

PR #1004, opened 2018-09-02
add docs re: optional map / array types of #906

agilgur5:
- a8c7c60 implemented this but did not change the docs

My literal process without first knowing these were optional:
When I was writing a few models, I first wrote `types.array` without `types.optional`, then came back like "oh sh*t, don't I need to wrap these in `types.optional`?". Looked at the docs, didn't see anything about them being optional by default, and then re-wrote them all as `types.optional(types.array(...), [])`. Then I did a search and found the answer in #906 and the changelog 😅

Figure these should be documented and explicit so someone in the future can save themselves that triple-checking haha.

P.S. I finally got around to using MST after hyping it up on Twitter for a while :) Some of the advanced usage in the docs took a bit to wrap my head around, but was very intuitive / straightforward once I just started writing code, with gotchas like `flow` documented well. This was one of the only things not explicitly answered in the docs (another might be usage of `getSnapshot` vs. `toJS`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -892,6 +892,8 @@ These are the types available in MST. All types can be found in the `types` name
 -   `types.array(type)` Declares an array of the specified type.
 -   `types.map(type)` Declares a map of the specified type.
 
+When used in a model type as property type, `types.array` and `types.map` will be wrapped in `types.optional` by default, with `[]` and `{}` set as their default values, respectively.
+
 ## Primitive types
 
 -   `types.string`
--- a/README.md
+++ b/README.md
@@ -892,7 +892,7 @@ These are the types available in MST. All types can be found in the `types` name
 -   `types.array(type)` Declares an array of the specified type.
 -   `types.map(type)` Declares a map of the specified type.
 
-When used in a model type as property type, `types.array` and `types.map` will be wrapped in `types.optional` by default, with `[]` and `{}` set as their default values, respectively.
+Note that since MST v3 `types.array` and `types.map` are wrapped in `types.optional` by default, with `[]` and `{}` set as their default values, respectively.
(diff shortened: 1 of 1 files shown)


xaviergonz:
Thanks! care to add "since version 3" somewhere not to confuse users of older versions?

xaviergonz:
Edited it myself, and merged. Thanks for the contributionª

agilgur5:
Cool thanks @xaviergonz ! 

You removed the part about being a model property type though, which unless that's changed in a recent version (I would think that would be breaking), was an important detail. It meant that if `types.array` wasn't a direct property of a model, i.e. if it were wrapped in something else, it would not be defaulted to optional. Is that no longer the case?

xaviergonz:
Yes, I removed it because actually you can still do types.array(TModel).create() outside a model property and it will work even when no parameter to create is provided (because it is optional in this case as well)

xaviergonz:
see for example this:
https://codesandbox.io/s/86o9r079

agilgur5:
I see. So it's only defaulted then in when _directly_ `.create`'d or when the _direct_ property of a model.
So `types.array(types.array(types.number))` in both scenarios will be `[]` and not `[[]]`.

And
```javascript
types.union(types.array(types.number), types.number).create()
```

will error with an empty argument, but
```javascript
types.union(types.optional(types.array(types.number), []), types.number).create()
```

will not error.

I think that's intuitive for the most common use cases, and for more complex cases you tend to want to be very explicit and therefore seems intuitive as well. There are some gotchas to wrap your head around though, makes me wonder whether making them optional by default created more of those than it solved.

PR #1010, opened 2018-09-08
Cleanup deps, use jest for mst-middlewares, remove sinon

xaviergonz:
Basically chores:
- Updates deps
- Uses jest for mst-middlewares and removes ava and sinon
- Moves some packages to the root (tslint, typescript, etc)
- Removes some unused dev-deps
- Removes trailing comma rule from linter (covered by prettier and made tslint on visual studio crash, loosing linter)
- Adds linting to mst-middlewares
- mst-middlewares tests now using more strict ts
- Probably some other chores I forgot


xaviergonz:
merging since it is just chores without any effect on actual functionality

PR #1011, opened 2018-09-10
A way to get strong typing of self referencing models

xaviergonz:
```ts
    const BinaryTreeNode = types.model(selfType => ({
        data: types.string,
        leftNode: types.maybe(selfType),
        rightNode: types.maybe(selfType)
    }))
```
:D

There are some constraints though:
- the type will only be ok up to 10 levels deep (can be modified though)
- only maybe/maybeNull/array/map can be used with it for now (well, technically more constructs can, but TS will break)

still looking on a way to make double refs work... but are they even a good idea in trees?

PS: although unrelated I found a way to make late work again without a temp var
```ts
types.late((): any => X) // any placed there is the key
```


xaviergonz:
Ok, i just found it doesn't show children self-typed actions, working on it...

xaviergonz:
ok, disregard it, i could make it work for properties, but the "child" properties did not have actions, views, etc (unless you put props after those, but then actions wouldn't be able to use such self-props unless typecasted)

I tried another approach, but TS doesn't support circular refs in types :(

mweststrate:
Darn, just hoped you worked a miracle 😊. All static type systems are
limited in the end I guess

Op wo 12 sep. 2018 20:06 schreef Javier Gonzalez <notifications@github.com>:

xaviergonz:
Yeah, it is a shame

However I found that if you do something like
```
const Node = types.model({
  x: 5,
  me: types.maybe(types.late((): IAnyModelType => Node))
})
```
then at least you got strong typings for everything for the "me" property, but you can just typecast it whenever it needs to be used. E.g:
```
const node = Node.create({...})
(me as Instance<typeof Node>).x = 8
```
I'm making a PR about removing any as much as possible from the unit tests (plus tiding up a little bit the model typing)

RainerAtSpirit:
Here's an approach that combines the factory pattern as discussed with @xaviergonz in https://github.com/mobxjs/mobx-state-tree/issues/1029 with a manual recursive interface.
What do you think?

```typescript
import { IAnyModelType, Instance, types } from "mobx-state-tree";

export const TNullOrOptionalString = types.maybeNull(
  types.optional(types.string, "")
);

const SomeModel = types.model({
  id: TNullOrOptionalString,
  name: TNullOrOptionalString
});

interface ISomeModel extends Instance<typeof SomeModel> {}

// Recursive interface: https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#3.11.8
interface IRecursiveSomeModel extends ISomeModel {
  me: IRecursiveSomeModel;
}

// model factory
interface ICreateRecursiveModel<IT> {
  modelName: string;
  modelType: IT;
}

const createRecursiveModel = <IT extends IAnyModelType>({
  modelName,
  modelType
}: ICreateRecursiveModel<IT>) => {
  const someStore = types.compose(
    modelName,
    modelType,
    types.model({
      me: types.maybe(types.late((): IAnyModelType => modelType))
    })
  );

  return someStore;
};


const RecursiveSomeModel = createRecursiveModel({
  modelName: "SomeModel",
  modelType: SomeModel
});

const store: IRecursiveSomeModel = RecursiveSomeModel.create();
store.me.me.me.me.me.id; // ok. id string | null



```

PR #1012, opened 2018-09-12
simplified model typings and removed any in lots of tests

xaviergonz:
also added a readme entry about how to minimize the results of using typescript with self-referencing types by using late((): IAnyModelType => ...)


mweststrate:
Great stuff! No further comments :)

xaviergonz:
thanks! merging

PR #1013, opened 2018-09-13
{pre,post}ProcessSnapshot on compose

### LINKED ISSUE #613: preProcessSnapshot not copied on compose
Sometimes I encode a single *behaviour* of my objects into a helper function. This helper function then makes it easier to make type definitions more DRY and composable. Below is an example for adding an `id` identifier strategy.

Right now we have to wrap the type definition call. It would be nice to have a `.mixin` method, to allow for mixin in these behaviour generating functions.

```javascript
  // adds an automatic generated unique id to the object
  // and adds `shortId` property for displaying in the ui
  const withIdProperty = type =>
    type
      .props({
        // add id property with random default
        id: t.optional(
          t.identifier(t.string),
          () => `${type.name}-${generateUUID()}`
        ),
      })
      .views(self => ({
        get shortId() {
          // use this in the app
          return self.id.split('-')[1];
        },
      }));

  // current way
  const Record = withIdProperty(
    t.model('Record', {
      name: t.string,
    })
  );

  // proposed way
  const Record = t
    .model('Record', {
      name: t.string,
    })
    .mixin(withIdProperty);
```

Where `.mixin` could be implemented like:
```javascript
type.mixin = function(behaviour) {
  return behaviour(this);
}
```

theRealScoobaSteve:
Closes #613

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/type-system.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/type-system.test.ts
@@ -386,6 +386,37 @@ test("it should type compose correctly", () => {
     x.drive()
     x.log("z")
 })
+test("it should extend {pre,post}ProcessSnapshot on compose", () => {
+    const CompositionTracker = types
+        .model({
+            composedOf: types.array(types.string)
+        })
+        .preProcessSnapshot(snapshot => ({
+            ...snapshot,
+            composedOf: (snapshot.composedOf || []).concat("CompositionTracker")
+        }))
+    const Car = types.model({}).preProcessSnapshot(snapshot => ({
+        ...snapshot,
+        composedOf: (snapshot.composedOf || []).concat("Car")
+    }))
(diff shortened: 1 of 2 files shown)


packages/mobx-state-tree/__tests__/core/type-system.test.ts
```diff
@@ -386,6 +386,37 @@ test("it should type compose correctly", () => {
...
+        .compose(
+            CompositionTracker,
+            Car,
+            Logger
+        )
+        .props({
+            composedOf: types.array(types.string)
+        })
+    const x = LoggableCar.create({})
+    expect(x.composedOf).toContain("CompositionTracker")
```
xaviergonz: maybe also check the order by using a single toEqual([....])? or the array indexes and toBe

mweststrate:
Agreed, this can be merged. Thanks for finding & fixing @theRealScoobaSteve!

xaviergonz:
merging!

PR #1014, opened 2018-09-13
fix for this failing to resolve on actions

xaviergonz:
Under some circumstances this failed to resolve as self on actions.
This PR fixes that, making actions that depend on other actions easier to write

example
```ts
actions(self => ({
  a(...) { ... },
  b(...) { this.a(...) }
}))
```

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/this.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/this.test.ts
@@ -1,0 +1,85 @@
+import { types } from "../../src"
+import { isObservableProp, isComputedProp } from "mobx"
+
+test("this support", () => {
+    const M = types
+        .model({ x: 5 })
+        .views(self => ({
+            get x2() {
+                return self.x * 2
+            },
+            get x4() {
+                return this.x2 * 2
+            },
+            boundTo() {
+                return this
+            },
+            innerBoundTo() {
+                return () => this
+            },
+            isThisObservable() {
+                return (
+                    isObservableProp(this, "x2") &&
+                    isObservableProp(this, "x4") &&
+                    isObservableProp(this, "localState") &&
+                    isComputedProp(this, "x2")
+                )
+            }
+        }))
+        .volatile(self => ({
+            localState: 3,
+            getLocalState() {
+                return this.localState
+            },
+            getLocalState2() {
+                return this.getLocalState() * 2
+            }
+        }))
+
+        .actions(self => {
+            return {
+                xBy(by: number) {
+                    return self.x * by
+                },
+                setX(x: number) {
+                    self.x = x
+                },
+                setXBy(x: number) {
+                    this.setX(this.xBy(x))
+                },
+                setLocalState(x: number) {
+                    self.localState = x
+                }
+            }
+        })
+
+    const mi = M.create()
+
+    expect(mi.isThisObservable()).toBe(true)
+
+    expect(mi.boundTo()).toBe(mi)
+    expect(mi.innerBoundTo()()).toBe(mi)
+
+    expect(mi.x).toBe(5)
+
+    mi.setX(6)
+    expect(mi.x).toBe(6)
+
+    mi.setXBy(2)
+    expect(mi.x).toBe(12)
+    expect(mi.x2).toBe(12 * 2)
+    expect(mi.x4).toBe(12 * 4)
+    expect(mi.xBy(2)).toBe(24)
+
+    expect(mi.localState).toBe(3)
+    expect(mi.getLocalState()).toBe(3)
+    expect(mi.getLocalState2()).toBe(3 * 2)
+
+    mi.setLocalState(6)
+    expect(mi.localState).toBe(6)
+    expect(mi.getLocalState()).toBe(6)
+    expect(mi.getLocalState2()).toBe(6 * 2)
+
+    mi.setLocalState(7)
+    expect(mi.localState).toBe(7)
+})
--- a/packages/mobx-state-tree/__tests__/core/this.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/this.test.ts
@@ -44,6 +44,9 @@ test("this support", () => {
                 setX(x: number) {
                     self.x = x
                 },
+                setThisX(x: number) {
+                    ;(this as any).x = x // this should not affect self.x
+                },
                 setXBy(x: number) {
                     this.setX(this.xBy(x))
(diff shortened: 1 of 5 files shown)


packages/mobx-state-tree/__tests__/core/this.test.ts
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }))
+
+        .actions(self => {
+            return {
+                xBy(by: number) {
+                    return self.x * by
+                },
+                setX(x: number) {
+                    self.x = x
```
mweststrate: Just to double check, `this.x` won't work, correct? `this` revers to the definition and not to the instance? I would make sure to mention that at the place in the readme where you introduce the `this` workaround in the Readme, otherwise people will assume it refers to the same thing as `self`

Really smart solution btw!

packages/mobx-state-tree/src/types/complex-types/model.ts
```diff
@@ -325,8 +325,10 @@ export class ModelType<S extends ModelProperties, T> extends ComplexType<any, an
                     `Cannot define action '${POST_PROCESS_SNAPSHOT}', it should be defined using 'type.postProcessSnapshot(fn)' instead`
                 )
 
+            let action2 = actions[name].bind(self) // bind to self to allow 'this' to always work
```
mweststrate: shouldn't this be `actions[name].bind(actions)`?

packages/mobx-state-tree/__tests__/core/this.test.ts
xaviergonz: actually I think it would work, but TS would complain
I can add a unit test for it

of course that depends on the wanted behaviour (which depends on the other line of code you commented on)

packages/mobx-state-tree/src/types/complex-types/model.ts
xaviergonz: depends on what you want. if you want it to be able to reference other pieces of code beyond the current scope then it should be "this" (although TS will complain though)
if you **don't** want it to work then yes, it should be "this"

which behaviour is the wanted one?

mweststrate:
I think we want to encourage people to use `self`, and use `this` *only* to
access other actions in the same definition block, which would be
consistent with the `volatile` and `views` (and `extend`!) as well if I am
not mistaken

Op vr 14 sep. 2018 om 16:42 schreef Javier Gonzalez <
notifications@github.com>:

xaviergonz:
sounds good, ill change it to bind to the passed object instead of self then

xaviergonz:
PS: actually volatile, views and extends could also be modified so this works as self if wanted :)

mweststrate:
You might want to check the action creation code; I think we bind actions
explicitly to `undefined` somewhere in there, if we don't do that, we might
also not need this bind :).

Op vr 14 sep. 2018 om 16:50 schreef Javier Gonzalez <
notifications@github.com>:

xaviergonz:
@mweststrate apparently binding to the object makes the middleware loose interdependent actions
this is, given
```
actions(self => ({
  a(...) { ... },
  b(...) { this.a(...) }
}))
```
if b is invoked then only b will be recorded, but not a

xaviergonz:
ok, found a way to make it work, PR updated

mweststrate:
@xaviergonz ok, looks good! Would you mind checking what this means for the performance suite? The additional binding might instance creation slower?

xaviergonz:
@mweststrate looks about the same to me

before fix:
```
-----------
Small Model
-----------
      17ms |      x 100 |  0.2ms avg
     138ms |    x 1,000 |  0.1ms avg
     951ms |   x 10,000 |  0.1ms avg
     114ms |    x 1,000 |  0.1ms avg
      18ms |      x 100 |  0.2ms avg

------------
Medium Model
------------
      26ms |      x 100 |  0.3ms avg
     242ms |    x 1,000 |  0.2ms avg
   1,654ms |   x 10,000 |  0.2ms avg
     220ms |    x 1,000 |  0.2ms avg
      27ms |      x 100 |  0.3ms avg

------------------------
Large Model - 0 children
------------------------
      86ms |      x 100 |  0.9ms avg
     555ms |    x 1,000 |  0.6ms avg
      75ms |      x 100 |  0.8ms avg

-------------------------------------------
Large Model - 10 small & 10 medium children
-------------------------------------------
      95ms |       x 50 |  1.9ms avg
     333ms |      x 250 |  1.3ms avg
      80ms |       x 50 |  1.6ms avg

-------------------------------------------
Large Model - 100 small & 0 medium children
-------------------------------------------
      94ms |       x 50 |  1.9ms avg
     328ms |      x 250 |  1.3ms avg
      88ms |       x 50 |  1.8ms avg

-------------------------------------------
Large Model - 0 small & 100 medium children
-------------------------------------------
     297ms |       x 50 |  5.9ms avg
   1,405ms |      x 250 |  5.6ms avg
     327ms |       x 50 |  6.5ms avg
```

after fix:
```
-----------
Small Model
-----------
      14ms |      x 100 |  0.1ms avg
     150ms |    x 1,000 |  0.1ms avg
     929ms |   x 10,000 |  0.1ms avg
     118ms |    x 1,000 |  0.1ms avg
      15ms |      x 100 |  0.1ms avg

------------
Medium Model
------------
      24ms |      x 100 |  0.2ms avg
     255ms |    x 1,000 |  0.3ms avg
   1,733ms |   x 10,000 |  0.2ms avg
     225ms |    x 1,000 |  0.2ms avg
      29ms |      x 100 |  0.3ms avg

------------------------
Large Model - 0 children
------------------------
      92ms |      x 100 |  0.9ms avg
     491ms |    x 1,000 |  0.5ms avg
      77ms |      x 100 |  0.8ms avg

-------------------------------------------
Large Model - 10 small & 10 medium children
-------------------------------------------
      96ms |       x 50 |  1.9ms avg
     363ms |      x 250 |  1.5ms avg
     107ms |       x 50 |  2.1ms avg

-------------------------------------------
Large Model - 100 small & 0 medium children
-------------------------------------------
     111ms |       x 50 |  2.2ms avg
     311ms |      x 250 |  1.2ms avg
      87ms |       x 50 |  1.7ms avg

-------------------------------------------
Large Model - 0 small & 100 medium children
-------------------------------------------
     319ms |       x 50 |  6.4ms avg
   1,312ms |      x 250 |  5.2ms avg
     380ms |       x 50 |  7.6ms avg
```

PR #1017, opened 2018-09-15
Better flow typing for ts3

xaviergonz:
Better flow typings, but require Typescript 3

Arguments are now properly inferred, and return types are chosen like this:

```
// this gives as definition: () => Promise<5>
flow(function*() {
  yield somePromise // this adds a Promise<any> return type, which will be ignored
  yield somePromise2 // this adds a Promise<any> return type, which will be ignored
  return 5; // this is the actual type that will be returned
}
```

```
// this gives as definition: () => Promise<void>
flow(function*() {
  // why?
  yield somePromise // this adds a Promise<any> return type, which will be ignored
  yield somePromise2 // this adds a Promise<any> return type, which will be ignored
  // since nothing is returned, it is transformed into void
}
```

A couple of caveats still exist though:
- if a Promise is actually returned as final value for the flow it will be inferred as return type void
- const a = yield somePromiseThatReturnsANumber // a will be any, know TS issue, see https://github.com/Microsoft/TypeScript/issues/2983
  - maybe could be fixed by using async function generators rather than just plain generators? (see https://jakearchibald.com/2017/async-iterators-and-generators/ - Async generators: Creating your own async iterator)

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/async.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/async.test.ts
@@ -12,7 +12,7 @@ import {
 } from "../../src"
 import { reaction, configure } from "mobx"
 
-function delay(time: number, value: any, shouldThrow = false) {
+function delay<TV>(time: number, value: TV, shouldThrow = false): Promise<TV> {
     return new Promise((resolve, reject) => {
         setTimeout(() => {
             if (shouldThrow) reject(value)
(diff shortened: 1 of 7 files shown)


packages/mobx-state-tree/__tests__/core/async.test.ts
```diff
@@ -153,7 +155,7 @@ test("can handle throw from yielded promise works", t => {
 test("typings", done => {
     const M = types.model({ title: types.string }).actions(self => {
         function* a(x: string) {
-            yield delay(10, "x", false)
+            const aa = yield delay(10, "x", false)
```
hamczu: "aa" not used anywhere later?

xaviergonz: good catch

igimanaloto:
This would be really nice to have!

PR #1023, opened 2018-09-25
Fixed union type in README

hamczu:
When I tried to use the `union` type with dispatcher I found that there's a mismatch between the docs (README) and implementation. I tried to update it based on how I understand that piece of code - any feedback is welcome.


xaviergonz:
That's right the docs were right for v2 but wrong for v3, thanks!

PR #1026, opened 2018-09-27
Fix after hooks inside computed

### LINKED ISSUE #967: afterCreate is firing when a computed accesses a child causing a side effect
FYI: I've posted this on spectrum.  It looks like a possible bug. 

Here's a small (50 ln) codesandbox example showing what I'm up against:
https://codesandbox.io/s/94zy6n96jo

My issue is with a tree where a child has an afterCreate action that modifies another child.  If I access the child data using a computed I get the following error: 
_Error: [mobx] Computed values are not allowed to cause side effects by changing observables that are already being observed. Tried to modify: ObservableObject@69.selected_

I assumed afterCreate would fire after the object had been created?  Can anyone shed some light on when afterCreate and afterAttach really fire?
Thanks!!

xaviergonz:
Closes #967 

Requires mobx 4.5.0 || 5.5.0

Also included:
- updated dependencies
- fixed examples


xaviergonz:
@mweststrate does the requirement of mobx 5.5 / 4.5 make this a breaking change?

xaviergonz:
Avoid making it a breaking change by checking if the peer mobx version exposes the required functionality. Either way when installing the version and the wron mobx peer dep is installed yarn will issue a warn like mobx-state-tree@3.5.0" has unmet peer dependency "mobx@>=4.5.0 <5.0.0 || >=5.5.0 <6.0.0"., but will work as it used to.

xaviergonz:
merging since the fix itself is pretty simple:
```ts
fireHook(name: string) {
        const fn =
            this.storedValue && typeof this.storedValue === "object" && this.storedValue[name]
        if (typeof fn === "function") {
            // we check for it to allow old mobx peer dependencies that don't have the method to work (even when still bugged)
            if (_allowStateChangesInsideComputed) {
                _allowStateChangesInsideComputed(() => {
                    fn.apply(this.storedValue)
                })
            } else {
                fn.apply(this.storedValue)
            }
        }
    }
```

the rest is just fixing the unmaintained examples and updating dev deps

PR #1027, opened 2018-09-27
fix for cast for ts3.1.1

xaviergonz:
Sadly Typescript 3.1.1 broke the cast function

this PR fixes it, but cast looses some capabilities :(
- cast can no longer cast from primitive to primitive (not like it was too useful anyway though). Although that can be fixed then it will break the ability to cast properly for custom types.
- cast will now compile outside assignations, however it will yield an empty object, so not much can be done with it anyway
- cast can no longer cast from reference ids to references, as any has to be used in this case now

The fix is compatible all the way down to Typescript 2.9.2


xaviergonz:
disregard this pr, found a better fix

PR #1028, opened 2018-09-27
better fix for cast in ts 3.1.1

xaviergonz:
Fixes cast being broken in Typescript 3.1.1

the only downside is that cast outside assignations will compile, yet it will yield an unusable "never" type

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/type-system.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/type-system.test.ts
@@ -768,14 +768,13 @@ test("cast and SnapshotOrInstance", () => {
     unprotect(map)
     map.set("a", cast({ n2: 5 })) // not really needed in this case, but whatever :)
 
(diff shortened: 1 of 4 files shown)


xaviergonz:
merging since it is a pretty small fix

packages/mobx-state-tree/__tests__/core/type-system.test.ts
```diff
@@ -768,14 +768,13 @@ test("cast and SnapshotOrInstance", () => {
...
 
-    // and the best part, it actually doesn't work outside assignments :DDDD
-    // all this fails to compile
-    // cast([])
-    // cast({a:5})
-    // cast(NumberArray.create([]))
-    // cast(A.create({n2: 5}))
-    // cast({a: 2, b: 5})
-    // cast(NumberMap({a: 2, b: 3}))
+    // although this compiles, it yields a never type
```
lookfirst: Did you want to test that this yields a never type? In case that ever changes, at least the test will break.

PR #1031, opened 2018-10-03
update compose to check for model type resolves #1021

### LINKED ISSUE #1021: compose should throw an exception if any of the types is not a model type
After update from 3.3.0 to 3.4.0 have an error when creating store instance.

```TypeError: cur.applySnapshotPreProcessor is not a function```
image

danecando:


xaviergonz:
Thanks!

PR #1032, opened 2018-10-04
[Fixed #750] Flow actions mutations not syncing to Redux DevTools.

### LINKED ISSUE #750: Flow actions mutations not syncing to Redux DevTools
Not sure if bug, or again a question (and if so, sorry for posting them all here, I tried the spectrum.chat room, but code pasting in there is a huge pain)

I have hooked up MST with Redux DevTools like so:

```js
const store = RootStore.create()
connectReduxDevtools(remotedev, store)
```

And I have some action that mutate one of the node like so:

```js
const SessionModel = types
  .model({
    token: types.maybe(types.string)
  })
  .actions(self => ({
    bootstrap: flow(function* bootstrap() {
      const sessionRaw = yield AsyncStorage.getItem(SessionStorageKey)
      if (sessionRaw) {
        const session = JSON.parse(sessionRaw)
        applySnapshot(self, session)
      }
    }),
    setToken: flow(function* setToken(token: string) {
      self.token = token
      yield AsyncStorage.setItem(
        SessionStorageKey,
        JSON.stringify(getSnapshot(self))
      )
    }),
    logout: flow(function* logout() {
      yield AsyncStorage.removeItem(SessionStorageKey)
      applySnapshot(self, {})
    })
  }))
```

When I trigger these actions, I notice no updates to the state tree in RDT. I see the actions firing, but the state does not update along with the actions.

Did I set this up wrong?

bourquep:
By only hooking into the `onAction` notification, the Redux middleware was handling flow (async) actions at their beginning, where the state has not changed yet. So this MR injects another middleware to intercept `flow_return` action types, which is when the async action is done and the new state is available. 

Ideally, we would exclude the regular `action` type for these, but at the time the middleware is invoked with `call.type = "action"` there is no way to determine if this is in fact an async action.

So, when an async action is invoked, the Redux DevTools will display:

```
myAction
myAction [flow_return]
```

And it's the `[flow_return]` one that will include the new state.

There weren't any existing tests for the Redux middleware, and I wouldn't really know how to implement them (looks like it would require mocking `remotedev`).

### DIFF
--- a/packages/mst-middlewares/src/redux.ts
+++ b/packages/mst-middlewares/src/redux.ts
@@ -79,6 +79,28 @@ export function connectReduxDevtools(remoteDevDep: any, model: any) {
         true
     )
 
+    // Fix for issue #750
+    //
+    // By only hooking into the 'onAction' notification, the Redux middleware
+    // was handling flow (async) actions at their beginning, where the state
+    // has not changed yet. So we define another middleware to intercept
+    // "flow_return" action types, which is when the async action is done and
+    // the new state is available. Ideally, we would exclude the regular "action"
+    // type for these, but at the time the middleware is invoked with
+    // `call.type = "action"` there is no way to determine if this is in fact
+    // an async action.
+    mst.addMiddleware(model, actionMiddleware)
+    function actionMiddleware(call: mst.IMiddlewareEvent, next: any) {
+        if (!applyingSnapshot && call.type === "flow_return") {
+            const copy: any = {}
+            copy.type = `${call.name} [${call.type}]`
+            if (call.args) call.args.forEach((value, index) => (copy[index] = value))
+            remotedev.send(copy, mst.getSnapshot(model))
+        }
+
+        next(call)
+    }
+
     function handleMonitorActions(remotedev2: any, model2: any, message: any) {
         switch (message.payload.type) {
             case "RESET":


xaviergonz:
makes me wonder, would it make sense for all the flow types to be shown on the redux dev tools?
the point being, what if a flow_resume of action A changes property (PA), then flow_resume of action B changes property PB, and then flow_return of B changes PB2, would the devtools assume PA, PB and PB2 changed because of action B because the resume ones were not tracked?

xaviergonz:
also while at it I'd refactor the "onAction" code to be inside the new middleware so there's less duplicated code

packages/mst-middlewares/src/redux.ts
```diff
@@ -79,6 +79,28 @@ export function connectReduxDevtools(remoteDevDep: any, model: any) {
...
+    // was handling flow (async) actions at their beginning, where the state
+    // has not changed yet. So we define another middleware to intercept
+    // "flow_return" action types, which is when the async action is done and
+    // the new state is available. Ideally, we would exclude the regular "action"
+    // type for these, but at the time the middleware is invoked with
+    // `call.type = "action"` there is no way to determine if this is in fact
+    // an async action.
+    mst.addMiddleware(model, actionMiddleware)
+    function actionMiddleware(call: mst.IMiddlewareEvent, next: any) {
+        if (!applyingSnapshot && call.type === "flow_return") {
```
xaviergonz: I'd probably remove  `&& call.type === "flow_return"` and the onAction code

xaviergonz:
or maybe it could detect if there was no change between the last snapshot and the new one and if that's the case then avoid showing the event (though that might be kind of slow for slow states).

Maybe keeping a counter of json patch changes and when an action comes
```
  if (deltaPatches <= 0) skip logging
  else log and deltaPatches = 0;
```
?

bourquep:
I have tried various things, but each attempt results in Chrome running out of memory. Not sure if it's an issue with devtools or with mst-middleware, but I don't have time right now to look into it.

bourquep:
Closing this PR in favor of #1035

PR #1043, opened 2018-10-12
Types optimization

### LINKED ISSUE #973: 3.x tsc compile performance with declarations & deeply nested stores with arrays
I am upgrading a typescript mst project from 2.0.1 to 3.x.  The project is a library referenced by other packages.  My model has a root store with nested stores, where some of the nested stores contain arrays of those stores.  

With 3.x, the compile time in my actual project is significant (over 2 min) and produces massive ts.d files. 

I have a very scaled down repro here:  https://github.com/garygentry/mst-decl-test

On this sample, compiling with declarations takes about 26 seconds, without declarations its about 5 seconds.  The declaration for the root store is 38,000 loc!

Clearly this has something to do with the typescript magic to represent the model, and I get this may be stretching things, but this exact model was working fine in 2.0.1.

I tried 3.2.1 and 3.1.1 with similar performance.

Thoughts?

xaviergonz:
Following https://github.com/mobxjs/mobx-state-tree/issues/973#issuecomment-428773023 I embarked into the task to see if there was any way to optimize declaration types so they were smaller and took less time to generate.

I took the example from https://github.com/garygentry/mst-decl-test-2 and tried to compile it wit TS3.1.3, this were the results:
- declarations: false - compiled in around 4 seconds
- declarations: true - won't even compile on my machine due to out of RAM errors

After checking the generated typings I noticed that IModelType had 5generics (P, O, C, S, T), and that IModelInstance had 4 (P, O, C, S), but that they could really be reduced to P, O in both cases

The upside of that change is that for example things typing (https://github.com/garygentry/mst-decl-test-2/blob/master/dist/stores/storeTypes/patientProfile.d.ts) size changed as follows:
- TS2.8.2 - around 607 lines
- TS3.1.3 - around (gasp) 6k lines
- TS3.1.3 after this PR - 90 lines 💃 

And the compilation time for TS3.1.3 for the project with declarations enabled (that didn't even compile in my machine) changed to a more reasonable 14 secs

Another good side effect is that type completion should be faster

The downside of this change is that pre/post-process snapshot typings don't work properly and they need casting (since they modify the internal C and S types and those modifications are not tracked anymore in the generics), but I don't see it as a big deal since:
- they are usually used to transform from the same C to the same C and from the same S to the same S, which still works ok
- the typings were only OK as long as props wasn't used after they were used
- we still want to kill them right?

Should fix #973 

@mweststrate thoughts?


garygentry:
To verify, compile times on my machine:

ts 3.1.3: ~8.5s 
ts 2.8.4: 6.5s (was ~ 40 seconds)

Largest decl file is this one: 
https://github.com/garygentry/mst-decl-test-2/blob/master/dist/stores/teams/teams.d.ts

Was 10.4 MB, 3.3 MB with types-optimization-3.

xaviergonz:
Updated the PR description, pre/post process snapshots work ok again

xaviergonz:
PS: @garygentry I'd recommend to switch to TS3.1.3 once this comes out, since I noticed that TS2.8.4 generated any as types for some sub-models where it should not
TS 2.9.4 is also ok, but it takes double the compilation time

xaviergonz:
Also, wasn't teams.d.ts 19MB? that's what it shows on github at least

packages/mobx-state-tree/src/types/utility-types/reference.ts
```diff
@@ -201,7 +202,7 @@ export interface ReferenceOptions<T> {
     set(value: T, parent: IAnyStateTreeNode | null): string | number
 }
 
-export interface IReferenceType<IR extends IAnyModelType>
+export interface IReferenceType<IR extends IAnyComplexType>
```
xaviergonz: had to change this to complex type since now the union of models is a complex type (which I guess makes more sense since you can't use props, actions etc over a union of models)

packages/mobx-state-tree/src/types/utility-types/optional.ts
```diff
@@ -72,7 +72,9 @@ export class OptionalValue<C, S, T> extends Type<C, S, T> {
 
     getDefaultInstanceOrSnapshot() {
         const defaultInstanceOrSnapshot =
-            typeof this.defaultValue === "function" ? this.defaultValue() : this.defaultValue
+            typeof this.defaultValue === "function"
+                ? (this.defaultValue as IFunctionReturn<C | S | T>)()
```
xaviergonz: compilation fix when compiled with TS3 (not the case, but for the future :) )

packages/mobx-state-tree/src/types/complex-types/model.ts
```diff
@@ -117,65 +119,78 @@ export type ModelCreationType<T extends ModelProperties> = {
 } &
     { [K in keyof OptionalProps<T>]?: ExtractC<T[K]> }
 
+export type ModelCreationType2<T extends ModelProperties, FixedC> = FixedOrOther<
```
xaviergonz: At first I tried having the if inside ModelCreationType directly, but this ended up in types generation being messed up, therefore I had to split it. Either way the type generation will show/use ModelCreationType, not ModelCreationType2 (same for ModelSnapshotType)

packages/mobx-state-tree/src/types/complex-types/model.ts
```diff
@@ -117,65 +119,78 @@ export type ModelCreationType<T extends ModelProperties> = {
...
 }
 
-export interface IModelType<
-    PROPS extends ModelProperties,
-    OTHERS,
-    C = ModelCreationType<PROPS>,
-    S = ModelSnapshotType<PROPS>,
-    T = ModelInstanceType<PROPS, OTHERS, C, S>
-> extends IComplexType<C, S, T> {
+export interface IModelType<PROPS extends ModelProperties, OTHERS, FixedC, FixedS>
```
xaviergonz: FixedC/S are the overrides applied when using pre/postprocesssnapshot, they are false by default (rather than the whole blown type) to indicate that they are not overriden
I tried 'never' instead of false, but that wouldn't work well for some reason (maybe never never extends never? whatever)

xaviergonz:
Just confirmed that this makes intellisense with vscode is a lot snappier too 👍

garygentry:
> PS: @garygentry I'd recommend to switch to TS3.1.3 once this comes out, since I noticed that TS2.8.4 generated any as types for some sub-models where it should not
> TS 2.9.4 is also ok, but it takes double the compilation time

Yes, wanted to go to 3 and only went back to 2.84 b/c this issue.

> Also, wasn't teams.d.ts 19MB? that's what it shows on github at least

You're correct, mis-typed

Thanks, nice work!

xaviergonz:
Optimized array typings as well, teams.d.ts went down from 3.3mb to 2.2mb

xaviergonz:
And added more optimizations for array, map, optional, maybe, maybeNull etc.
Now teams.d.ts went down from the original 19MB to 40KB, and compilation time to 5 secs

xaviergonz:
Now using NotCustomized interface rather than false for not customized C, S (just in case a pre/postprocesssnapshot accepts false for some reason

xaviergonz:
I consider this pretty much done (can't think of anything else to improve it right now)

mweststrate:
Will try to review tomorrow, looks super promising! (sorry, still catching
up backlog after holiday :-|)

Op za 13 okt. 2018 om 20:09 schreef Javier Gonzalez <
notifications@github.com>:

packages/mobx-state-tree/src/types/complex-types/map.ts
```diff
@@ -42,43 +42,58 @@ import {
...
-    forEach(callbackfn: (value: T, key: string, map: IMSTMap<C, S, T>) => void, thisArg?: any): void
-    get(key: string): T | undefined
+    forEach(
+        callbackfn: (value: ExtractT<IT>, key: string, map: IMSTMap<IT>) => void,
+        thisArg?: any
+    ): void
+    get(key: string): ExtractT<IT> | undefined
     has(key: string): boolean
-    set(key: string, value: C | S | T): this
+    set(key: string, value: ExtractC<IT> | ExtractS<IT> | ExtractT<IT>): this
```
mweststrate: We could introduce `ExtractCST<IT>` to simplify?

packages/mobx-state-tree/src/types/utility-types/reference.ts
mweststrate: Should be fine I think

packages/mst-middlewares/src/action-logger.ts
```diff
@@ -1,12 +1,14 @@
-import { getPath, IMiddlewareEvent } from "mobx-state-tree"
+import * as mst from "mobx-state-tree"
```
mweststrate: I think `* as ` imports make tree shaking worse?

mweststrate: https://developers.google.com/web/fundamentals/performance/optimizing-javascript/tree-shaking/

packages/mobx-state-tree/src/types/complex-types/map.ts
xaviergonz: will do

packages/mst-middlewares/src/action-logger.ts
xaviergonz: just wanted to avoid the possibility of missing an import of an internal type when using TS <= 2.9 (although in theory that's not supported anymore), but guess I was just lazy :) will change

xaviergonz: ok, just checked it. it is kind of annoying, since for TS to pick up types that are being used internally you have to actually import them from the internal path

e.g. atomic uses (indirectly) IMiddlewareHandler, so I tried:
`
// not actually used but required for proper typing generation
import { IMiddlewareHandler } from "mobx-state-tree"
`
which still generates
`
declare const atomic: import("mobx-state-tree/dist/core/action").IMiddlewareHandler;
export default atomic;
`

so to fix it I have to mark the inferred types as actually `IMiddlewareHandler`, which for action middleware might be trivial, but for the redux one it is not.

by using the star import this is not needed and less error prone (and anyway in the old middlewares it was already importing all of mst since the redux middleware actually uses the star import.

so the question is, should I make the code uglier by adding all these manual type annotations to allow tree shaking or is the star import ok?

xaviergonz:
@mweststrate  ExtractCST done, but please read my comment about tree-shaking

packages/mst-middlewares/src/action-logger.ts
xaviergonz: Hmm, I take that back, both using star import and not using it end up in the same "relative" import.

I will do the tree shaking, though the question now is, should I:
- painfully import and mark inferred types as imported to support TS <= 2.8 (which in theory is not supported anymore)
- let TS generate the relative imports

PR #1046, opened 2018-10-15
fix for default value of eager option for union

xaviergonz:
I found a funny issue, when no options object is used for union the default for eager is true
when an empty option was used as options for union the default for eager was false

This PR aims to unify them into true (which is what the unit tests assume) and correct the readme accordingly

At first I thought of correct them into what the docs say (false), but most unit tests were failing (and I guess most people use union without arguments than with, so better to keep that one working)

Technically this is a breaking change, but should it be marked as such?


mweststrate:
I think in 98% of the cases the no arguments version is used, and default to `false` with options is a bug fix. Thanks for noticing!

xaviergonz:
Thanks! Updated changelog and merging

PR #1047, opened 2018-10-15
Fix for array reconciliation with ids of union types

### LINKED ISSUE #1045: Array item not updated when snapshot applied for certain union types
I have:

* [x] I think something is not working as it should.
  * [x] Describe expected behavior
  * [x] Describe observed behavior

**Sandbox:** https://codesandbox.io/s/43mv0yp0v4

**Expected:**
When snapshot for an array is applied - the new value is applied to an array item.

**Observed:**
Array item is not updated. 
screen shot 2018-10-15 at 15 05 18

Speculation - it has something to do with union type definition order in combination with maybe types, but I don't know MST well enough to be able to track this issue down.

xaviergonz:
Fixes #1045 

Basically when reconciliating unions of types with ids the first type to try should be the current node one


xaviergonz:
I'd like to add another unit test to see what happens when the snapshot is actually of the other type

xaviergonz:
just as I suspected it didn't work well if the type provided was of the second type yet had the same ids
now fixed as well

PR #1053, opened 2018-10-20
fix for references to objects being broken after replacing the node

### LINKED ISSUE #1052: Reference returns destroyed model after subtree replacing
**_Bug report_**

* https://codesandbox.io/s/lx003ww20m

If we have reference to the some part of subtree and replacing that subtree completely, reference still points to old, destroyed, model, so reference's read/write leads to livelyness error, even if new subtree contains correct new model with same type/identifier as in old model.

There is workaround - in action we can save id from reference in temporary variable, set reference to undefined and after replacing reassign identidier (`Safe reload` button in example). I think, it should be done by library, not by user. Of course, user still needs check if new subtree contains referenced identifier.

Reference should check if model was removed from current tree and return new model with same type/identifier or throw `Failed to resolve reference` if there is no that model in the tree.

xaviergonz:
Fixes #1052

### DIFF
--- a/packages/mobx-state-tree/src/types/utility-types/reference.ts
+++ b/packages/mobx-state-tree/src/types/utility-types/reference.ts
@@ -30,7 +30,9 @@ class StoredReference {
     }
 
     constructor(value: any, private readonly targetType: IAnyType) {
-        if (isStateTreeNode(value)) {
+        if (typeof value === "string" || typeof value === "number") {
+            this.identifier = value
+        } else if (isStateTreeNode(value)) {
             const targetNode = getStateTreeNode(value)
             if (!targetNode.identifierAttribute)
                 return fail(`Can only store references with a defined identifier attribute.`)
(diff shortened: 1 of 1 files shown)


Amareis:
Does you completely disable optimization with "object" mode in StoredReference?

xaviergonz:
it wasn't much of an optimization given there is an identifier cache underneath that keeps a pointer to the node

xaviergonz:
Also, at first I tried to keep it, but apparently making a mutation (in this case of the ref object) in the middle of a read operation (getValue) makes the system go nuts

xaviergonz:
Btw, the define a property when a property is set in another "this" is basically how JS does things normally internally.
From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain
```
Setting a property to an object creates an own property. The only exception to the getting and setting behavior rules is when there is an inherited property with a getter or a setter.
```

xaviergonz:
@k-g-a thanks for the review! all requested changes done :)

mweststrate:
Feel free to release :-)

mweststrate:
I think given the docs one would already assume this to work, so patch
should be ok?

Op di 23 okt. 2018 om 13:27 schreef Javier Gonzalez <
notifications@github.com>:

PR #1059, opened 2018-10-25
addDisposer enhancement

xaviergonz:
Superseeds #753 
Now that mobx-react itself has disposeOnUnmount there's no need for mst addDisposer to do that as well

Nevertheless this PR:

1. makes addDisposer return the passed disposer so early disposing is easier
`const d = addDisposer(self, ...); d()`

2. Fixes a potential bug where disposers added via addDisposer on the beforeDestroy hook might never be disposed (by moving the disposer execution to after the "beforeDestroy")

3. Adds some checks to make sure a same disposer is never registered more than once

4. Adds some internal methods that could be useful in the future (hasDisposer, removeDisposer)


xaviergonz:
Thanks :) merging then

PR #1066, opened 2018-10-27
fix: reduxdevtools when flow action is called first

### LINKED ISSUE #1065: ReduxDevTools throws an error - when flow action is called first
An issue from v3.7.0 - after redux middleware refactor https://github.com/mobxjs/mobx-state-tree/commit/d8f449efe11fe688238f0f9264793e562d209ebe

I have an issue when registering redux dev tools. On store creation, I'm immediately performing an API call (which utilizes asynchronous flow  - generator function) which happens inside `afterCreate` hook, the code looks like this:

```
export const Root = types
    .model({
        teams: types.optional(Teams, { data: [] })
    })
    .actions(self => {
        return {
            afterCreate: () => {
                self.teams.loadTeams();
            }
        };
    });

export const store = Root.create({}, { fetch });

if (process.env.NODE_ENV !== 'production') {
    const { connectReduxDevtools } = require('mst-middlewares');
    const remotedev = require('remotedev');

    connectReduxDevtools(remotedev, store);
}
```

But in that case, redux middleware throws an error:

screenshot 2018-10-27 at 21 10 24

I think the problem is in the redux middleware code flow, which doesn't account for `flow_*` being called before any `action` and in that case the `context` is `undefined`.

mkramb:
Providing initial empty context for redux middleware - which fixes #1065

### DIFF
--- a/packages/mst-middlewares/src/redux.ts
+++ b/packages/mst-middlewares/src/redux.ts
@@ -62,12 +62,13 @@ function runMiddleWare(action: any, runners: any[], next: any) {
 // devtools
 
 type ChangesMadeSetter = () => void
+type ActionContextWithNameAndTypePath = Pick<ActionContext, "name" | "targetTypePath">
 
 interface ActionContext {
     parent?: ActionContext
-    name: string
-    targetTypePath: string
-    id: number
+    name?: string
+    targetTypePath?: string
+    id?: number
     runningAsync: boolean
     errored: boolean
     errorReported: boolean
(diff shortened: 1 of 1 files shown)


xaviergonz:
hey thanks for the fix!
can you check what gets logged in that edge case is correct? (or even better, add a unit test for it)

xaviergonz:
actually a better fix might be to use true on addMiddleware so it just logs hooks as standard actions, and therefore the flow would be part of the "afterCreate" action

could you try just changing
` addMiddleware(model, actionMiddleware, true /* from false*/)`
and see how it goes?

xaviergonz:
Closing in favor of #1079

PR #1074, opened 2018-11-07
fix typescript types after type optimization and cast, add castToSnapshot and castToReferenceSnapshot

xaviergonz:
Basically since the fix to make types faster apparently some stuff got broken. This PR fixes:
* Creation type being downgraded to any in some circumstances
* cast being broken and basically becoming a fancy "as any"

Additionally:
* lots of type fixes to by moving IStateTree to the T part of the type rather than to create/model instance type
* added castToSnapshot to cast models used inside a create method.
* added castToReferenceSnapshot to cast models to their reference type (string | null) inside a create method
* cast once more now doesn't compile if used outside an assignment operation
* reference types snapshots are properly typed as string|number (therefore the new castToReferenceSnapshot)
* arrays mutate operations (push, splice, etc) now support creation/snapshot types without the need to cast them
* IComplexType has been removed since now it is equivalent to IType<any, any, ... & IStateTreeNode>


xaviergonz:
@mweststrate thoughts about the addition of castToSnapshot?

Bnaya:
Maybe we could advice the users to use getSnapshot instead?

xaviergonz:
well, that's what mst does internally, but one of its key features is that it also supports constructs such as this
```
function f(childModel: SnapshotOrInstance<typeof ChildModel>) {
  self.rootModel = RootModel.create({
    child: castToSnapshot(childModel)
  });
}
```

to do that otherwise you'd have to either force cast it to `as any as SnapshotIn<typeof ChildModel>`
or with getSnapshot something like
```
function f(childModel: SnapshotOrInstance<typeof ChildModel>) {
  self.rootModel = RootModel.create({
    child: isNode(childModel) ? getSnapshot(childModel) : childModel
  });
}
```
which would be unneeded code (and I think won't compile, not 100% sure)

xaviergonz:
Just made sure the .d.ts files generated don't get bigger (spoiler: they don't, they are exactly the same)

xaviergonz:
@mweststrate pr done

xaviergonz:
Yep I did! Quoting myself
> Just made sure the .d.ts files generated don't get bigger (spoiler: they don't, they are exactly the same)

mweststrate:
Doh! Approved then :)

Op ma 19 nov. 2018 om 16:56 schreef Javier Gonzalez <
notifications@github.com>:

PR #1082, opened 2018-11-20
Properly destroy non-initialized nodes (fixes #1080)

### LINKED ISSUE #1080: Destroying a model with not accessed children results in dangling nodes
**_Bug report_**

 **

> Cannot resolve a reference to type 'Branch' with id: '2' unambigously, there are multiple candidates: /1, /user/branches/1

**

* [x ] I've checked documentation and searched for existing issues
* [ x] I've made sure my project is based on the latest MST version
* [ x] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

https://github.com/feluna/mobx-state-tree-bug-test

When we create a nested data model instance, all of its children should be attached to its parent. 
When we destroy a node, all of its children should also be destroyed. 

If we do not access a child node directly after it's been created, it never gets attached to its parent, therefore calling destroy on the parent node does not destroy those children. Later in the model's lifetime, the destroyed models might be created with the same data. In this case, the new created model contains the same data with some other dangling node.

I have included simple models, and a jest test case with comments to describe what i expect from the destroy call. I don't know if its the expected behavior of attaching to the parents or destroy calls, but i could not find any answers while searching the docs.

k-g-a:
Fixes #1080 

Based on 7d55724fda4b905ffee77e5f6f1c921c23dbae74 (v3.7.1).
This fix also makes nodes' destruction a little bit faster as we used to ```walk``` over subtree, which internaly takes ```storedValue``` and then jumps back to it's ```ObjectNode```.
Now we just use internal ```getChildren()``` on node which also respects non-initialized nodes (i.e. withoud ```storedValue```).


k-g-a:
BTW, shouldn't we keep master in par with latest release? 
For now this PR is blocked by conflicts, meaning this bugfix can't be released untill other changes are verified.

k-g-a:
@xaviergonz could you please take care of changelog? :)

xaviergonz:
Latest release is actually 3.8.0 😁 I'll update the change log and release a 3.8.1

k-g-a:
Ops, seems like I had pulled changes just a few hours before you made 3.8! )

PR #1100, opened 2018-12-06
Reactive isAlive

xaviergonz:
This PR makes isAlive reactive, useful if you want to react when a node becomes dead (for example by setting it to undefined or whatever)

### DIFF
--- a/packages/mobx-state-tree/src/core/node/BaseNode.ts
+++ b/packages/mobx-state-tree/src/core/node/BaseNode.ts
@@ -111,9 +111,9 @@ export abstract class BaseNode {
         Object.keys(this.hookSubscribers).forEach(k => this.hookSubscribers[k].clear())
 
         this.parent = null
-        this.state = NodeLifeCycle.DEAD
         this.subpath = this.escapedSubpath = ""
         this.subpathAtom.reportChanged()
+        this.state = NodeLifeCycle.DEAD
     }
 
     abstract aboutToDie(): void
--- a/packages/mobx-state-tree/src/core/node/BaseNode.ts
+++ b/packages/mobx-state-tree/src/core/node/BaseNode.ts
@@ -6,7 +6,7 @@ import {
     Hook,
     escapeJsonPath
 } from "../../internal"
-import { createAtom, observable } from "mobx"
+import { createAtom } from "mobx"
 
 /**
  * @internal
--- a/packages/mobx-state-tree/src/core/node/BaseNode.ts
+++ b/packages/mobx-state-tree/src/core/node/BaseNode.ts
@@ -50,11 +50,11 @@ export abstract class BaseNode {
     constructor(type: IAnyType, parent: ObjectNode | null, subpath: string, environment: any) {
         this.environment = environment
         this.type = type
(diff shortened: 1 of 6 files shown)


xaviergonz:
Maybe I'm wrong, but I think if isAlive is used outside an "observed" chunk it won't actually consume more memory, like a computed does?

Also, it saves you from making the mistake of using it inside an observer render/reaction/autorun and leave you wondering why the component is not re-rendering / the reaction is not running when the node dies

packages/mobx-state-tree/src/core/node/BaseNode.ts
```diff
@@ -15,9 +15,23 @@ import { createAtom } from "mobx"
...
+    get state() {
+        return this._state
+    }
+    set state(val: NodeLifeCycle) {
+        const wasAlive = this.isAlive
+        this._state = val
+        const isAlive = this.isAlive
+
+        if (wasAlive !== isAlive) {
+            this.aliveAtom.reportChanged()
```
k-g-a: I'm not sure if it's safe to call ```reportChanged()``` if you did not ```reportObserved()``` before?
I mean if you had not used ```observableIsAlive``` and node became ```DEAD``` so we hit this code: wouldnt it accidantely throw? It's just a suggestion, as I am not aquainted with mobx internals deep enough.

xaviergonz: I think it is ok (or else a gazillion unit tests would throw) but I'll change it so the atom is only created when the observable is alive is used and that way we kill two birds with one stone

xaviergonz:
Added 3 small optimizations:
- made is alive atom creation lazy
- made path atom creation lazy
- made escape path generation lazy

packages/mobx-state-tree/src/core/node/BaseNode.ts
mweststrate: `reportChanged()` and `reportObserved()` are allowed to be called without any further assumptions on the other to be called

packages/mobx-state-tree/src/core/node/BaseNode.ts
```diff
@@ -6,40 +6,77 @@ import {
...
         this.environment = environment
         this.type = type
-        this.parent = parent
-        this.subpath = subpath
-        this.escapedSubpath = escapeJsonPath(this.subpath)
+        this.partialSetParent(parent, subpath)
+    }
+
+    private pathAtom?: IAtom
+    protected partialSetParent(parent: ObjectNode | null, subpath: string) {
```
mweststrate: note the most descriptive name. What is "partial" about it? Probably deserves a clearer name or a comment :)

xaviergonz: That there's an abstract setParent, what's a good prefix for internal functions? An underscore? Internal? I'm open to suggestions

mweststrate: I typically use base as well for such cases indeed

xaviergonz: sounds good, renamed partialXXX to baseXXX

PR #1109, opened 2018-12-11
Fix for castFlowReturn returning the actual promise rather than the promise result

xaviergonz:
As discovered in https://github.com/mobxjs/mobx/pull/1827 by @mayorovp when a flow returns a promise it actually waits for it to finish and returns the result rather than returning the promise itself.

This PR fixes that

All this is only related to TS typings, not to the actual runtime.


xaviergonz:
@mweststrate should castFlowReturn be actually removed? Now that I know that returned promises are actually awaited TS users can just write
```
const x = yield somePromise;
return x;
```
rather than
```
return castFlowReturn(somePromise);
```
to achieve the same effect

PR #1113, opened 2018-12-14
Fix for ids not getting unregistered when unacessed nodes die

### LINKED ISSUE #1112: Nested nodes with identifiers are not detached and not removed from the identifiers cache
Works good:

```
const mock1 = [{ id: '1', name: 'Kate' }, { id: '2', name: 'John' }]
const mock2 = [{ id: '3', name: 'Andrew' }, { id: '2', name: 'John' }]

const Entity = types.model({
  id: types.identifier,
  name: types.string
})

const Store = types.model({
  list: types.optional(types.array(Entity), []),
  selectedId: 2
})
.views(self => ({
  get selectedEntity() {
    return resolveIdentifier(Entity, self, self.selectedId)
  }
}))

const store = Store.create()
unprotect(store);

(async function () {
    store.list.replace(mock1)
    await delay(500)
    store.list.replace(mock2)
  })()

// Render selectedEntity
```

Throws an error:

```
const mock1 = [{ id: '1', name: 'Kate' }, { id: '2', name: 'John' }]
const mock2 = [{ id: '3', name: 'Andrew' }, { id: '2', name: 'John' }]

const mock1_2 = mock1.map((i, index) => ({ text: `Text${index}`, entity: i }))
const mock2_2 = mock1.map((i, index) => ({ text: `Text${index}`, entity: i }))

const Entity = types.model({
  id: types.identifier,
  name: types.string
})

const Wrapper = types.model({
  text: types.string,
  entity: Entity,
})

const Store = types.model({
  list: types.optional(types.array(Wrapper), []),
  selectedId: 2
})
.views(self => ({
  get selectedEntity() {
    return resolveIdentifier(Entity, self, self.selectedId)
  }
}))

const store = Store.create()
unprotect(store);

(async function () {
    store.list.replace(mock1_2)
    await delay(500)
    store.list.replace(mock2_2)
  })()

// Render selectedEntity
```

**[mobx-state-tree] Cannot resolve a reference to type 'AnonymousModel' with id: '2' unambigously, there are multiple candidates: /list/1/entity, /list/1/entity**

I tried to call the "detach" method for each existing "Wrapper" entity manually before replacing of the array items, and it works, but i think it's not good idea, because it's difficult to control the all array modifications. For example, someone can call the "applySnapshot" method on any parent node and this fix wouldn't work.

xaviergonz:
Fixes #1112

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,8 +1,10 @@
+-   Fix for unaccessed nodes not unregistering their identifiers [#1112](https://github.com/mobxjs/mobx-state-tree/issues/1112) through [#1113](https://github.com/mobxjs/mobx-state-tree/pull/1113) by [@xaviergonz](https://github.com/xaviergonz)
+
 # 3.9.0
 
-- TypeScript 3.0 or later is now required when using TypeScript. This brings some improvements:
-    - `flow` arguments and return types are now correctly inferred automatically. One exception is when the last return of a `flow` is a `Promise`. In these cases `castFlowReturn(somePromise)` needs to be used so the return type can be inferred properly.
-    - `create` method is now smart enough to warn when no snapshot argument is provided on types that have some mandatory properties.
+-   TypeScript 3.0 or later is now required when using TypeScript. This brings some improvements:
+    -   `flow` arguments and return types are now correctly inferred automatically. One exception is when the last return of a `flow` is a `Promise`. In these cases `castFlowReturn(somePromise)` needs to be used so the return type can be inferred properly.
+    -   `create` method is now smart enough to warn when no snapshot argument is provided on types that have some mandatory properties.
 -   Added `setLivelinessChecking` and `getLivelinessChecking`, the old `setLivelynessChecking` will eventually be deprecated.
 -   Added `onInvalidated` option for references and `types.safeReference` (see readme) through [#1091](https://github.com/mobxjs/mobx-state-tree/pull/1091) by [@xaviergonz](https://github.com/xaviergonz)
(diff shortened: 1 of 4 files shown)


xaviergonz:
I think storedValue only becomes a state tree node once it is accessed

k-g-a:
What I meant by the previous question is: should we really check for ```storedValue``` at all, as after [1082](https://github.com/mobxjs/mobx-state-tree/pull/1082) we do not rely on it durning node's death?
The thing that stopped me from removal of this check is failing "it should trigger lifecycle hooks" test. But this test is quite complex (it pushes strings on hooks into array and then compares those to predefined referene) and I had no time to dig, so I just left the check.
I've just took a look at it and the only difference is that for ```pop()``ing non-initialized value:
- currently we: call ```afterCreate```, then call ```afterAttach```
- after stripping ```storedValue``` check - ```afterAttach``` is not called at all.

k-g-a:
BTW, we do not call beforeDetach for ```pop()```ing, which seems like a bug?

k-g-a:
Just made a quick debug to find out whats going on. We ```die()``` pop'ed node during reconcilation, but then initialize it back. I propose merging this PR to fix the issue. Later on I will look closer and make a separate PR if needed.

xaviergonz:
@k-g-a you lost me there :) I'm ok with waiting if there's a possibly better fix.

xaviergonz:
as suggested by @k-g-a I'll merge this until a possibly better fix is found

PS: maybe pop is not calling beforeDetach but it is calling beforeDestroy?
I'd say that's the right thing to do since strictly speaking the node is being "removed" from the tree

Note: I did not check if that's the case or not

PR #1125, opened 2018-12-28
improve docs by using typedoc

xaviergonz:
This PR:

Uses typedoc instead of document to generate documents, which has the pro of actually using the types included in TS rather than manually having to keep the types up to date with the TS types using jsdocs.
Also, since it uses TS internally it is able to show interfaces/types properly.

A con however is that it shows all overloads, which for compose/union are kind of big.

It is kind of big, but it is mostly about removing useless jsdoc type annotations.


mweststrate:
Approved! Too bad the overloads are so ugly atm :-P.

PR #1133, opened 2019-01-07
Replace InstanceOf mention in @deprecation with Instance

dsabanin:
Probably was forgot there during type rename, as InstanceOf type doesn't exist.


xaviergonz:
Thanks!

PR #1152, opened 2019-01-25
Refactor of the internal type system

xaviergonz:
This is like the refactor that was made for the external typescript types, but this time for the internals of the project without touching the externals.

Maybe I got some types wrong, but it should be a first improvement.

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/custom-type.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/custom-type.test.ts
@@ -37,7 +37,7 @@ class Decimal {
         toSnapshot(value: Decimal) {
             return value.toString()
         },
-        isTargetType(value: string | Decimal): boolean {
+        isTargetType(value: string | Decimal): value is Decimal {
             return value instanceof Decimal
         },
         getValidationMessage(value: string): string {
@@ -120,7 +120,7 @@ class Decimal {
         toSnapshot(value: Decimal) {
             return [value.number, value.fraction]
         },
-        isTargetType(value: [number, number] | Decimal): boolean {
+        isTargetType(value: [number, number] | Decimal): value is Decimal {
             return value instanceof Decimal
         },
         getValidationMessage(value: [number, number]): string {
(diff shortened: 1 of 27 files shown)


xaviergonz:
Thanks for looking at it!
`$treenode` can be made internal for sure, since the actual shape of the `IStateTreeNode` interface really relies on `!!types`, so I'll hide it :)

The other stuff used for typings (usually stuff that starts with `!!`) cannot be made internal, since typescript relies on the shape of the interface rather than the interface name to assert if something is of a given interface "type" or not.

This is why:
```ts
interface A { x: number }
interface B { x: number; }
type Test = B extends A ? true : false; // this is true
```

Once (if ever) they add support for nominal typings (https://github.com/Microsoft/TypeScript/issues/202) it should be possible to remove all that fake stuff.

About is/validate/describe, I remember when making stuff internal that I made it internal but @mweststrate said that they could be kept public since they are useful, so it was kept that way. I guess adding docs is the way to go there.

mweststrate:
Sorry for the late review. Really awesome stuff @xaviergonz ! Neat trick with the unique symbols.

Yeah, we definitely need nominal types on some of this stuff :D. 

P.s. I think it is possible to configure in .gitattributes that diffs on those generated md files are not shown in the future (like binaries). I'll look it up if not familiar with it. 

One question: what does the `this["N"]` and such represent / how does that work?

xaviergonz:
> P.s. I think it is possible to configure in .gitattributes that diffs on those generated md files are not shown in the future (like binaries). I'll look it up if not familiar with it.

I just created a PR with 
```
/API/**/* -diff -merge
```
not sure if that's how it works though

> One question: what does the this["N"] and such represent / how does that work?

It is a trick to avoid having to redeclare the node type in every complex/scalar type. Some kind of "type alias" inside a class (which is not supported)

First I was trying something like
```
export class CustomType<C, S, T, N extends ScalarNode<S, T> = any> extends Type<C, S, T, N> {
```
But that got super annoying and repetitive, so instead I added to BaseType
```
export abstract class BaseType<C, S, T, N extends BaseNode<C, S, T>>
    implements IType<C, S, T & IStateTreeNode<C, S>> {

    // these are just to make inner types available to inherited classes
    readonly C!: C
    readonly S!: S
    readonly T!: T
    readonly N!: N
```

And that way derived classes can just use this["N"] to refer the node type without having to have it all the time in the template part

PR #1153, opened 2019-01-26
Fix broken link in API docs

lostfictions:
The link in `addMiddleware` towards [the middleware doc](https://github.com/mobxjs/mobx-state-tree/blob/master/docs/middleware.md) was broken. This PR fixes it!

### DIFF
--- a/docs/API/README.md
+++ b/docs/API/README.md
@@ -602,7 +602,7 @@ ___
 
 Middleware can be used to intercept any action is invoked on the subtree where it is attached. If a tree is protected (by default), this means that any mutation of the tree will pass through your middleware.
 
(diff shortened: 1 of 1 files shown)


xaviergonz:
Thanks! However that file is autogenerated from a bunch of files (in this case the link comes from https://github.com/mobxjs/mobx-state-tree/blob/66761a3a45c33825297e996e3ff5e88b512d00c0/packages/mobx-state-tree/src/core/action.ts), so to fix it it has to be changed there and then the build-docs command has to be run

xaviergonz:
Did that myself, but thanks for pointing it out! Should be fixed now.

PR #1154, opened 2019-01-26
Fix typos in middleware.md and point to updated CodeSandbox example

lostfictions:
Fixes some small typos in middleware.md. I also noticed the CodeSandbox example provided didn't actually seem to be working -- the middleware that intercepted calls and changed the arguments didn't actually modify the final state. Updating the dependencies seemed to fix it; I also tweaked the example a little bit to try to illustrate the flow of middleware better (eg. logging both before and after calling `next()`).


xaviergonz:
thanks!

PR #1158, opened 2019-01-30
Gitattributes for generated api docs

xaviergonz:


xaviergonz:
@mweststrate is this the right way to do gitattributes?

xaviergonz:
I added .gitattributes to docs in master to see if it would work with this
```
/API/**/* -diff -merge
```
but it seems the PR still shows the changes...

mweststrate:
Fixed! Merging

PR #1165, opened 2019-02-03
More refactors for inner typings, added types.snapshotProcessor

### LINKED ISSUE #947: Feature - types.snapshotProcessor(subModel, { preProcessor, postProcessor})

xaviergonz:
Closes #947 

* Adds/fixes more internal types
* Adds types.snapshotProcessor

`types.snapshotProcessor(type, processors, name?)` runs a pre snapshot / post snapshot processor before/after serializing a given type. Example:
```ts
    const Todo1 = types.model({ text: types.string })
    // in the backend the text type must be null when empty
    interface BackendTodo {
        text: string | null
    }
    const Todo2 = types.snapshotProcessor(Todo1, {
        // from snapshot to instance
        preProcessor(sn: BackendTodo) {
            return {
                text: sn.text || "";
            }
        },
        // from instance to snapshot
        postProcessor(sn): BackendTodo {
            return {
                text: !sn.text ? null : sn.text
            }
        }
    })
```
In the next major version preProcessSnapshot / postProcessSnapshot in model can be deprecated in favor of this one, which works for all kind of types.

Also once deprecated, the "applyProcessor" optional argument of getSnapshot should be removed as well, since anyway it is confusing since it only works when it is applied over a snapshot of a model with processors directly, but not with any inner models/types. The alternative to such parameter is just using getSnapshot over the model itself before it is wrapped inside types.snapshotProcessor


xaviergonz:
Btw, I know that name at the end breaks how usually mst puts optional names at the beginning, but those are just a pain to type, having to have 3 type definitions just because of it (one with name, another without name, then the base one)

README.md
```diff
@@ -290,7 +290,8 @@ For that reason a comma between each member of a model is mandatory, unlike clas
 -   `.props(props)` produces a new type, based on the current one, and adds / overrides the specified properties
 -   `.actions(self => object literal with actions)` produces a new type, based on the current one, and adds / overrides the specified actions
 -   `.views(self => object literal with view functions)` produces a new type, based on the current one, and adds / overrides the specified view functions
--   `.preProcessSnapshot(snapshot => snapshot)` can be used to pre-process the raw JSON before instantiating a new model. See [Lifecycle hooks](#lifecycle-hooks-for-typesmodel)
+-   `.preProcessSnapshot(snapshot => snapshot)` (deprecated in favor of `types.snapshotProcessor`) can be used to pre-process the raw JSON before instantiating a new model. See [Lifecycle hooks](#lifecycle-hooks-for-typesmodel)
```
mweststrate: If it simplifies code, I am pro bringing out a next major quickly. But it is nice to add in a minor first to see whether there are cases we can't handle

packages/mobx-state-tree/src/types/complex-types/map.ts
```diff
@@ -393,17 +392,19 @@ export class MapType<IT extends IAnyType> extends ComplexType<
...
         const currentKeys: { [key: string]: boolean } = {}
         Array.from(target.keys()).forEach(key => {
             currentKeys[key] = false
         })
-        // Don't use target.replace, as it will throw all existing items first
-        for (let key in snapshot) {
-            target.set(key, snapshot[key] as any)
-            currentKeys["" + key] = true
+        if (snapshot) {
+            // Don't use target.replace, as it will throw all existing items first
```
mweststrate: throw away ?

packages/mobx-state-tree/src/types/utility-types/late.ts
```diff
@@ -87,6 +89,11 @@ export class Late<C, S, T> extends Type<C, S, T, false> {
         const t = this.getSubType(false)
         return t ? t.isAssignableFrom(type) : false
     }
+
+    getSubTypes() {
+        const subtype = this.getSubType(false)
+        return subtype ? subtype : "cannotDetermine"
```
mweststrate: NIT: use a constant? (also on the previous locations)

packages/mobx-state-tree/src/types/utility-types/snapshotProcessor.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    reconcile(current: this["N"], newValue: this["C"] | this["T"]): this["N"] {
+        const node = this._subtype.reconcile(
+            current,
+            isStateTreeNode(newValue) ? newValue : this.preProcessSnapshot(newValue)
+        )
+        if (node !== current) {
+            const oldGetSnapshot = node.getSnapshot
+            node.getSnapshot = () => {
```
mweststrate: I can imagine there is a quite unlikely edge case to run into, suppose you have

```ts
const Store = types.model({
  a: types.array(types.snapshotPreprocessor(Task)),
  b: types.array(Task)
})
```

What would happen if a Task is moved from `a` to `b` and vice versa? Since the nodes stay alive, the `getSnapshot` override is not (un)done, correct? Or is it forbidden anyway because `Task` is not assignable to `snapshotProcessor(Task)`?

README.md
xaviergonz: sounds good, anyway this PR doesn't remove the old hooks (which I guess brings the eternal question, is something deprecated after it is no longer available or before? :) )

packages/mobx-state-tree/src/types/complex-types/map.ts
xaviergonz: fixed

packages/mobx-state-tree/src/types/utility-types/late.ts
xaviergonz: fixed

README.md
xaviergonz: renamed the wording so it is alternatively instead of deprecated

packages/mobx-state-tree/src/types/utility-types/snapshotProcessor.ts
xaviergonz: you mean with an actual processor that does something or with no processors at all? I can create a unit test to see what would happen.
What is the wanted behavior when a) the snapshot processor does some change and b) when it doesn't?

xaviergonz: @mweststrate I see now what you mean. Added unit tests and disallowed assignations unless they are of the exact same "snapshot processor" type.

Had to do it in a kind of "hackish" way, but it works.

That method could also be applied to optionals in case optionals should not be assignable to non-optionals (or vice versa)

PR #1175, opened 2019-02-08
fix detaching killing nodes

### LINKED ISSUE #1173: Detaching an array eliminates all of it's children
**_Bug report_**

```detach```'ing array from ```model``` effectively eliminates all the children
```js
const Model = types.model(items: types.array(SomeModel));
const model = Model.create(items: [{}, {}, {}, ...])
const detachedItems = detach(model.items);
detachedItems[0] // is undefined
model.items = detachedItems // does nothing, as there are no items
```

[Reproduction](https://codesandbox.io/s/3v530zoz76)

Whole array with it's contents whould be detached, so it can be reused later on.

The array is detached and preserved, but it looses own content making its quite useless.

xaviergonz:
Fixes #1173


packages/mobx-state-tree/src/core/node/object-node.ts
```diff
@@ -453,12 +453,17 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
...
 
-        this.environment = this.root.environment // make backup of environment
-        this.identifierCache = this.root.identifierCache!.splitCache(this)
-        this.parent!.removeChild(this.subpath)
-        this.baseSetParent(null, "")
+        const newEnv = this.root.environment
+        const newIdCache = this.root.identifierCache!.splitCache(this)
 
-        this.state = NodeLifeCycle.FINALIZED
+        try {
```
k-g-a: Just wonder, why did it become a try/catch?

xaviergonz: To recover properly in case the conversion to undefined upon detaching fails because it's not valid

xaviergonz: It will rethrow the error anyway, but at least it won't leave the node in an inconsistent state

PR #1178, opened 2019-02-11
getValue small refactor

xaviergonz:
A small refactor around getValue, moving its main implementation to the simple/complex type implementation rather than to the single types themselves

Also fixes a possible issue where createObservableInstanceIfNeeded was not getting called for any complex types other than model

### DIFF
--- a/packages/mobx-state-tree/src/core/node/BaseNode.ts
+++ b/packages/mobx-state-tree/src/core/node/BaseNode.ts
@@ -5,7 +5,9 @@ import {
     escapeJsonPath,
     EventHandlers,
     IAnyType,
-    IDisposer
+    IDisposer,
+    SimpleType,
+    ComplexType
 } from "../../internal"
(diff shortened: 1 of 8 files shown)


mweststrate:
Just curious, would this speed things up?

xaviergonz:
I doubt it will speeds things up, it was just to tidy things up a bit.
What could maybe speed things up however is moving all node methods to plain functions and their data to plain objects. But just in theory.

mweststrate:
Yeah tried that a few times before, but the impact was usually at best
equal. Classes (constructor functions) create inner classes which optimize
very well on current engines

xaviergonz:
@mweststrate same, I made a speedup branch (https://github.com/mobxjs/mobx-state-tree/tree/speedup-node) that used plain objects + plain functions and it performed the same or even a bit worse while making the code uglier

PR #1188, opened 2019-02-20
add optionalNull type

xaviergonz:
Basically like the types.optional type, except that it transforms null into a default value rather than undefined.
That way it should be easier for people to transform their nulls from backend into default values.


packages/mobx-state-tree/src/core/node/BaseNode.ts
```diff
@@ -43,9 +43,7 @@ export abstract class BaseNode<C, S, T> {
 
     storedValue!: any // usually the same type as the value, but not always (such as with references)
     get value(): T {
-        return (this.type as SimpleType<any, any, any> | ComplexType<any, any, any>).getValue(
-            this as any
-        )
+        return (this.type as any).getValue(this)
```
xaviergonz: had to change this since the docs tool uses an older TS version that didn't understand this

PR #1190, opened 2019-02-21
types.optionalNull improvements

### LINKED ISSUE #1189: bootstrap command should be removed from package.json
**_Bug report_**

* [ ] I've made sure my project is based on the latest MST version

I am using node v8.11.1 , npm v5.6.0, yarn v1.12.3 on Windows 10.
I had cloned the mobx-state-tree repository. Then I ran `yarn bootstrap`.

 I expect all dependencies to be installed.

I get an error message:

```C:\Users\<user>\projects\mobx-state-tree> yarn bootstrap
yarn run v1.12.3
$ npx lerna@3.3.0 bootstrap --hoist
npx: installed 637 in 33.875s
lerna notice cli v3.3.0
lerna ERR! EWORKSPACES --hoist is not supported with --npm-client=yarn, use yarn workspaces instead
lerna ERR! EWORKSPACES A guide is available at https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/
error Command failed with exit code 1.
```

xaviergonz:
Actually the other types.optionalNull implementation had some mistakes:
- if undefined was passed as snapshot and it was not a valid value for the subtype the default would be used nevertheless
- it was not possible to get undefined as output snapshot if undefined was a valid value
- some other stuff

Therefore I've done an alternate implementation, based on adding some params to the constructor of OptionalValue

The thing is, @mweststrate  should it really be there as a standalone `types.optionalNull` or should the "replaced value" just be a third parameter for types.optional that defaults to undefined and that could be changed to any kind of primitive / an array of primitives? This is:
```
types.optional(type, valOrFunc)
types.optional(type, valOrFunc, null) // would replace types.optionalNull
types.optional(type, valOrFunc, 5)
types.optional(type, valOrFunc, [null, undefined]) // would use the default value when the input snapshot is either undefined or null
```

Also on this PR:
- Removed IOptionalProperty and instead it checks if the Creation Snapshot has undefined as a possible value to decide if a property is optional or not
- Removed useless bootstrap command in favor of plain yarn install - Fixes #1189


xaviergonz:
Declined in favor of PR #1192

PR #1192, opened 2019-02-22
improve types.optional rather than add types.optionalNull

### LINKED ISSUE #1189: bootstrap command should be removed from package.json

xaviergonz:
Replaces PR #1190 

Actually the other types.optionalNull implementation that was merged had some mistakes:
- if undefined was passed as snapshot and it was not a valid value for the subtype the default would be used nevertheless
- it was not possible to get undefined as output snapshot if undefined was a valid value
- some other stuff

After some thought though I figured out a possibly better way to do it is that, instead of adding yet another type (optionalNull), rather to add an optional third argument to the current types.optional.
This third argument would be an array of primitives (string | number | boolean | null | undefined) that would be considered optional (that would be replaced by the default value)
```
types.optional(type, valOrFunc) // as usual, equivalent to types.optional(type, valOrFunc, [undefined])
types.optional(type, valOrFunc, [null]) // would replace types.optionalNull
types.optional(type, valOrFunc, [0]) // 0 would be replaced by the default
types.optional(type, valOrFunc, [null, undefined]) // would use the default value when the input snapshot is either undefined or null
```

Also on this PR:
- Removed IOptionalProperty and instead it checks if the Creation Snapshot has undefined as a possible value to decide if a property is optional or not
- Removed useless bootstrap command in favor of plain yarn install - Fixes #1189


packages/mobx-state-tree/__tests__/core/optional-extension.test.ts
```diff
@@ -0,0 +1,264 @@
+import { getSnapshot, types, unprotect } from "../../src"
+
+describe("null as default", () => {
+    describe("basic tests", () => {
+        const M = types.model({
```
mweststrate: test suggestion, with `[null]`, `x: undefined` or `y` being omitted should give type errors

packages/mobx-state-tree/src/types/utility-types/optional.ts
```diff
@@ -181,20 +170,66 @@ export function optional<IT extends IAnyType>(
...
+export function optional<IT extends IAnyType, OptionalVals extends ValidOptionalValues>(
+    type: IT,
+    defaultValueOrFunction: OptionalDefaultValueOrFunction<IT>,
+    optionalValues: OptionalVals
+): IOptionalIType<IT, OptionalVals>
+/**
+ * `types.optional` - Can be used to create a property with a default value.
+ * If the given value is not provided in the snapshot (or matches one of the provided optional values), it will default to the provided `defaultValue`.
+ * If `defaultValue` is a function, the function will be invoked for every new instance.
+ * Applying a snapshot in which the optional value is _not_ present causes the value to be reset
```
mweststrate: I think this API is way better! Super neat!

To see if I understand description correctly:

```javascript
const t = types.model({
  x: types.optional(types.number, 3, [0, 1])
})

t.create({x : 0 }) // x = 3 (would expect 0 though, since it matches types.number?)
t.create({ x: undefined }) // error (not missing, not in the optional set) (should `undefined` be implied always?)
t.create({ }) // x = 3
```

mweststrate: optional is basically a snapshot pre-processor :thinking:

xaviergonz: except that if when used as an snapshot pre-processor (such as the old implementation) then this:
```
const m = M.create({
  x: undefined
})
```
was still "valid" and transformed into the default value, since in the inside it transformed 0 || 1 into undefined for types.optional, and types.optional would accept that undefined

xaviergonz: A more clear example
```
const M = types.model({ 
  x: types.optionalNull(types.number, 10),
  y: types.optionalNull(types.number, 10)
})

// with the old implementation that uses only a pre-snapshot processor + types.optional
const m = M.create({
  x : undefined  // NOT ok, x will become 10, since it only transforms null into undefined, and undefined will become 10
  y: null // ok, will become undefined then undefined will become 10
})
// with the new implementation
const m = M.create({
  x : undefined, // ok, will throw that undefined is not assignable to number
  y: null // ok, y will become 10
})
```

packages/mobx-state-tree/__tests__/core/optional-extension.test.ts
xaviergonz: they do (tested manually), but how to test typings automatically? :-/

packages/mobx-state-tree/src/types/utility-types/optional.ts
xaviergonz:
```
const t = types.model({
  x: types.optional(types.number, 3, [0, 1])
})

t.create({x : 0 })
// x = 3, since 3 is an optional value (imagine that for some reason -1 has to be turned into an empty string or whatever, who knows

t.create({ x: undefined })
// error (and TS error) (not missing, not in the optional set), undefined shouldn't be implied since it is not declared as a valid optional value, if wanted then the array could just be [0, 1, undefined]
// this allows to validate stuff such as that the back-end value can be null, but if undefined that's an error

t.create({ }) // error, not 0 or 1 or number, if this should work then use [ 0, 1, undefined ] instead (this will also be a TS error if undefined is not a valid value)
```

xaviergonz: So answering the question, optional is a type pre-processor but with input snapshot validation on top I guess :)

xaviergonz:
Ah, another thing added by this PR I guess is that if a value can take undefined then it can be missing
e.g.
```
const  M = types.model({ x: types.union(undefined, types.number) })
const m = M.create({x: undefined }) // is ok
const m = M.create({ }) // is ok as well (even in TS)
```

xaviergonz:
Some more clarifications
```
types.optional(types.number, 5) // snapshotIn: number | undefined (optional, can be missing)
types.optional(types.number, 5, ['empty']) // snapshotIn: number | 'emtpy' (cannot be missing)
types.optional(types.number, 5, [undefined, 'empty']) // snapshotIn: number | 'emtpy' | undefined (optional, can be missing)
types.optional(types.number, 5, [null]) // snapshotIn: number | null (cannot be missing)
types.optional(types.number, 5, [-1, undefined]) // snapshotIn: number | undefined (optional, can be missing)
```
for all of them the snapshoutOut is number and the instance type is number

packages/mobx-state-tree/__tests__/core/optional-extension.test.ts
xaviergonz: I think I know now what you mean. added

```ts
        const M = types.model({
            a: types.optional(types.number, 5, [null]),
            b: types.optional(types.number, 6, [null])
        })

        {
            const m = M.create({
                a: null,
                b: null
            })
            expect(m.a).toBe(5)
            expect(m.b).toBe(6)
        }

            expect(() => {
                M.create({
                    a: null,
                    b: undefined as any // undefined is not valid
                })
            }).toThrowError("value `undefined` is not assignable to type: `number`")

            expect(() => {
                M.create({
                    a: null
                    // b: null missing, but should be there
                } as any)
            }).toThrowError("value `undefined` is not assignable to type: `number`")
        }
```

packages/mobx-state-tree/src/types/utility-types/optional.ts
xaviergonz: Added clarifications to the function docs

xaviergonz:
Just removed applyOptionalValuesToSnapshot, since apparently they were doing nothing for a long time

Apparently now this
```
processInitialSnapshot(childNodes: IChildNodesMap): this["S"] {
        const processed = {} as any
        Object.keys(childNodes).forEach(key => {
            processed[key] = childNodes[key].getSnapshot() // this already includes the replaced value from the default
        })
        // so applyOptionalValuesToSnapshot would always get the optional values already applied
        return this.applySnapshotPostProcessor(applyOptionalValuesToSnapshot(processed))
    }
```

I even run all tests with a console log to see what that function was doing and all it was doing all the time was replacing undefined by undefined values

Added some unit tests to make sure it is ok without that code though, all passing

xaviergonz:
Btw, maybe it is true and types.snapshotProcessor could be used to do optional

Something like 
```
function typesOptional(type, default, optionalValues) {
  const opt = types.snapshotPreprocessor(type, {
    preprocess(sn) {
      if (optionalValues.includes(sn)) return default; // or default() if a function
      return sn;
    }
  })
  return types.refinement(type.name + "?", opt)
}
```
the only difference is that it wouldn't be able to use actual instances as default, they'd need to be transformed into snapshots, so their node state would be lost

k-g-a:
> Just removed applyOptionalValuesToSnapshot, since apparently they were doing nothing for a long time

Yep, good catch! Its obsolete since ```_getCachedInitialSnapshot()``` introduction, which fixed pre/post process snapshot bugs short after v3 relese! :) I've checked that you had removed related dead code either.

> the only difference is that it wouldn't be able to use actual instances as default

It seems to me that using actual instance for default value should not be allowed at all, becasue:
- if one creates in in-place: ```types.optional(MyType, () => MyType.create(MY_TYPE_SNAPSHOT))``` - it's just a more verbose way to do ```types.optional(MyType, () => MY_TYPE_SNAPSHOT)``` - no performance benefits, just moving the same work to another place in the call tree; I think people go that way, becasue it's not clear enough, that __new instance__ will be created every time under the hood;
- if one uses imported (or somehow obtained, i.e. via DI-container) instance: in this case there is a high chance to get errors about 'it is already part of the same or another state tree' - because one should use reference, not instance to reuse nodes. We used to stumble upon it in our codebase for a few times at the beginning, becuase we did not understand how MST works internally.

So I upvote using ```preProcessor```. 
But even if we decide not to, I'd forbid returning instances from ```defaultInstanceOrSnapshot``` at least type level.

xaviergonz:
So I see these options:

1. keep the PR as is, enhancing types.optional
  - (breaking) plus maybe in the future re-implement the enhanced types.optional as a plain snapshotprocessor, dropping the instance as a default support
2. remove types.optional enhancements but keep the typing improvements so that any snapshotProcessors that takes undefined as input become automatically optional for TS
  - (breaking) plus maybe in the future re-implement the good old types.optional as a plain snapshotprocessor, dropping the instance as a default support

Btw, once thing that can be done when using instances as default is something like this
```
types.optional(M, () => {
  const m = M.create(...)
  m.doSomethingWithVolatileOrInternalState()
});
```

but given how many times I've seen people just do 
```
const def = M.create(...)
types.optional(M, () => def)
```
and never the other one I'm not sure if that's enough to keep it.

xaviergonz:
@mweststrate which option from the two above should this pr take?

mweststrate:
@xaviergonz I think I would go for 1. Maybe add: print a warning (once) if instance is returned from the default function? @k-g-a strong opinion about it?

xaviergonz:
Since the warning would be a new feature I'll merge it in the meanwhile :)

PR #1196, opened 2019-03-02
undo manager improvements, new createActionTrackingMiddleware2

### LINKED ISSUE #1195: UndoManager Middleware withoutUndo() used inside an action affects all patches within that action
**_Bug report_**

### Version info:

mobx 5.9.0

mobx-state-tree 3.10.2

mst-middlewares 3.10.2

When using the UndoManager middleware's `withoutUndo` method, in an action which causes multiple patches, if any one of the patch-creating changes is made within a `withoutUndo` call, all of the patches from that action are excluded from the undo history.

https://codesandbox.io/s/7mxk54x7nx?fontsize=14
```
const { types } = require(`mobx-state-tree`);
const { UndoManager } = require(`mst-middlewares`);

let undoManager = {};
const setUndoManager = targetStore => {
    undoManager = UndoManager.create({}, { targetStore });
};

const store = types
    .model(`Person`, {
        name: types.string,
        age: types.number
    })
    .actions(self => {
        setUndoManager(self);

        return {
            setName(name) {
                undoManager.withoutUndo(() => {
                    self.name = name;
                });
            },

            setAge(age) {
                self.age = age;
            },

            setInfo({ name, age }) {
                self.setName(name);
                self.setAge(age);
            }
        };
    });


const person = store.create({ name: `Jon`, age: 20 });

person.setInfo({ name: `Bob`, age: 50 });
console.log(undoManager.history.length); // 0
```
 
I would expect this to generate 2 patches, 1 for each property changed. I would expect the patch to the `name` property not to be recorded in the undo history and the patch for the `age` property to be added to the undo history.

No patches are added to the undo history at all. If you remove the `withoutUndo`, 2 patches are created in the history.

xaviergonz:
**EDIT:** please skip this and read https://github.com/mobxjs/mobx-state-tree/pull/1196#issuecomment-473595557 instead

Fixes #1195 

* undo manager:
  * refactor to undo manager so it is easier to read and more resilient to weird usages (e.g. using undo/redo inside other actions, using flows within withoutUndoFlows, etc)
  * added an optional object to clear history in case the user wants to only clear the redo and/or undo
  * added undoLevels / redoLevels to undo manager
  * made undo/redo atomic
* createActionTrackingMiddleware 
  * fixes so the flow steps use the proper context from the closest action that spawned them (rather than the context of the root action always)
  * fix for a possible leak when using filter (it might not clear the runningActions when filtering was enabled for a particular action and flows were spawned)
  * fix for filtering so it doesn't allow flow steps if the a parent action is supposed to be filtered (this one might be breaking, but I think it is a fix)
  * fix so finishing flows don't remove the root action from the runningActions (ending in the dreaded error when root.context was accessed when root was undefined)
  * fix for sync actions where suspend was not being called after resume and before fail


packages/mobx-state-tree/src/core/mst-operations.ts
```diff
@@ -210,7 +210,7 @@ export function recordPatches(subject: IAnyStateTreeNode): IPatchRecorder {
             applyPatch(target || subject, recorder.patches)
         },
         undo(target?: IAnyStateTreeNode) {
-            applyPatch(target || subject, recorder.inversePatches.slice().reverse())
+            applyPatch(target || subject, recorder.inversePatches.reverse())
```
mweststrate: The `.slice` was here as `reverse` tends to update in place

xaviergonz: that was done because it uses recorder.inversePatches, which is:
```
get inversePatches() {
            return this.rawPatches.map(([_, b]) => b)
        },
```
and since map does a copy it means the slice makes a copy of a copy

xaviergonz: I think ill change it so it doesn't need to do all those transformations

xaviergonz: Improved in this commit: https://github.com/mobxjs/mobx-state-tree/pull/1196/commits/0df3a2ff5145bfc278ae7dfae4b248f423950ba6

xaviergonz:
Fixes #1195 now

Ok, in order to not break things too much I changed it so it adds a "createActionTrackerMiddleware2" (new name suggestions welcome)

What's new with this one? it is a VASTLY simplified version of createActionTrackerMiddleware with only 3 hooks:

```
export interface IActionTrackingMiddleware2Hooks<TEnv> {
    filter?: (call: IActionTrackingMiddleware2Call<TEnv>) => boolean
    onStart: (call: IActionTrackingMiddleware2Call<TEnv>) => void
    onFinish: (call: IActionTrackingMiddleware2Call<TEnv>, error?: any) => void
}
```
it might look simpler but in reality it is actually better at helping users create a middleware that handles both sync and async actions in a more universal way. How?
1. it only cares about actions, when they start and end (no more flow resume, error, suspend, resume, etc crazyness)
2. when it means actions it means both sync actions and flow async actions
3. sync actions run as usual, they start and then they finish, and inner actions start and finish in the middle
4. async actions start as usual, but they FINISH once its own flow is finished (and they also have their inner actions start and finish in the middle)

Taken from the docs:
```
 * The flow is like this:
 * - for each action: if filter passes -> `onStart` -> (inner actions recursively) `onFinish`
 *
 * Example: if we had an action `a` that called inside an action `b1`, then `b2` the flow would be:
 * - `filter(a)`
 * - `onStart(a)`
 *   - `filter(b1)`
 *   - `onStart(b1)`
 *   - `onFinish(b1)`
 *   - `filter(b2)`
 *   - `onStart(b2)`
 *   - `onFinish(b2)`
 * - `onFinish(a)`
 *
 * The flow is the same no matter if the actions are sync or async.
```

So it doesn't matter if a, b1 and b2 are sync or async, the hooks are called exactly the same.

Other goodies are:
- call.name, args, etc are always present and pointing to the proper current action (thing which didn't happen with flow event types)
- each call has a pointer to the parentCall (if available), so the call stack of actions can be easily traversed to the root if needed
- the call includes an environment (call.env), which is shallow copied to the child call and serves the same purpose as the old context (and since it is editable in any phase, not only in start, it can be set for example in the filter phase if needed for whichever reason) <- I'm open to alternative suggestions about this one
- since call.env is available in call, and call.parentCall is too, you can even traverse back to "parent" environments (contexts) if desired
- it is unit tested (the old one wasn't at all, and I found some funny weird bugs)

Other stuff included in this PR:
* undo manager:
  * uses the new createActionTrackerMiddleware2
  * refactor to undo manager so it is easier to read and more resilient to weird usages (e.g. using 
  * undo/redo inside other actions, using flows within withoutUndoFlows, etc) - covered by unit tests
  * added an optional object to clear history in case the user wants to only clear the redo and/or undo
  * added undoLevels / redoLevels to undo manager
  * made undo/redo atomic
* recordPatches
  * include a "recording" filed to know if a recorder is currently recording or not
  * the method now has an optional filter function to skip certain patches
  * improvements to recorderPatches performance so it copies arrays around only when needed (on demand)
* atomic
  * uses the new createActionTrackerMiddleware2
  * added unit tests to make sure inner actions and decorate work properly

cc @mweststrate 
cc @k-g-a

xaviergonz:
You mean measured using the current perf suite?

xaviergonz:
Doesn't seem to impact much

this branch
```
-----------
Small Model
-----------
      35ms |      x 100 |  0.3ms avg
     125ms |    x 1,000 |  0.1ms avg
     565ms |   x 10,000 |  0.1ms avg
     121ms |    x 1,000 |  0.1ms avg
      16ms |      x 100 |  0.2ms avg

------------
Medium Model
------------
      33ms |      x 100 |  0.3ms avg
     240ms |    x 1,000 |  0.2ms avg
   1,270ms |   x 10,000 |  0.1ms avg
     208ms |    x 1,000 |  0.2ms avg
      25ms |      x 100 |  0.3ms avg

------------------------
Large Model - 0 children
------------------------
      85ms |      x 100 |  0.8ms avg
     438ms |    x 1,000 |  0.4ms avg
      70ms |      x 100 |  0.7ms avg

-------------------------------------------
Large Model - 10 small & 10 medium children
-------------------------------------------
      77ms |       x 50 |  1.5ms avg
     232ms |      x 250 |  0.9ms avg
     101ms |       x 50 |  2.0ms avg

-------------------------------------------
Large Model - 100 small & 0 medium children
-------------------------------------------
      63ms |       x 50 |  1.3ms avg
     227ms |      x 250 |  0.9ms avg
      82ms |       x 50 |  1.6ms avg

-------------------------------------------
Large Model - 0 small & 100 medium children
-------------------------------------------
     199ms |       x 50 |  4.0ms avg
     946ms |      x 250 |  3.8ms avg
     236ms |       x 50 |  4.7ms avg

16.31user 3.35system 0:09.44elapsed 208%CPU (0avgtext+0avgdata 278212maxresident)k
0inputs+0outputs (0major+200993minor)pagefaults 0swaps

```

master
```
-----------
Small Model
-----------
      23ms |      x 100 |  0.2ms avg
     119ms |    x 1,000 |  0.1ms avg
     565ms |   x 10,000 |  0.1ms avg
     112ms |    x 1,000 |  0.1ms avg
      17ms |      x 100 |  0.2ms avg

------------
Medium Model
------------
      34ms |      x 100 |  0.3ms avg
     216ms |    x 1,000 |  0.2ms avg
   1,245ms |   x 10,000 |  0.1ms avg
     228ms |    x 1,000 |  0.2ms avg
      25ms |      x 100 |  0.3ms avg

------------------------
Large Model - 0 children
------------------------
      81ms |      x 100 |  0.8ms avg
     474ms |    x 1,000 |  0.5ms avg
      63ms |      x 100 |  0.6ms avg

-------------------------------------------
Large Model - 10 small & 10 medium children
-------------------------------------------
      82ms |       x 50 |  1.6ms avg
     286ms |      x 250 |  1.1ms avg
      71ms |       x 50 |  1.4ms avg

-------------------------------------------
Large Model - 100 small & 0 medium children
-------------------------------------------
      70ms |       x 50 |  1.4ms avg
     219ms |      x 250 |  0.9ms avg
      66ms |       x 50 |  1.3ms avg

-------------------------------------------
Large Model - 0 small & 100 medium children
-------------------------------------------
     224ms |       x 50 |  4.5ms avg
     907ms |      x 250 |  3.6ms avg
     225ms |       x 50 |  4.5ms avg

15.89user 3.07system 0:09.48elapsed 199%CPU (0avgtext+0avgdata 274716maxresident)k
0inputs+0outputs (0major+201542minor)pagefaults 0swaps
```

PR #1205, opened 2019-03-08
fix undead nodes on array reconciliation / when using pop, shift, splice (ignore red x, it is coveralls)

xaviergonz:
I found out that sometimes while performing array reconciliation some dead nodes wouldn't actually die, this PR fixes this.

This is specially awkward when using array methods that remove a node and also return them (pop, shift, splice, etc). Before this PR it would return "undead nodes", this is, nodes that would be killed (with all the lifecycle being called) yet they would be resurrected by the mobx enhancer by calling createObservableIfNeeded again in order to return them from those methods.

After this PR these nodes will actually die, following their natural lifecycle, but if the dehancer tries to return them again what will be returned is a clone of such nodes. The bad side effect is that such auto-creation is done even when the user won't use the returned values (since it cannot know beforehand), but I still think that's better than returning a half-broken node in a inconsistent state as right now.

The second option is to disable pop/shift/splice from returning any complex values at all (for dead nodes) or alternatively return their snapshots, but maybe that'd be a breaking change? Though the good thing is that no nodes would be auto created when using such methods.

A third option would be to detach the node instead, but that'd mean that pop-ed nodes wouldn't die on their own and the user would need to call destroy on them manually :-/

Personally I like the second option best, but that's a breaking change for sure, so maybe add a new parameter option to types.array ('fast' | 'slow' defaulting to 'slow' for compatibility) that will change the pop/push/splice implementation to return nothing on fast mode?

**UPDATE:** **Ok, better idea implemented. pop/shift/etc will actually return dead nodes, but will ensure they are finalized before killing them, so the user will have to use clone/getSnapshot over them manually if he wants to access their data (or deal with the warning/error of liveliness data depending on the selected setting).**

This way the nodes will follow their proper life-cycle before dying, there's no auto cloning involved and liveliness error checking will kick in if they try to access their data for some reason.

@mweststrate @k-g-a thoughts on this?

Additionally:
* added extra assertions to check everything is ok (not hit by coveralls, that's why the coverage decreased)
* fixed some nodes not actually following the proper lifecycle when dying
* uses node.environment rather than node.root.environment wherever possible for a very slight speed boost (no need to traverse the tree to the root all the time, since environments are already on the node itself)

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/hooks.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/hooks.test.ts
@@ -86,15 +86,9 @@ test("it should trigger lifecycle hooks", () => {
 
     events.push("--")
 
-    // access (new, attach) get biscuit
-    const oldBiscuit = store.todos[store.todos.length - 1]
-
-    // destroys the old get biscuit, news a new cloned biscuit
-    const clonedBiscuit = store.todos.pop()!
+    // access (new, attach), then destroy biscuit
+    const oldBiscuit = store.todos.pop()!
     expect(isAlive(oldBiscuit)).toBe(false)
-    expect(isAlive(clonedBiscuit)).toBe(true)
-    expect(hasParent(clonedBiscuit)).toBe(false)
-    expect(oldBiscuit).not.toBe(clonedBiscuit)
 
     events.push("---")
     // new and then attach "add sugar"
@@ -113,11 +107,6 @@ test("it should trigger lifecycle hooks", () => {
     destroy(talk)
     expect(isAlive(talk)).toBe(false)
 
-    events.push("------")
-    // destroy "Give biscuit"
-    destroy(clonedBiscuit)
-    expect(isAlive(clonedBiscuit)).toBe(false)
-
     expect(events).toEqual([
         "new store: 3",
         "-",
@@ -130,7 +119,6 @@ test("it should trigger lifecycle hooks", () => {
         "destroy todo: Get biscuit",
         "custom disposer 2 for Get biscuit",
         "custom disposer 1 for Get biscuit",
-        "new todo: Get biscuit",
         "---",
         "new todo: add sugar",
         "attach todo: add sugar",
@@ -143,11 +131,7 @@ test("it should trigger lifecycle hooks", () => {
         "-----",
         "destroy todo: Give talk",
         "custom disposer 2 for Give talk",
-        "custom disposer 1 for Give talk",
-        "------",
-        "destroy todo: Get biscuit",
-        "custom disposer 2 for Get biscuit",
-        "custom disposer 1 for Get biscuit"
+        "custom disposer 1 for Give talk"
     ])
 })
 
--- a/packages/mobx-state-tree/__tests__/core/hooks.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/hooks.test.ts
@@ -85,9 +85,6 @@ test("it should trigger lifecycle hooks", () => {
     expect(hasParent(talk)).toBe(false)
 
     events.push("--")
-    // "Get biscuit" never gets actually created upon removal
-    // no attach/detach since it didn't even get to be attached
-    // and since a clone is returned, then "new" and should be alive
 
     // access (new, attach) get biscuit
     const oldBiscuit = store.todos[store.todos.length - 1]
--- a/packages/mobx-state-tree/__tests__/core/hooks.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/hooks.test.ts
@@ -85,12 +85,19 @@ test("it should trigger lifecycle hooks", () => {
     expect(hasParent(talk)).toBe(false)
 
     events.push("--")
-    // access (new) but immediately destroy "Get biscuit"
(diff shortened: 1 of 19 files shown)


xaviergonz:
Ok, better idea implemented. pop/shift/etc will actually return dead nodes, but will ensure they are finalized before killing them, so the user will have to use clone over them manually if he wants to access their data.

This way the nodes will follow their proper life-cycle before dying, there's no auto cloning involved and liveliness error checking will kick in if they try to access them for some reason.

packages/mobx-state-tree/__tests__/core/hooks.test.ts
```diff
@@ -73,33 +75,60 @@ function createTestStore(listener: (s: string) => void) {
...
+        "-",
         "new todo: Give talk",
         "attach todo: Give talk",
         "detach todo: Give talk",
-        "-",
+        "--",
         "new todo: Get biscuit",
         "attach todo: Get biscuit",
-        "--",
+        "destroy todo: Get biscuit",
```
xaviergonz: as you see here, previously pop-ed nodes wouldn't actually get destroyed

packages/mobx-state-tree/src/core/mst-operations.ts
```diff
@@ -729,7 +729,7 @@ export function clone<T extends IAnyStateTreeNode>(
     return node.type.create(
         node.snapshot,
         keepEnvironment === true
-            ? node.root.environment
+            ? node.environment
```
xaviergonz: we can use environment from the node directly since they properly have the same environment as their parent all the way to the root
that way we don't need to traverse the tree all the way to the root

packages/mobx-state-tree/src/core/node/object-node.ts
```diff
@@ -495,17 +494,16 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
 
     @action
     die(): void {
-        if (this.state === NodeLifeCycle.DETACHING) return
-        if (this._observableInstanceState === ObservableInstanceLifecycle.CREATED) {
-            this.aboutToDie()
-            this.finalizeDeath()
-        } else {
-            // get rid of own and child ids at least
-            this.unregisterIdentifiers()
```
xaviergonz: this hack is not needed anymore since now array children are properly destroyed when poped/shifted

packages/mobx-state-tree/src/types/complex-types/array.ts
```diff
@@ -333,11 +338,16 @@ function reconcileArrayChildren<TT>(
...
         } else if (!hasNewNode) {
-            // new one does not exists, old one dies
-            oldNode.die()
+            // new one does not exists
+            nothingChanged = false
             oldNodes.splice(i, 1)
+            if (oldNode instanceof ObjectNode) {
+                // since it is going to be returned by pop/splice/shift better create it before killing it
+                // so it doesn't end up in an undead state
+                oldNode.createObservableInstanceIfNeeded()
```
xaviergonz: here's the key, since the dehancer will be called again by mobx for this value (which will call .value, that will call createObservableInstance) we have to make sure it gets called BEFORE death to follow a proper lifecycle

mweststrate:
I think the most beautiful solution would be to return the snapshots. This nicely mirrors the idea to add elements by using snapshots as well, and i don't think the performance is an issue, since snapshots are normally always available for living nodes, and don't need to be actively computed. 

The important downside is the breaking change indeed, for that reason I think your last solution is great as well; kill the nodes. However, that is probably still breaking as well. 

Doublecheck: Would this still work as expected without creating new Todo's?:

```javascript
.actions(self => ({
  clearFinishedTodos() {
     self.todos = self.todos.filter(todo => !todo.done)
  }
})
```

xaviergonz:
@mweststrate yep, still works, just added a unit test for it

I also think returning snapshots is the best solution, but like you said it is a breaking change.
And if destroying the nodes is a breaking change or a fix, well, might be debatable... (I think the original intent is that when you pop a node from an array it should die?)

xaviergonz:
```ts
test("assigning filtered instances works", () => {
    const Task = types.model("Task", {
        done: false
    })
    const store = types
        .model({
            todos: types.array(Task)
        })
        .actions(self => ({
            clearFinishedTodos() {
                self.todos = cast(self.todos.filter(todo => !todo.done))
            }
        }))
        .create({
            todos: [{ done: true }, { done: false }, { done: true }]
        })

    expect(store.todos.length).toBe(3)
    const done = store.todos.filter(t => t.done)
    const notDone = store.todos.filter(t => t.done)
    expect(store.todos.every(t => isAlive(t)))
    store.clearFinishedTodos()
    expect(store.todos.length).toBe(1)
    expect(done.every(t => !isAlive(t)))
    expect(notDone.every(t => isAlive(t)))
})
```

mweststrate:
Ok, let's kill the nodes for now. I think it can be a patch version; if people would rely currently on the results being still alive, then they would have run into issues downstream when using those objects, because paths might have become invalid etc etc

xaviergonz:
Cool! Just created an issue for possible breaking changes in a next major so we don't forget all this

PR #1219, opened 2019-03-14
fix volatile docs

xaviergonz:


urugator:
Hm... I think that with the provided fix there is no "volatile state". 
"State" means something mutable, "volatile" means non-observable, right?
The idea is to make clear that user must not access volatile state from within a computed ...  so the  `get x()` (getter) should be just changed to `getX()` (plain method)

xaviergonz:
Hmm, to me volatile means something that is "not persisted" (or at least that's what I infer from the "volatile" section that MST has, where state is not persisted, yet it is public and observable. But I might be mistaken...

Maybe it is better just to present an example of the two variants?

cc @mweststrate

xaviergonz:
@urugator 
created a new PR: https://github.com/mobxjs/mobx-state-tree/pull/1220

urugator:
No, you're right. I am going thought docs now and I've initially misunderstood the prupose of "volatile" ... As you pointed out, the idea actually is to make sure that `volatile` is `observable` ...

urugator:
Actually... looking at the example above with  `const pendingRequest = null // a Promise` (btw there is a mistake, it can't be `const`) ... it's not that trivial to come up with general definition/recommendation ...  
Or perhaps the `pendingRequest` should be wrapped into `observable` as well? Just in case the user will try to access it anywhere from `computed`?
To me it seems that using non-observable variables (non-constants) should be completely forbidden...

xaviergonz:
Fixed, the const promise thing

I think the rule could be that if it is going to end up on a view getter (computed) it should be observable, else up to you. I added the two approaches on the pr https://github.com/mobxjs/mobx-state-tree/pull/1220

could you comment there what you think?

PR #1220, opened 2019-03-14
Fix volatile docs (part 2)

xaviergonz:


README.md
```diff
@@ -862,13 +862,38 @@ Some tips:
 4.  It is possible to share volatile state between views and actions by using `extend`. `.extend` works like a combination of `.actions` and `.views` and should return an object with a `actions` and `views` field:
 
 ```javascript
+// if your local state is NOT part of a view getter (computed value)
```
urugator: To me it seems too error prone to suggest it as viable approach ("usually good enough").
If something is mutable, it's state and state should be observable, otherwise you are risking staleness.
The variable can be eventually accessed in some derivation, perhaps transitively, it can be hard to make assumptions about that during variable declaration.
I am not a mobx-state-tree user, but I think, if I would know, there is a way to shoot myself to leg, I would try to minimize a chance of doing so as much as possible.
*"Make everything observable"* seems easier to follow, then *"this doesn't have to be observable, but you have to make sure that [some conditions]"*
I think it's good that it's mentioned in docs, I would just be hesitant advertising it as common pattern.
But as I say I don't use mobx-state-tree, possibly it's just some edge case scenario and it's fine most of the time...

xaviergonz: sounds good, I reworded it a bit, how does it look now?

PR #1240, opened 2019-03-31
add mstRunInAction

xaviergonz:
- Added `mstRunInAction`, similar to mobx's `runInAction` but for MST node instances.
- Removed `castFlowReturn` since when a flow returns a promise what is actually returned from the flow is the resolved value. (this makes mobx and mst flow defs similar)

`mstRunInAction(node, name?, thunk)`

Usually used to do async/await actions without the need for superfluous actions. Not as good as flow, but has better type inference.

That being said I'm really looking forward [Strongly typed iterators and generators](https://github.com/Microsoft/TypeScript/issues/2983) in the roadmap for TS 3.5 to finally fix the typings for flow

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,12 +465,7 @@ someModel.actions(self => {
 })
 ```
 
-Note that, since MST v3.9, TypeScript correctly infers `flow` arguments and usually infers correctly `flow` return types,
-but one exception to this case is when a `Promise` is returned as final value. In this case (and only in this case) this construct needs to be used:
-
-```ts
-return castFlowReturn(somePromise)
-```
+Note that, since MST v3.9, TypeScript correctly infers `flow` arguments and usually infers correctly `flow` final return types, however intermediate yielsd will have `any` as type.
 
 #### Action listeners versus middleware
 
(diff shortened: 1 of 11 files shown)


xaviergonz:
To be honest if flows were to be properly typed in TS (or if async/await were to work ok) I'd not have much of a use case. I usually skip flows in the original mobx and use runInAction after awaits exclusively for this reason (yield always returning any).

Another alternative is using anonymous actions, with private actions that are not exposed, but that's not possible unless there's some way in models to create those (which mstRunInAction would allow).

Maybe a better way to do this is to rename it to mstAction and make an example like this?
```ts
actions(self => {
  const privateAction = mstAction(self, (x: number) => { self.x = x })

  return {
    async asyncAction() {
      const x = await something()
      privateAction(x)
    }
  }
})
```

jamonholmgren:
@xaviergonz Any update on this? I'm taking over MST as primary maintainer, just checking to see if you still want to pursue this.

xaviergonz:
Sorry, feel free to close it

PR #1245, opened 2019-04-01
safeReference with acceptsUndefined: false to use inside collections

### LINKED ISSUE #1236: [Typescript] types.array and types.map of (types.safeReference(Model)) should not believe it can take undefined values
**_Question_**

* [x] I tried the [spectrum channel](https://spectrum.chat/mobx-state-tree/general/usage-of-types-array-types-safereference-model~3ee219c7-c582-4c40-892d-a162e3f0d693)

As discussed in the Spectrum chat link above, I need to run the property that is created with `types.array(types.safeReference(Model))` through a filter function to remove undefined instances. Otherwise, this "view" into the data isn't compatible with something of type `Model[]`.

It's my understanding from the documentation that instance of `Model` will be removed from `types.array` if they're managed with `types.safeReference`. So I'm not sure why this filter function is necessary...

If I were to clone the list (i.e. `fooList = [...modelCollection.theArray]`) then certainly `fooList` could contain elements that are undefined (the referenced object could no longer exist). Is this why I need to filter it? Is this a bug with the typing of the safe reference MST array?

xaviergonz:
Fixes #1236 

Add a safeReferenceForCollection, which is like safeReference, but without maybe, so it is better suited for collections (arrays and maps), since usually don't want those to take undefined as a valid value.

In retrospective it would have been better to make safeReference not take maybe and tell the users to wrap it in maybe when it was being used as a model property, but that'd be now a breaking change.

Such change could be considered for a next major version though?


xaviergonz:
> I'm wondering whether safeReference(Type, { acceptUndefined: false }) would be clearer to achieve the same?

sounds good, although the docs will have to state that clearly. i'll try that

xaviergonz:
@mweststrate done

mweststrate:
Released as 3.14.1

Issue #1208, opened 2019-03-11
Using types.optional(Model, {}) with typescript will throw an error


bdwenxi:
**_Question_**

* [] I've checked documentation and searched for existing issues
* [] I tried the [spectrum channel](https://spectrum.chat/?t=dad48299-3dfc-4e10-b6da-9af1e39498a3)

I have a parent store. This parent store contains two child stores. I use types.optional(Model, {}) to set two child stores. Typescript will throw an error, as shown below.
![image](https://user-images.githubusercontent.com/33823489/54104557-4eddd280-440b-11e9-9741-eae5ea2d8a58.png)
![image](https://user-images.githubusercontent.com/33823489/54104577-5f8e4880-440b-11e9-8479-c0f62921fb94.png)

**Version**
mobx: "4.9.2"
mobx-react: "5.4.3"
mobx-state-tree: "3.11.0"
typescript: "3.3.3333"

I use "cast" method or "castToSnapshot" method has no effect, the same will throw an  #error.

ksjogo:
{} isn't a CustomSkillsModel, so you need a constructor for CustomSkillsModel there (or have actual static one. The solution you are likely looking for is `() => CustomSkillsModel.create({})`, but I don't know the rest of your codebase.

simmons8616:
@ksjogo 
The same problem, as shown below
![image](https://user-images.githubusercontent.com/33823489/54106048-ea714200-440f-11e9-98c9-0d322ed40926.png)

The following picture is the details of the child store
![image](https://user-images.githubusercontent.com/33823489/54106212-67042080-4410-11e9-96c6-f5651338f662.png)

xaviergonz:
and what is PagerModel

simmons8616:
The following picture is the PagerModel
![image](https://user-images.githubusercontent.com/33823489/54112348-ebf73600-4420-11e9-8770-feda8ad24870.png)

xaviergonz:
![image](https://user-images.githubusercontent.com/6306796/54159286-f6c7c000-444c-11e9-823f-e9f036fee417.png)

no errors for me
do you have all strict checks enabled in your tsconfig.json?

xaviergonz:
This worked for me:
```ts
test("", () => {
    const CustomSkillsModel = types.compose(
        types.model({
            pageNo: 1,
            pageSize: 15,
            totalCount: 0,            
        }), 
        types.model({
            skills: types.array(types.null)
        })
    )
[...]

PR #1251, opened 2019-04-06
Fix: Type instantiation is excessively deep and possibly infinite. TS2589 (TS3.4)

### LINKED ISSUE #1247: TypeScript error: Type instantiation is excessively deep and possibly infinite. TS2589
**_Bug report_**

* [x]  I've made sure my project is based on the latest MST version

https://github.com/benkuly/mst-issue-ts2589

I defined the mst-model in a seperate/shared project (in the example it is named "common").
Now I want to use it in another project (in the example it is named "client").

This already works, if I define the mst-model in the same project.

The "common" project compiles without errors, but I get `TypeScript error: Type instantiation is excessively deep and possibly infinite. TS2589` when I want to compile the "client" project.
It seems to be related with the generated declaration files in the "common" project

Thanks for your help!

xaviergonz:
Fixes #1247 

Due to a new check for type nesting in TS3.4 (capped at 50 iterations) that only applies to .d.ts files (AFAIK), generated type definitions were giving this error with relatively small models.
This PR reduces the type nesting (from around 51 from the provided test case to around 12) by refactoring a bit the type system.

The only con is that the smarter "create" function, which was able to magically infer if create can or cannot take a snapshot has to be gone for this change to work. (Though anyway that feature seems to give some users some problems, see #1208 )

The good thing is that since the type nesting is greatly reduced IDE auto-completion should be even faster now.

I also checked type definition generation, now they are slightly smaller.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,4 @@
-- Reduced type nesting to avoid Typescript 3.4 errors about infinite types. Sadly due to this change `types.create` is no longer smart enough in TS to know skipping the create parameter is valid or not.
+- Reduced type nesting to avoid Typescript 3.4 errors about infinite types. Sadly due to this change `types.create` is no longer smart enough in TS to know skipping the create parameter is valid or not. Through [#1251](https://github.com/mobxjs/mobx-state-tree/pull/1251) by [@xaviergonz](https://github.com/xaviergonz).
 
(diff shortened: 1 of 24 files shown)


mweststrate:
Great stuff @xaviergonz !

PR #1269, opened 2019-04-29
Fix for empty models / models with all properties set to optional being able to take any value in TypeScript

### LINKED ISSUE #1268: Type mismatch when using “types.reference” with Typescript
**_Bug report_**

https://codesandbox.io/s/09q10597w

**Minimal reproduction code:**
```js
import { types } from "mobx-state-tree";

export const Book = types.model({
  id: types.identifier
});

export const BooksStore = types.model({
  // Exception only if uses "types.reference"
  books: types.array(types.reference(Book))
});

export const RootStore = types.model({
  booksStore: BooksStore
});

const booksStore = BooksStore.create({ books: [] });

// TS Exception here
const rootStore = RootStore.create({ booksStore });
```

Looks like a closed issue: https://github.com/mobxjs/mobx-state-tree/issues/994

xaviergonz:
Apparently, in TS this is valid
```
const x: {} = 5
const x: {} = [5]
```
which means that for empty objects / objects where all properties were optional any value was valid

This PR fixes this by turning empty objects into an object like 
```
{
  [some unique symbol]?: undefined
}
```
which doesn't have any properties in common with array, classes, etc.

Also updates some dev deps and fixes some typos


packages/mobx-state-tree/src/types/complex-types/model.ts
```diff
@@ -135,11 +135,25 @@ type DefinablePropsNames<T> = {
...
+export interface EmptyObject {
+    [$emptyObject]?: undefined
+}
+
 /** @hidden */
 export type ExtractCFromProps<P extends ModelProperties> = { [k in keyof P]: P[k]["CreationType"] }
 
 /** @hidden */
-export type ModelCreationType<PC> = { [P in DefinablePropsNames<PC>]: PC[P] } & Partial<PC>
+export type KeylessToEmptyObject<O> = keyof O extends never ? EmptyObject : O
```
mweststrate: `keyof O extends never` :100:

xaviergonz:
While it works I'm going to mark it as WIP since merging this would make the 50 iterations error from TS3.4 appear again :/

Hopefully this new feature from TS (https://github.com/Microsoft/TypeScript/pull/30637) means what I think it means and then it might even resolve by itself.

Anyway it only exhibits in two cases
1) when the model has no props the creation/snapshot/instance types accept anything
2) when all model props are optional then creation accepts arrays and numbers, but not wrong objects, while the other types seem ok

xaviergonz:
Ok, i just checked with the next TSC and it doesn't appear to mean what I think :-/ (unless it wasn't still there in the nightly build)

mweststrate:
If it is troublesome with TS 3.4, I would wait a bit before merging this

xaviergonz:
managed to reduce the number of iterations to 46, but still too high I think

mweststrate:
> Maybe it would be better to change types.model typings so it cannot take an empty object as props?

Sounds as a good idea anyway

> The problem would still be there when all props are optional though

That is actually quite a common case I presume :cry:

mweststrate:
I'm not sure if the issue is big enough to warrant the additional complexity in the typings. Ok to close this one?

xaviergonz:
I'd like to try some ideas I had to reduce the number of types nesting before giving up on this one.

mweststrate:
np!

FredyC:
I still haven't got to proper upgrading to 3.5 due to #1341, but I am expecting this is gonna bite me too. I do have a couple of models with all props optional. I wonder what's wrong with the PR? Does it work with that? Doesn't seem to be overly complicated, so I am not sure why to abandon it.

xaviergonz:
The problem has to do with a limit (added in TS 3.4 I think?) on how deep conditional types can go, which is 50 levels deep. Some medium-big projects where reaching that limit and that got fixed, but this PR adds some more levels and makes it go over 50 again, thus making these medium-big projects un-compilable.

I was just thinking of trying to leverage interfaces as "cached" types to reduce nesting, something like
```
interface A<X> {
  $whatever: X extends Foo ? Bar : Whatever
}

types Ble<X> = A<X>["$whatever"]
```

rather than directly
```
types Ble<X> = X extends Foo ? Bar : Whatever
```

something like that (although it is an oversimplification), but I still have to check if that actually reduces nesting

xaviergonz:
@mweststrate  Ok I found a better way to fix it and not increase the depth levels at all, basically by always '&' a non empty object like { [someSymbol]?: any }

Also while at it updated the dev deps, which includes a nice update to the typedocs markdown plugin, which should result in a bit better looking autogenerated docs

FredyC:
@xaviergonz We got some approvals here. I suppose you can resolve conflicts and merge? :)

xaviergonz:
Should be fixed.
While at it also update deps and fixed flow typings for TS3.6, but this means that now flows require TS3.6.

xaviergonz:
@mweststrate since the fix for flow typings require ts3.6, is it ok to merge?

mweststrate:
I think it is OK if it is mentioned in the changelog, and at least a minor release

Issue #1271, opened 2019-04-29
createActionTrackingMiddleware breaks with nested flows


davidyxu:
**_Bug report_**

* [ ] I've checked documentation and searched for existing issues
* [ ] I've made sure my project is based on the latest MST version
* [ ] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

**Sandbox link or minimal reproduction code**
https://codesandbox.io/s/6w5p1qwr4k

**Describe the expected behavior**
There should be no issue with using middleware from createActionTrackingMiddleware with nested flows

**Describe the observed behavior**
There is always an error due to the middleware usage for nested flow actions:
![image](https://user-images.githubusercontent.com/1980008/56928845-e5825380-6a8c-11e9-809f-5012bbcd5deb.png)

I believe this issue is from the fact that the action middleware will remove the rootId on delete (which will remove the action context that the outside flow still depends on). Instead the action middleware should delete the `call.parentId`

https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/middlewares/create-action-tracking-middleware.ts#L86

xaviergonz:
Did you consider trying `https://github.com/mobxjs/mobx-state-tree/blob/master/docs/API/README.md#createactiontrackingmiddleware2` ?

davidyxu:
I took a brief look at it but it lacks the onSuspend and onResume callbacks which are convenient for tracking async actions.

xaviergonz:
Just wondering, what does your middleware do? Maybe there's some way to write it with the newone (or add something to the new one so it supports it).

benjamind:
@davidyxu is away on PTO, but I can help answer this one.

The middleware is a transaction tracking implementation. It allows us to build transactions that record model property changes over an action (include async flow actions), building up a collection of snapshots of model changes which can then be all 'commited' at once to a backend. So it makes use of `onResume` to swap out the transactions that are being written to as asynchronous actions are paused/resumed.

We've tested out the suggested fix, which is to change the tracking middleware v1 to only delete the `rootId` if the `rootId` and the `parentId` of the call are the same. Seems to work. But if middleware v1 is going to be completely deprecated then I think middleware v2 needs to support a resume hook similar to the existing middleware for our use-case.

xaviergonz:
That sounds doable, since it is pretty close to what atomic does.

[...]

PR #1285, opened 2019-05-14
fix nested flows with middleware

davidyxu:
For the issue mentioned in https://github.com/mobxjs/mobx-state-tree/issues/1271

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/create-action-tracking-middleware.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/create-action-tracking-middleware.test.ts
@@ -1,0 +1,36 @@
+import { addMiddleware, createActionTrackingMiddleware, types, flow } from "../../src"
+
+const middleware = createActionTrackingMiddleware({
+    onStart: () => {},
+    onResume: () => {},
+    onSuspend: () => {},
+    onSuccess: () => {},
+    onFail: () => {}
+})
+
+const M = types
+    .model({})
+    .actions(self => ({
+        innerFlow: flow(function*() {
+            yield new Promise(resolve => setTimeout(resolve))
+            return true
+        })
+    }))
+    .actions(self => ({
+        outerFlow: flow(function*() {
+            yield self.innerFlow()
+            return true
+        })
+    }))
+
+test("can handle inner flow", async () => {
+    const m = M.create()
+    addMiddleware(m, middleware)
+    await expect(m.innerFlow()).resolves.toBe(true)
+})
+
+test("can handle outer flow", async () => {
+    const m = M.create()
+    addMiddleware(m, middleware)
+    await expect(m.outerFlow()).resolves.toBe(true)
+})
--- a/packages/mobx-state-tree/src/middlewares/create-action-tracking-middleware.ts
+++ b/packages/mobx-state-tree/src/middlewares/create-action-tracking-middleware.ts
@@ -77,13 +77,17 @@ export function createActionTrackingMiddleware<T = any>(
             }
             case "flow_throw": {
                 const root = runningActions.get(call.rootId)!
-                runningActions.delete(call.rootId)
+                if (call.rootId === call.parentId) {
+                    runningActions.delete(call.rootId)
+                }
                 hooks.onFail(call, root.context, call.args[0])
                 return next(call)
             }
             case "flow_return": {
                 const root = runningActions.get(call.rootId)!
-                runningActions.delete(call.rootId)
+                if (call.rootId === call.parentId) {
+                    runningActions.delete(call.rootId)
+                }
                 hooks.onSuccess(call, root.context, call.args[0])
                 return next(call)
             }


packages/mobx-state-tree/__tests__/core/create-action-tracking-middleware.test.ts
```diff
@@ -0,0 +1,36 @@
...
+            yield self.innerFlow()
+            return true
+        })
+    }))
+
+test("can handle inner flow", async () => {
+    const m = M.create()
+    addMiddleware(m, middleware)
+    await expect(m.innerFlow()).resolves.toBe(true)
+})
```
xaviergonz: could you also check that the actionTrackingMiddlewares are properly called in the right order?
Something like having a string array and then each action tracking middleware hook pushes something to that event and doing a expect(events).toMatchInlineSnapshot

same for the other test

davidyxu: Working through the suggested test changes and I think it's uncovering some more issues in the middleware from nested flow.

Just to make sure we're on the same page on expectations, given actions that look like the following:
```
innerFlow: flow(function *() {
  console.log("innerFlow - run 1")
  yield new Promise(resolve => setTimeout(resolve))
  console.log("innerFlow - run 1")
})
outerFlow: flow(function *() {
  console.log("outerFlow - run 1")
  yield self. innerFlow();
  console.log("outerFlow - run 2")
});
```

I would expect to see:
```
outerFlow - onStart
outerFlow - onResume
outerFlow - run 1
innerFlow - onStart
innerFlow - onResume
innerFlow - run 1
innerFlow - onSuspend
outerFlow - onSuspend
outerFlow - onResume
innerFlow - onResume
innerFlow - run 2
innerFlow - onSuspend
innerFlow - onSuccess
outerFlow - onSuspend
outerFlow - onResume
outerFlow - run 2
outerFlow - onSuspend
outerFlow - onSuccess
```

If the plan is to deprecate the old middleware API, would you be open to changing createActionTrackingMiddleware2 interface as follows:
```
export interface IActionTrackingMiddleware2Hooks<TEnv> {
    filter?: (call: IActionTrackingMiddleware2Call<TEnv>) => boolean
    onStart: (call: IActionTrackingMiddleware2Call<TEnv>) => void
    onSuspend?: (call: IActionTrackingMiddleware2Call<TEnv>) => void
    onResume?: (call: IActionTrackingMiddleware2Call<TEnv>) => void
    onFinish: (call: IActionTrackingMiddleware2Call<TEnv>, error?: any) => void
}
```
such that onSuspend and onResume are only necessary if a user needs those hooks (but doesn't complicate the usage for the simple cases)?

xaviergonz: Now you see why the second one was created :)

At first I was trying to fix the old one, but the old one was never unit tested and it had some kind of "weird" patterns that I didn't dare to fix in order not to break backwards compatibility, therefore the second one was born

About the new hooks, as long as resume and suspend also gets called properly for sync actions (basically resume right after start and suspend right before finish) I don't see why not :)

davidyxu: :) Iterating on the new api is definitely feeling more appealing now, I'll take a stab at it sometime this week.

davidyxu:
@xaviergonz 

I started looking getting the onResume and onSuspend hooks in on this fork:
https://github.com/davidyxu/mobx-state-tree/commit/cd1b71c2d9308ac37c1bf6c7f20adad5678b1058

But I'm still seeing a similar issue as with the old API, where the `outerFlow` hooks aren't being called again when the `innerFlow` resumes (which I would expect to happen because in this case we're calling innerFlow within the context of the outerFlow, so the middleware on innerFlow and outerFlow should both be used firing).
```
    outerFlow [ 'outerFlow (27) - onStart',                                                                                                                  
      'outerFlow (27) - onResume',                                                                                                                           
      'outerFlow - run 1',                                                                                                                                   
      'innerFlow (29) - onStart',                                                                                                                            
      'innerFlow (29) - onResume',
      'innerFlow - run 1',
      'innerFlow (29) - onSuspend',
      'outerFlow (27) - onSuspend',
      'innerFlow (29) - onResume',
      'innerFlow - run 2',
      'innerFlow (29) - onSuspend',
      'innerFlow (29) - onFinish (error: false)',
      'outerFlow (27) - onResume',
      'outerFlow - run 2',
      'outerFlow (27) - onSuspend',
      'outerFlow (27) - onFinish (error: false)' ]
```
For this particular case I could fix it by basically iterating over the call chain and digging up runningActions for parents and calling their onResume/onSuspend in the right order.

However, I don't think this approach would work in more complicated cases because to my understanding, the weak map would only contain running actions for the same middleware. If for example outerFlow was decorated with a different middleware (but innerFlow wasn't), then there isn't a way with the data model to correctly fire off the outerFlow's onResume hook.

It's a pretty significant change, but I think solution would be to have a global map of call -> array of hooks.

xaviergonz:
Can you share your unit test code? I might give it a try

davidyxu:
It should be in the link that I shared: https://github.com/davidyxu/mobx-state-tree/commit/cd1b71c2d9308ac37c1bf6c7f20adad5678b1058#diff-c64a56beaa47a06dd719c54b249a5823

Currently it's just doing console.logs instead of actual expectations but should be pretty straight forward to fix that

davidyxu:
@xaviergonz 
Just wanted to check in to see if you had any luck on your end investigating this. Happy to help setup a few more unit tests or try out any alternative approaches.

mweststrate:
@xaviergonz @davidyxu I notice that this one is open for a really long time already :). Any suggestions on how to proceed?

davidyxu:
@mweststrate 
This PR will fix the exception in the linked issue using the original middleware API. I also have a branch which extends the middleware 2 API with onResume/onSuspend flags.

Unfortunately the way in which middlewares fire when combined with flows is a little unexpected, and from my limited understanding of the MST codebase, it would require some refactoring in the middleware system to get into a more .

If other people are running into the same issue with flows and middlewares mentioned below, I think it makes sense to merge this PR:
https://github.com/mobxjs/mobx-state-tree/issues/1271

The issues around when middleware hooks are called during flows could be tracked with a new giithub issue.

mweststrate:
@davidyxu so if I understand correctly from the issue, you were able to fix the problem with then existing middleware api's, by doing some more work, correct?

xaviergonz:
I tried to make it work without tweaking many things but failed. 

To make this properly work we'd need a refactor to the action context internals so that action each contexts include their root context, the async action that spawned the action context and so on. I did it for mobx-keystone, so the idea is sound, but it is no small feat.

Relevant files if interested:
https://github.com/xaviergonz/mobx-keystone/blob/master/packages/lib/src/action/context.ts
https://github.com/xaviergonz/mobx-keystone/blob/master/packages/lib/src/action/modelFlow.ts
https://github.com/xaviergonz/mobx-keystone/blob/master/packages/lib/src/action/wrapInAction.ts
https://github.com/xaviergonz/mobx-keystone/blob/master/packages/lib/src/actionMiddlewares/actionTrackingMiddleware.ts

davidyxu:
@mweststrate  yes, the immediate problem of the exception is fixed in this PR, but @xaviergonz's suggestion of checking the order in which middleware hooks are called revealed that there's deeper issues to address which requires a larger refactor.

jamonholmgren:
Hey folks, I'm taking over MST as primary maintainer. Could I get an update on where this PR is at and whether you think it'll move forward or if we should close it for now? 🤗

jamonholmgren:
Gentle ping @davidyxu  ^^

davidyxu:
I haven't touched MST in quite a while, but as far as I remember, the PR does fix the uncaught errors mentioned in the linked issue. It however does not fix the underlying issues around the ordering of multiple middlewares.

Those underlying issues look fixable but would involve a much larger refactor than I'm comfortable committing to.  

TLDR: This PR is a small improvement to the original issue, but a larger change is necessary to fix everything discussed here. Please feel free to close it if you'd prefer to wait for a more comprehensive fix.

jesse-savary:
@jamonholmgren I feel that fixing a bug is worth it even though certain MST internals **do need** to be reworked to solve greater issues. Thoughts?

jamonholmgren:
> @jamonholmgren I feel that fixing a bug is worth it even though certain MST internals **do need** to be reworked to solve greater issues. Thoughts?

@jesse-savary I'd be willing to review a new PR that fixes those greater issues; I won't be pushing this forward myself, though, as it's still a fairly unusual edge case to run into any of these. Are you looking for something to work on? 😅

PR #1289, opened 2019-05-17
update mst-middlewares addMiddleware examples

gsouquet:
Hi,

Thank you so much for sharing `mobx-state-tree`. When I was going through the documentation today I believe I encountered two typos in the `README.md` for the `mst-middlewares` package.

Here's a PR to fix that


xaviergonz:
nice catch again :)

PR #1290, opened 2019-05-17
change postCreate to afterCreate

gsouquet:
Reading through the documentation, I could not find any reference to `postCreate` anywhere. I feel like it was meant to be `afterCreate`.
Correct me if I am wrong


xaviergonz:
nice catch, thanks!

PR #1292, opened 2019-05-18
Add size-limit

pustovalov:
Size Limit is a linter for your JS application or library performance. It calculates the real cost of your JS for end users and throws an error if the cost exceeds the limit.

Example:
```
yarn run v1.15.2
$ size-limit

  packages/mobx-state-tree/dist/mobx-state-tree.min.js
  Size limit:   25 KB
  Package size: 20.24 KB minified and gzipped
  Loading time: 405 ms   on slow 3G

  packages/mst-middlewares/dist/mst-middlewares.min.js
  Size limit:   5 KB
  Package size: 3.68 KB minified and gzipped
  Loading time: 74 ms   on slow 3G

Done in 0.74s.
```

https://circleci.com/gh/pustovalov/mobx-state-tree/17

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,6 +108,14 @@ jobs:
             - store_test_results:
                   path: ./test-results
 
+    test-size:
+        executor: my-executor
+        steps:
+            - attach_workspace:
+                  at: .
+
+            - run: yarn size
+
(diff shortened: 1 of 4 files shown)


xaviergonz:
looking good, thanks!

PR #1296, opened 2019-05-22
Fix browser event click that calls mobx action

agrublev:
```
Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'FiberNode'
    |     property 'stateNode' -> object with constructor 'HTMLSpanElement'
    --- property '__reactInternalInstance$qcz9vcggugc' closes the circle
    at JSON.stringify (<anonymous>)
    at eval (mst-middlewares.module.js:182)
    at Array.map (<anonymous>)
    at getActionContextNameAndTypePath (mst-middlewares.module.js:182)
    at logStep (mst-middlewares.module.js:355)
    at actionMiddleware (mst-middlewares.module.js:387)
    at runNextMiddleware (mobx-state-tree.module.js:2669)
    at runMiddleWares (mobx-state-tree.module.js:2682)
    at runWithActionContext (mobx-state-tree.module.js:2501)
    at res (mobx-state-tree.module.js:2528)
```

Happens when you do this
```
<Button
	size={"large"}
	onClick={toggleAddEditConversation}
	className={"start-conversation-btn"}
>
	Start Conversation
</Button>
```
Instead of
```
<Button
	size={"large"}
	onClick={() => {
		toggleAddEditConversation();
	}}
	className={"start-conversation-btn"}
>
	Start Conversation
</Button>
```
Which is silly

### DIFF
--- a/packages/mst-middlewares/src/redux.ts
+++ b/packages/mst-middlewares/src/redux.ts
@@ -80,7 +80,13 @@ function getActionContextNameAndTypePath(actionContext: ActionContext, logArgsNe
     let targetTypePath = actionContext.targetTypePath
 
     if (logArgsNearName) {
-        let args = actionContext.callArgs.map(a => JSON.stringify(a)).join(", ")
+        let args = actionContext.callArgs.map(a => {
+            try {
+                return JSON.stringify(a);
+            } catch (e) {
+                // Seems it may be a browser event?
+            }
+        }).join(", ")
 
         if (args.length > 64) {
             args = args.slice(0, 64) + "..."


packages/mst-middlewares/src/redux.ts
```diff
@@ -80,7 +80,13 @@ function getActionContextNameAndTypePath(actionContext: ActionContext, logArgsNe
     let targetTypePath = actionContext.targetTypePath
 
     if (logArgsNearName) {
-        let args = actionContext.callArgs.map(a => JSON.stringify(a)).join(", ")
+        let args = actionContext.callArgs.map(a => {
+            try {
+                return JSON.stringify(a);
+            } catch (e) {
+                // Seems it may be a browser event?
```
xaviergonz: could you return here something like `**UNSERIALIZABLE**` so it doesn't leave an empty space?

xaviergonz: That should also fix the build error

dnizetic:
Any news here?

When combined with hot reloading, if I have a self referencing model:

> parentItem: late(() => safeReference(TodoModel)),

The hot reloading will fail with 

Uncaught TypeError: Converting circular structure to JSON

Screen Shot 2019-10-18 at 7 16 17 PM

mweststrate:
@dnizetic that looks pretty unrelated, please open a new issue. Same sympton !== same cause

mweststrate:
@agrublev, as silly as it sounds, it is technically correct. As `() => x()` is simply not the same as `x` when both of these are passed as a callback. Although your solution supresses the symptoms, it would also suppress all true negatives, so I think this is not the way to go. (Generally speaking empty catches should always raise some alarms ;-))

alex-shamshurin:
I have the same issue with `reactotron-mst`

PR #1328, opened 2019-07-02
Update getting-started.md

lauri-kaariainen:
Change old code in getting-started.md to reflect new behavior of types.maybe

NOTE:
The "View sample in the playground" codesandbox url/code after the part should be also be updated to reflect the change in docs, but I can't do it. Only the last box needs the change. Correctly changed frozen codesandbox link is here but with me as author, if it's of any use: https://codesandbox.io/s/mobx-state-tree-getting-started-13-ron8y

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -588,8 +588,8 @@ const Todo = types
         },
         setUser(user) {
             if (user === "") {
-                // When selected value is empty, set as null
-                self.user = null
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks!

mweststrate:
cc @xaviergonz  for the sandbox update

xaviergonz:
sandboxes updated

Issue #1304, opened 2019-05-30
Production Mode MST Error message is too vague.


OverseePublic:
**_Feature request_**

**Is your feature request related to a problem? Please describe.**
Currently, if I try to do a `map.put` operation on a complex model with a different type, I get the following error. When this error happens in the production, the relevant stack information is removed in the error object that was originally thrown, making debugging extremely difficult in a very complex system (us). We would like more relevant information to be shown so we can debug the issue more easily. (We want to know the stack of root cause in our code and more context in regards to how the MST failed).

Because our production users get message below and they have no idea as what happened, they just send us information like this to the developers, and it is difficult for us to tell what is wrong unless we run our local instance and reproduce the issue with exact steps.
![image](https://user-images.githubusercontent.com/20207137/58664369-80748480-82fc-11e9-840e-2033f7e69c43.png)

Note that both message and stack does not provide much relevant information.
![image](https://user-images.githubusercontent.com/20207137/58664488-c0d40280-82fc-11e9-8a89-623def56c305.png)

I was able to get the root stack where the MST error was caused (our code) in chrome dev tool. I am unsure as why that information is omitted in the stack.

Some others are also experiencing this issue: https://medium.com/@jiyinyiyong/some-pain-points-using-mobx-state-tree-in-our-team-f4e38890d66

**Describe the solution you'd like**
More relevant error message and stack trace.

**Describe alternatives you've considered**
Importing typecheck directly and use it every operation (Very cumbersome)
Setting NODE_ENV to something other than "production", but this is a hack.

**Are you willing to (attempt) a PR?**
* [ x ] Yes
* [ ] No

xaviergonz:
Which error reporting system are you using? Maybe there's some config to allow them to show the full stack trace rather than just the beginning? The whole stack trace should be there (like you saw in chrome dev tools), it is not like MST is trimming it on purpose :)

That being said, feel free to submit a PR if you think the error reporting can be improved

OverseePublic:
One another question. Why does this operation fail? When do we allow model creation without proper type checking and when do we force fail on prod mode? Should we fail but let the application proceed instead of hard stopping? Basically, i see a convuluted typechecking with fail stop. What is the reason behind this design? Any explanations are helpful! Thanks for your time

OverseePublic:
Could @mweststrate perhaps help us understand? I need this to propagated to our very large team. We serve U.S federal level clients and we would like to know the reasoning behind it.

[...]

PR #1332, opened 2019-07-10
Enable typechecking

OverseePublic:
https://github.com/mobxjs/mobx-state-tree/issues/1304


OverseePublic:
@mweststrate so seems like the approach you suggested was not acceptable in MobX repository... how about we go through ENV route and figure out the configuration object later?

https://github.com/mobxjs/mobx-state-tree/pull/1337

xaviergonz:
I'd close this in favor of the other PR, mobx shouldn't have to change to accommodate to one of its dependencies requirements

xaviergonz:
Not to mention that this would effectively kill the optimization done by the minimizer in production mode since now it would not be able to remove the dev mode code (since now that flag would depend on a simple variable) 

IOW, closing in favor of the approach in #1337

mweststrate:
Released as 3.14.1

PR #1337, opened 2019-07-12
Enable Production Development Error Message

OverseePublic:
https://github.com/mobxjs/mobx-state-tree/pull/1332


xaviergonz:
Looks good, though there should be something on the readme explaining this

xaviergonz:
Actually I think it would be better to create a new typeCheckingEnabled() function that uses that code and move only the type checking stuff from using devMode() to that new one and leaving devMode as is, since devMode for example also enables stuff unrelated to type checking such as freezing objects and the like

OverseePublic:
> Actually I think it would be better to create a new typeCheckingEnabled() function that uses that code and move only the type checking stuff from using devMode() to that new one and leaving devMode as is, since devMode for example also enables stuff unrelated to type checking such as freezing objects and the like

I made updates to the PR. Could you take a look? Thanks for the comment.

OverseePublic:
@mweststrate or @xaviergonz could one of you approve and allow me to merge this? Thanks.

xaviergonz:
Will check it later today, but so far looks good to me

OverseePublic:
> Will check it later today, but so far looks good to me

@xaviergonz good afternoon. Have you got a chance to look over? Our team is waiting. Thank you

mweststrate:
Merged

xaviergonz:
Just crossed my mind , will this work on environments where process or process. env is not defined?

mweststrate:
good catch, I'll add a guard

mweststrate:
Released as 3.14.1

Issue #1307, opened 2019-06-04
Issues with Typescript 3.5.1 release


ConneXNL:
Typescript 3.5.1 just got released. After upgrading from typescript@3.5.0-dev.20190512 the following no longer seems to be working:

```
import { types, IAnyType } from "mobx-state-tree";

export const createCustomType = <ICustomType extends IAnyType>({
  CustomType
}: {
  CustomType: ICustomType;
}) => {
  return types
    .model("Example", {
      someProp: types.boolean,
      someType: CustomType
    })
    .views(self => ({
      get isSomePropTrue(): boolean {
        return self.someProp;
      }
    }));
};

```

Here is the complete error:

```
Property 'someProp' does not exist on type 'STNValue<ModelInstanceTypeProps<ModelPropertiesDeclarationToProperties<{ someProp: ISimpleType<boolean>; someType: ICustomType; }>>, IModelType<ModelPropertiesDeclarationToProperties<{ someProp: ISimpleType<boolean>; someType: ICustomType; }>, {}, _NotCustomized, _NotCustomized>>'.
  Property 'someProp' does not exist on type 'Exclude<ModelInstanceTypeProps<ModelPropertiesDeclarationToProperties<{ someProp: ISimpleType<boolean>; someType: ICustomType; }>>, object>'.ts(2339)
```

Are the typings of MST compatible with TS 3.5.1?

When hovering over `self` of `.views(self => ({`. In TS 3.4.3 it was:
```
(parameter) self: ModelInstanceTypeProps<ModelPropertiesDeclarationToProperties<{
    someProp: ISimpleType<boolean>;
    someType: ICustomType;
}>> & IStateTreeNode<IModelType<ModelPropertiesDeclarationToProperties<{
    someProp: ISimpleType<boolean>;
    someType: ICustomType;
}>, {}, _NotCustomized, _NotCustomized>>
```

While in TS 3.5.1 `self` of `.views(self => ({` seems to be:
```
(parameter) self: STNValue<ModelInstanceTypeProps<ModelPropertiesDeclarationToProperties<{
    someProp: ISimpleType<boolean>;
    someType: ICustomType;
}>>, IModelType<ModelPropertiesDeclarationToProperties<{
    someProp: ISimpleType<boolean>;
    someType: ICustomType;
}>, {}, _NotCustomized, _NotCustomized>>
```

[...]

PR #1339, opened 2019-07-13
Updated typescript to 3.5.3, to verify #1307

mweststrate:
Also added docs on recommended compiler flags. 

Also removed the "we recommend TS", I feel people run too often into it's limitation. Personally would still recommend it, but the price is higher than it should be :cry: (most notably on forward references to actions and such, and circular type definitions)


xaviergonz:
LGTM, though in theory these flags are all that's needed:

strict: true
noImplicitReturns: true

since strict true includes:
```
Enabling --strict enables --noImplicitAny, --noImplicitThis, --alwaysStrict, --strictBindCallApply, --strictNullChecks, --strictFunctionTypes and --strictPropertyInitialization
```
and the switch fallthrough one is not really used by the type system

mweststrate:
@xaviergonz tnx, updated and merged

ConneXNL:
@mweststrate 

As far as I can see this does nothing. The test was added actually uncomments the line that makes it fail:

https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/__tests__/core/type-system.test.ts#L1039

PR #1351, opened 2019-07-23
Make umd debug build work without 'process' global

leonalliander:
I noticed this when adding the umd debug (not-minified) version to an app that doesn't have the process global variable. Because of it, mobx-state-tree doesn't instantiate.

### DIFF
--- a/packages/mobx-state-tree/src/utils.ts
+++ b/packages/mobx-state-tree/src/utils.ts
@@ -407,7 +407,7 @@ export function isTypeCheckingEnabled() {
  * @hidden
  */
 export function devMode() {
-    return process.env.NODE_ENV !== "production"
+    return (typeof process === "undefined" || (process && process.env && process.env.NODE_ENV !== "production"));
 }
 
 /**
--- a/packages/mobx-state-tree/src/utils.ts
+++ b/packages/mobx-state-tree/src/utils.ts
@@ -407,7 +407,7 @@ export function isTypeCheckingEnabled() {
  * @hidden
  */
 export function devMode() {
-    return (typeof process === "undefined" || (process && process.env && process.env.NODE_ENV !== "production"));
+    return process.env.NODE_ENV !== "production"
 }
 
 /**
--- a/rollup.base-config.js
+++ b/rollup.base-config.js
@@ -4,7 +4,11 @@ import resolve from "rollup-plugin-node-resolve"
 import { terser } from "rollup-plugin-terser"
 import replace from "rollup-plugin-replace"
 
-const devPlugins = () => [resolve(), filesize()]
+const devPlugins = () => [
+    resolve(),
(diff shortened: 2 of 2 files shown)


packages/mobx-state-tree/src/utils.ts
```diff
@@ -407,7 +407,7 @@ export function isTypeCheckingEnabled() {
  * @hidden
  */
 export function devMode() {
-    return process.env.NODE_ENV !== "production"
+    return (typeof process === "undefined" || (process && process.env && process.env.NODE_ENV !== "production"));
```
mweststrate: Note that this will bring MST into dev mode if process is _not_ defined. I'd expect 

`typeof process !== "undefined"  && process && process.env && process.env.NODE_ENV !== "production"`

k-g-a:
I'd change `function` to `const` (i.e. `IS_DEV_MODE`) as there is no reason to calling it every time. It's not about to change at runtime.

packages/mobx-state-tree/src/utils.ts
leonalliander: Hi Michel,

My idea was that the debug umd build _should_ be dev mode by default. By using your approach, it will result in production by default and the developer using the umd debug build should then (still) explicitly set process.env.NODE_ENV to enable dev mode. I think this would be confusing for developers (like me) that are not using node.js based tooling (and not using React).

However, I did a quick test and it appears that the minified umd version will _also_ be dev mode. That should not be the case. I think that is where your comment is coming from.

I wonder what your thoughts are on this. If you agree with me, perhaps you know a good method of implementing the above behavior. Thanks!

xaviergonz:
There's another option which is setting the rollup replace plugin so it injects NODE_ENV to replacement for umd non minified builds

leonalliander:
> There's another option which is setting the rollup replace plugin so it injects NODE_ENV to replacement for umd non minified builds

Hmm, sounds like a good idea. I changed the PR accordingly. Now the debug build will always have dev mode enabled and the minified build will not. This cannot be altered, but ENABLE_TYPE_CHECK can still be set in a production build. I think that is how we want it, right?

xaviergonz:
Bear in mind that should only apply to umd builds, with that change it applies to all of them

leonalliander:
> Bear in mind that should only apply to umd builds, with that change it applies to all of them

Allright. I changed it so it will only apply to umd builds.

packages/mobx-state-tree/src/utils.ts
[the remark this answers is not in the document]
leonalliander: You're right. I also noted it returned _undefined_ and not a boolean when process is defined and process.env is not. But I reverted this change so this function is not changed in this PR anymore.

xaviergonz:
Merging, thanks!

PR #1352, opened 2019-07-24
Loosen model name check

### LINKED ISSUE #1349: ModelType.name check seems too strict
**_Question_**

* [ ] I tried the [spectrum channel](https://spectrum.chat/?t=dad48299-3dfc-4e10-b6da-9af1e39498a3)

There is a suspicious check for model name at ModelType:
```ts
// TODO: this test still needed?
if (!/^\w[\w\d_]*$/.test(name))
  throw fail$1("Typename should be a valid identifier: " + name);
```

We're trying to use scoped model names, i.e. ```MyModel@82d92470-d05e-4dab-92cc-25b45f7fa2e4```, which do not pass the check mentioned above, but is valid for most operations in js (can be used as a key for object or as identifier attribute for dom elements). I can't think of any reason for such a strict check.

I propose either loose the check or remove it completely. Can make a PR.

k-g-a:
Fixes #1349. As the question is unanswered I did not dare to remove it completely.


xaviergonz:
did you try if the tests pass when removing the whole limit? I can't think of a reason of why it couldn't just be any kind of string

k-g-a:
Nope. Will give it a shot tomorrow :)

k-g-a:
Well, all the tests pass without model name check, so i've removed it completely

xaviergonz:
in it goes

Issue #1151, opened 2019-01-24
Proposal: types.hooks


xaviergonz:
Right now hooks / snapshot pre/post processing can only be done over models, but what if it could be applied to any kind of type node with something like this:

```
// all of them are optional
const hooks = {
  preProcessSnapshot(sn) {},
  postProcessSnapshot(sn) {},
  afterCreate(self) {}, // not available in scalar nodes
  afterAttach(self) {},
  beforeDetach(self) {}, // not available in scalar nodes
  beforeDestroy(self) {}
}

// option A
const hookedString = types.hooks('optionalName or will inherit the name', types.string, hooks)

// option B
const hookedString = types.string.withHooks('optionalName or will inherit the name', hooks) // can be chained
```

With that in place, doing something like a `types.optionalNull` would be kind of trivial:

```
function optionalNull<IT extends IAnyType>(type: IT, defaultValueOrFunction: OptionalDefaultValueOrFunction<IT>) {
  const nullToUndefined = types.hooks(types.union(types.null, type), {
    preProcessSnapshot: (sn) => sn === null ? undefined : sn,
    postProcessSnapshot: (sn) => sn === undefined ? null : sn
  })

  return types.optional(nullToUndefined(type), defaultValueOrFunction)
```

Also this could coexist with the current way until (if?) the hooks inside model actions / .pre-.postProcessSnapshot get deprecated in a future v4

The only con I can see is that the hooks would no longer be able to access "internal state" on the model, e.g.
```
types.model(...).actions(self => {
  let x = 0;
  return {
    // hooks can access x
  }
})
```

But for that something like
```
setState(node, data)
getState(node, data)
```
could be created

Thoughts?

xaviergonz:
Related issue: https://github.com/mobxjs/mobx-state-tree/issues/947

k-g-a:
Making hooks as a separate wrapper type is a viable idea in general. To my opinion, ```types.hooks``` looks more naturall in MST semantics compared to ```TYPE.withHooks```.
Besides "internal state" flaw mentioned above (which I do not consider as a flaw at all) I have the following questions to discuss:
1) Do we really need to expose hooks on ```ScalarNodes```? Taking into account that two of those hooks are forbidden/won't work (silently or not?) - it's harder to teach/explain. 
[...]

PR #1355, opened 2019-07-25
Array/Map hooks

k-g-a:
I'm looking forward to implement [proposed hooks RFC](https://github.com/mobxjs/mobx-state-tree/issues/1151), but don't have enough free time for that now.

Meanwhile, hooks could be useful for `types.array`/`types.map`.
This PR brings really basic implementation (same API as for model, but is called `.hooks()`):
```ts
const myArray = types.array(MySubType).hooks(self => ({
  afterCreate();
  afterAttach();
  beforeDetach();
  beforeDestory();
}))
```

Primary use case for us is the following:
- we have - let's say - `StoreState` (mst `type.model`), which holds `itemsMap` of particular type by id;
- it's wrapped in a (non-mst) `Store` class which orchistrates fetching data from API, deserializing it (before putting into `StoreState`, as we discourage use of preProcessSnapshot/postProcessSnapshot) etc.
- `Store` has (alongside with others) the `findAll()` method, wich returns _live_ array of items; by _live_ I mean, that if for some reason (be it a subsiquent API request which returns less/more/other items or `Store.addItem(payload)` method being used to create client-side model before saving it to backend)  - returned array should reflect it;
- an arbitraty consumer (i.e. some react component/route/service) may call `findAll()` and use it's result as standalone object or as a part of another model (i.e. component state, viewmodel or something conceptually similar);
- the mechanic described above demands that we have to return `types.array(types.safeReference(ItemType, {get(), set()})).create(Array.from(this.storeState.itemsMap.values()))` - a custom safe reference, which plays nice with modifying/removing items, but lacks support of adding; nor we can not jsut return `this.storeState.itemsMap.values()` as it can't be used as part of another tree, neither we can return snapshot as it's not _live_;
- to handle adding items we have to manually 'register' every instance returned from the `Store.findAll()` and update it by hand; but this leads to memory leak as there is always a reference to returned collection inside the `Store`, even after it's consumer is disposed (unmounted/destroyed etc.);
- for that reason we need `beforeDetach`/`beforeDestroy` hooks for `types.array`; having `afterCreate/afterAttach` would simplify subscription management too.


k-g-a:
If the idea is accepted, I can polish the code, modify corresponding documentation, add more tests

mweststrate:
The impact doesn't seem to be too big, and api consistent. LGTM :)

k-g-a:
Ok, added the section about hooks for map/array to README.md and moved common interfaces to Hooks.ts.
Actually, could not think of any additional tests, but i'm open to suggestions.
I'd also appreciate if @xaviergonz could check the typesystem part :)

k-g-a:
@xaviergonz Hi! Any chance to realease it soon?

Baael:
Is it in npm version of MST, 3.14.1?

```
types.map(...).hooks is not a function
```

PR #1370, opened 2019-08-16
chore: Add & apply yarn deduplicate

Bnaya:
Motivation:
yarn is actually missing automatic dedup of deps,
Which leads to multiple-but-compatible versions of deps.
[yarn-deduplicate](https://github.com/atlassian/yarn-deduplicate) is a complementary tool to solve that.

It's recommended to run it after any dependencies add/remove/upgrade

Related:
https://github.com/yarnpkg/yarn/issues/3778
(And there are more issues around yarn repo related to that)

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
         "size": "size-limit",
         "coverage": "lerna run coverage --stream",
         "release": "yarn build && lerna publish",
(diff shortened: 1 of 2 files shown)


xaviergonz:
LGTM, thanks!

PR #1401, opened 2019-10-05
docs(getting started): update CRA step

pedro-mass:
It's now recommended to use `npx` for installs

https://github.com/facebook/create-react-app#quick-overview


xaviergonz:
Thanks!

PR #1405, opened 2019-10-23
Use gitbook 'Getting Started' link

jaythomas:
I was really confused by the unrendered docs until I realized it was a gitbook. There is no reference to the gitbook in the README otherwise.

Not sure if the gitbook is an officially-maintained url but I certainly find it more accessible than browsing the markdown files in the repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Supported devtools:
 
 # Getting started
 
-See the [Getting started](https://github.com/mobxjs/mobx-state-tree/blob/master/docs/getting-started.md#getting-started) tutorial or follow the free [egghead.io course](https://egghead.io/courses/manage-application-state-with-mobx-state-tree) (note however that the course is for MST v2, so it might be a bit outdated).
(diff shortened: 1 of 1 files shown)


xaviergonz:
That gitbook actually comes from an external repository. Does anybody know if it is being kept up to date with the current docs?

Either way I'd at least offer both links with a warning on the gitbook one that it might become deprecated.

jaythomas:
Just saw this issue on the mirror repo: https://github.com/epodgaetskiy/mobx-state-tree-docs/issues/2

It would be simpler if @mobxjs maintained the gitbook.io docs, but if the author is unavailable or unwilling to collaboratively maintain the published docs then this org should consider hosting the gitbook elsewhere (ie a [github pages](https://pages.github.com/) branch/folder). If it is linked in the README then search engines will do their thing and rank appropriately.

mweststrate:
Yes, the linked repository is generally outdated, and unrelated to this project. I'll try to create an official website for the docs soon, as they have outgrown the reasonable github readme size.

PR #1409, opened 2019-10-28
[Typescript] Improving flow typings #1378

### LINKED ISSUE #1378: [TypeScript 3.6+] flow() | Inferred Generator type is not assignable to IterableIterator
**_Feature request_**

**Is your feature request related to a problem? Please describe.**
Prior to TS 3.6 `flow()` yield return types would be typed as `any` due to typescript limitations. Those limitations are being addressed in TS 3.6 release as described [in this iteration plan](https://github.com/microsoft/TypeScript/issues/31639)

**Describe the solution you'd like**
Full support for yield return types in `flow()`

**Describe alternatives you've considered**
If the full inference is not possible due to some TS limitations (I haven't completely dived into new generators typings) I wouldn't mind having to manually declare generic parameters in `flow()`

**Additional context**
I've seen some mentions [in this PR](https://github.com/mobxjs/mobx-state-tree/pull/1240), but I have not seen the actual issue for tracking this particular case

**Are you willing to (attempt) a PR?**
* [ ] Yes
* [x] No

nulladdict:
This is an attempt to fix #1378

This is a breaking change to flow typings that requires TS version 3.6+
I am still not sold on the final solution, but I see two major alternatives:

- the one described in [this comment](https://github.com/mobxjs/mobx-state-tree/issues/1378#issuecomment-532639990) and implemented in this PR
- typing flow as `Generator<Promise<any>, R, unknown>`, so that we get inferred return type, type-check on yielded promises and that we're required to type yield results, which is somewhat similar to what we had before

Both of this are not perfect and still breaking, because TS changed `Generator` interface.
It seems that there's a possibility to escape the breaking change by utilising `typesVersions`  as mentioned [here](https://github.com/mobxjs/mobx-state-tree/issues/1378#issuecomment-527531474), however I am not quite sure I understand how to do that

It is also worth mentioning that with arriving of [`actionAsync`](https://github.com/mobxjs/mobx-utils#actionasync) this problem might not  be that important, and we might just use the second alternative to fix TS errors and not rely on lots of TS magic

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/async.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/async.test.ts
@@ -329,7 +329,7 @@ test("flow typings", async () => {
         numberToNumber: flow(function*(val: number) {
             yield promise
             return val
-        }), // should be () => Promise<Promise<number>>
+        }), // should be () => Promise<number>
         voidToNumber: flow(function*() {
             yield promise
             return Promise.resolve(2)
@@ -345,4 +345,8 @@ test("flow typings", async () => {
     expect(b).toBe(4)
     const c: number = await m.voidToNumber()
     expect(c).toBe(2)
+    await m.voidToNumber().then(d => {
+        const _d: number = d
+        expect(_d).toBe(2)
+    })
 })
--- a/packages/mobx-state-tree/src/core/flow.ts
+++ b/packages/mobx-state-tree/src/core/flow.ts
@@ -1,11 +1,16 @@
 /**
+ * @hidden
+ */
+export type FlowReturn<R> = R extends Promise<infer T> ? T : R
+
+/**
  * See [asynchronous actions](https://github.com/mobxjs/mobx-state-tree/blob/master/docs/async-actions.md).
  *
  * @returns The flow as a promise.
  */
 export function flow<R, Args extends any[]>(
-    generator: (...args: Args) => Generator<any, R, any>
-): (...args: Args) => Promise<R> {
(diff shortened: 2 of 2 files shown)


xaviergonz:
flow typings should be fixed (when using ts3.6) with version 3.15.0 that was just released, give it a try :)

nulladdict:
`3.15.0` looks good, however I believe it introduces a type regression when returning a Promise:

```ts
const f = flow(function*() {
  yield promise
  return Promise.resolve(2)
})
// f is () => Promise<Promise<number>>
```
```ts
const x = await f() // x is correctly inferred as number
f().then(x => { /* ... */ }) // x is Promise<number> when in fact it's just a number
```

taking a look at flow return code:
```ts
wrap((r: any) => { resolve(r) }, "flow_return", ret.value)
```
`resolve(r)` when `r` is a promise would produce a `Promise<T>` and not a nested one, so the correct return type in our example is `Promise<number>`, await just hides the issue

I've fixed this and added a test for this specific case
Also I've change yielded type from `any` to `Promise<any>` since we don't support any other yielded values anyway

xaviergonz:
Technically I THINK ` then` will also resolve to a number even when taking `Promise<Promise<number>>`, so that shouldn't matter much, but I guess it is better to keep API typing compatibility.

Also forcing yield to take promises should help catching bugs, so LGTM 😁

mweststrate:
Released in 3.16. Apologies for the late follow up!

PR #1410, opened 2019-10-28
[CustomType] Add 'environment' as second argument for options.fromSnapshot

k-g-a:
It seems like a common case to modify `custom-type`'s value using environment, i.e. for mocking dates in tests or using i18n.
This PR makes it possible via passing current tree's environment as second argument to the `fromSnapshot` method.
Initially I tried to pass `parent` - but it's `ObjectNode` which is non-conventional for MST (we ususally do not pass bare nodes to userland). My second attempt was to pass `parent.storedValue` (which is `IAnyStateTreeNode` and is more common), but it's `undefined` by the time we call `fromSnapshot` and we can not call `.value` either as we are actually in the middle of the observable node creation.
So the only valid option left is to pass at least the environment itself.

Typedoc edited. 
Test added.


packages/mobx-state-tree/src/types/utility-types/custom.ts
```diff
@@ -13,8 +13,8 @@ import {
 export interface CustomTypeOptions<S, T> {
     /** Friendly name */
     name: string
-    /** given a serialized value, how to turn it into the target type */
-    fromSnapshot(snapshot: S): T
+    /** given a serialized value and environment, how to turn it into the target type */
+    fromSnapshot(snapshot: S, env: any): T
```
mweststrate: Should be `env?: any` I think, since this is a public interface. Beyond that LGTM

mweststrate:
Released as 3.16.0. Terribly sorry for the late follow up! Do ping me next time if this is annoying!

PR #1419, opened 2019-11-09
(docs): fix unmatched italic in identifiers docs

agilgur5:
- didn't get italicized before this and instead had a visible
  underscore at the beginning


mweststrate:
Note: please don't merge yet, as I am migrating the docs :). So by merging now the changes might be lost accidentally

mweststrate:
Merged manually in website branch. Thanks!

PR #1422, opened 2019-11-11
Fix extraneous actions from ComplexType.getValue (and elsewhere)

### LINKED ISSUE #1421: The action "createObservableInstanceIfNeeded" is run every time object is accessed
**_Bug report_**

* [X] I've checked documentation and searched for existing issues
* [X] I've made sure my project is based on the latest MST version
* [X] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

https://codesandbox.io/s/mobx-state-tree-todolist-uo625

I expect that merely accessing data within a MST tree will not cause actions to be dispatched, which show up in the MobX devtools and make it hard to see the "real" actions.

Every time `ComplexType.prototype.getValue` is called, it runs the below:
https://github.com/mobxjs/mobx-state-tree/blob/d687833fd13d7b068e2792213a2c6296781775a6/packages/mobx-state-tree/src/core/type/type.ts#L403-L406

As can be seen, it calls `node.createObservableInstanceIfNeeded()`. The problem is, `node.createObservableInstanceIfNeeded` is a method with the `@action` decorator applied. Thus, every time a ComplexType's value is retrieved, it causes an action to be dispatched, which has the negatives described earlier (pollutes the MobX dev-tools Changes panel).

I know the MobX dev-tools Changes panel lets you set a "filter" that hides certain changes (eg: `/^(?!createObs)`).

However:
1) The filter does not persist between page refreshes.
2) It's an unpleasant workaround that new developers should not have to figure out on their own.
3) It may block out legitimate instances of that call that the developer might want to still see.

My suggestion: Split the method into two parts.
```
createObservableInstanceIfNeeded(): void {
	if (this._observableInstanceState !== ObservableInstanceLifecycle.UNINITIALIZED) {
		return
	}
	this.createObservableInstance()
}

@action
createObservableInstance(): void {
	[...]
}
```

Basically: Split off the portion that needs to be in an action into a separate `createObservableInstance` method. This way, for the 90%+ of times where you're just reading the value (and you don't need to create a new observable instance), the "if needed" check in the non-action part will see that an observable doesn't need to be created, thus avoiding the need to call the method with the `@action` decorator, thus greatly cutting down on extraneous actions that show up in the Changes panel.

I can create a pull-request for this change (as it's very simple -- just the code above), however want to bring it up in an issue first to make sure it's the desired solution to the problem.

Venryx:
* Fixed that calling "createObservableInstanceIfNeeded" would execute an action, even if the function immediately returned. (significant since the extraneous actions would pollute the mobx dev-tools on mere accesses, eg. by ComplexType.prototype.getValue)

Resolves #1421.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,7 @@
+# 3.15.1
+
+- Fixed that calling "createObservableInstanceIfNeeded" would execute an action, even if the function immediately returned. (significant since the extraneous actions would pollute the mobx dev-tools on mere accesses, eg. by ComplexType.prototype.getValue) Fixes [#1421](https://github.com/mobxjs/mobx-state-tree/issues/1421)
+
 # 3.15.0
 
 - Fix for flow typings. This means that now using flows requires at least TypeScript v3.6 and that `castFlowReturn` becomes deprecated.
--- a/packages/mobx-state-tree/src/core/node/object-node.ts
+++ b/packages/mobx-state-tree/src/core/node/object-node.ts
@@ -172,11 +172,14 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
         }
     }
 
-    @action
     createObservableInstanceIfNeeded(): void {
-        if (this._observableInstanceState !== ObservableInstanceLifecycle.UNINITIALIZED) {
-            return
+        if (this._observableInstanceState === ObservableInstanceLifecycle.UNINITIALIZED) {
+            this.createObservableInstance()
         }
+    }
+
+    @action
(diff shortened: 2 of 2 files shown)


Venryx:
What version do I put it under? 3.15.1?

mweststrate:
Thanks for this!

mweststrate:
Released in 3.16.0. I'm sorry I didn't release this earlier! Feel free ping me next time, and if you want to contribute / maintain or even coordinate releases, just let me know @Venryx!

Issue #1417, opened 2019-11-08
Limited history with UndoManager?


tibotiber:
**_Question_**

* [x] I've checked documentation and searched for existing issues
* [x] I tried the [spectrum channel](https://spectrum.chat/?t=dad48299-3dfc-4e10-b6da-9af1e39498a3)

Hi,

I'm a long time MST user and love it, thanks for the work that went into this!
I'm using the UndoManager for the first time and was wondering if there was a way to fix the maximum length of the history to avoid high memory / local storage (if persisted) consumption?
If it's not there, would this be easy to manage in userland or is it something that need to be added to the middleware?
If it must be added, is there an interest to add it or would you recommend a fork for my needs? And any pointers where / how you would think to implement it?

Thanks in advance!

See also [on Spectrum](https://spectrum.chat/mobx-state-tree/general/limited-history-with-undomanager~0972e1f4-63ba-4981-9972-415849a9fcc5).

mweststrate:
The middle ware is defined here, and not that complex as you see: https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mst-middlewares/src/undo-manager.ts

It saves patches (diffs), and not the full history, so in most cases memory usage will probably not be that high. There is a `clear` action to truncate the history. I guess it would be easy enough to add an action that removes everything but a limited set (e.g. `clear(100)` to clear everything but the last 100 items. Feel free to PR something like that :)

tibotiber:
Thanks! Will have a look and open a PR. I would add an optional argument to `clear`. Would doing the same to `clearUndo` and `clearRedo` make sense?

Extra but keen to explore if you're fine with it:
We could pass in an option when creating the middleware (`maxHistoryLength` or something like that), then make sure we automatically maintain a maximum history (or undo only?) length. From what I've understood so far, that would happen in `addUndoState` where we push to the history.

mweststrate:
Oh that would be the more natural approach indeed. Great idea!

Op za 9 nov. 2019 08:52 schreef Thibaut Tiberghien <notifications@github.com

tibotiber:
Cool, I'll do that. Traveling end of this week, so expect something in 10 days or so.

mweststrate:
Released in 3.16.0. Finally, sorry about that! Feel free to ping me next time if you need it faster :)

PR #1426, opened 2019-11-19
add maxHistoryLength option to UndoManager

tibotiber:
Hi @mweststrate, here is the PR discussed in #1417. Let me know if it requires any changes.


tibotiber:
@mweststrate thanks for merging. You can close #1417 as well.

PR #1431, opened 2019-12-03
fix broken links in README.md

ArturJS:
Some links in README.md seemed to be broken.
Let me know if there is anything else that should also be included in this PR.


ArturJS:
Hmm, some tests were failed.
@mweststrate Would it be ok if I add fix for them as a part of this PR?

mweststrate:
Thanks for looking into this! For the intro we can probably best link to https://mobx-state-tree.js.org/intro/getting-started now. 

Good catch about the tests! They are most probably failing because a filepath is wrong, as they check if all methods are documented in the api overview .md file

ArturJS:
@mweststrate Done.

mweststrate:
Awesome, thanks @ArturJS (& @auvipy)!

Issue #1156, opened 2019-01-28
Cannot resolve a reference with multiple candidates (two model instances with same identifier)


zisiszikos:
**_Bug report_**

* [x] I've checked documentation and searched for existing issues
* [x] I've made sure my project is based on the latest MST version
* [ ] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

**Sandbox link or minimal reproduction code**
[Sandbox example](https://codesandbox.io/s/2x10pl6j4n)

**Describe the expected behavior**
I have a `Base` model that is composed in two places `one, two` in the store tree. This model uses nested models `Country` and `City`. I add a new country and a new city in both places with the same identifiers. The cities are also placed in an array of references at the country model. Now I want to get the cities of a country by reading the list of cities references.

**Describe the observed behavior**
When trying to get the list of cities from references list of `Country` model, the following error throws:
```
[mobx-state-tree] Cannot resolve a reference to type 'late(function () {
    return City;
  })' with id: '1000' unambigously, there are multiple candidates: /one/cities/1000, /two/cities/1000
```
The same is NOT happening when trying to get a country, even though there are multiple countries with the same identifier, but not in a list of references.

It seams that resolving an item by its reference identifier not working if there are multiple items with the same identifier.

xaviergonz:
https://github.com/mobxjs/mobx-state-tree/blob/master/README.md#references

Identifiers for objects of the same kind must be unique across the whole tree that spawns the whole root store, this is because in their snapshot form references are saved as plain strings /numbers that must point to a single object. 

There are ways to escape this such as custom references or views that resolve the nodes in a custom way.

Not sure if that's what you meant though 🙂

zisiszikos:
Thanks for the answer.

I understand what you are saying but there might be situations where a model needs to be used in multiple places in the store. For example, imagine a `User` model that can be used for regular product users and for administrator users, both stored at different arrays. The only solution now is to duplicate the model.

Anyway, at the time, I solved my problem by replacing the `safeReference` type property with a view getter that returns the same thing.

[New example here](https://codesandbox.io/s/olvvl34z2y)

xaviergonz:
Another way to tackle the problme would be to use a custom reference that resolves to a certain array in the store rather than relying on the built-in resolveIdentifier.

To be honest I don't use references that much, I usually just use ids as properties and views to resolve them :)

xaviergonz:
[...]

PR #1439, opened 2019-12-12
[On hold] Add option to resolve non unique identifiers

filipkis:
This is PR partially related to #1156 

The main goal is to provide `resolveIdentifier` version that doesn't throw if there are multiple instances of the same id in the tree, but rather returns first one (i.e. first one in the cache - might not be the first one in the tree structure).

### DIFF
--- a/packages/mobx-state-tree/src/core/mst-operations.ts
+++ b/packages/mobx-state-tree/src/core/mst-operations.ts
@@ -517,17 +517,22 @@ export function resolvePath(target: IAnyStateTreeNode, path: string): any {
 
 /**
  * Resolves a model instance given a root target, the type and the identifier you are searching for.
+ * By default it throws error if there is more then one instance found (i.e. instance is not unique).
+ * This can be overriden by last optional parameter set to false, in which case it will return the
+ * first instance found.
  * Returns undefined if no value can be found.
  *
  * @param type
  * @param target
  * @param identifier
+ * @param unique
  * @returns
  */
 export function resolveIdentifier<IT extends IAnyModelType>(
     type: IT,
     target: IAnyStateTreeNode,
-    identifier: ReferenceIdentifier
+    identifier: ReferenceIdentifier,
+    unique = true
 ): IT["Type"] | undefined {
     // check all arguments
     assertIsType(type, 1)
@@ -536,7 +541,8 @@ export function resolveIdentifier<IT extends IAnyModelType>(
 
     const node = getStateTreeNode(target).root.identifierCache!.resolve(
         type,
-        normalizeIdentifier(identifier)
+        normalizeIdentifier(identifier),
+        unique
     )
     return node ? node.value : undefined
 }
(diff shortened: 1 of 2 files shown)


mweststrate:
I think having an identifier for the same type twice is a design smell; either the tree is not a tree, or redundant data is being stored. I think the use case for a feature like this would need extensive discussion with a scenario first.

filipkis:
Here is a use case:

Our app has comments and each comment has an author. The API returns array of comments and for each comment it provides author object as well. So what would you propose in that case? 

One thing I could imagine is removing authors object from comment and moving it to an authors set and then just using reference. But that would require a lot of pre processing of API results (our case is not as simple as above described). As the author data never actually gets modified on the frontend, it's just used to display information, so we really don't care if there are duplicates in the tree.

Now, in another place, (activity log) we have only author id and the easiest way for us to display the author name in the log for us would be to search our tree for the author with same id. But currently we can't do this with `resolveIdentifier`.

filipkis:
If philosophy is of concern, we could make this into another function so it's not a `resovleIdentifier`, but something like `findInCache` or something like that.

mweststrate:
If you are not intending to refer / modify / merge / normalize etc the
authors, I'd rather store them as frozen fields, and maintain your own
lookup map (e.g. in the lifecycle hooks of the comments), rather than
relying on MST's identifier mechanism, which cannot be seen independently
of the bigger picture of reconciliation, serialization, reference resolve
etc. No need to treat everything as a nail just because there is a hammer
provided :)

filipkis:
> I'd rather store them as frozen fields, and maintain your own lookup map

But they're more than frozen fields. We have some views on them and even some actions (that do async api calls) that are used in other cases the same model is used. Therefore, sure, we could "drop" the MST for this model and do everything manually, but if so much of what we have is already in MST, just because for this part of the code we (currently) don't need full might of MST, doesn't mean that some part of it are not already valuable. And later, if needed we might normalize stuff and it will for sure be easier if we already have everything in MST and just need to shift things around a bit.

filipkis:
OK, I guess I'm not going to convince you to nudge here (and get the benefit of already existing cache and functions that need just slight modification vs. implementing my own lookups). 

I do want to point out that we have this setup (multiple instances with same id) in our tree for some time now and we have had no problems/errors (until this feature we wanted to implement now). So if the goal is early detection of modelling errors, then it's happening only when identifiers are used in resolution. So maybe the detection should be stricter or you might accept that certain modelling errors are allowed and left to users to decide on level of purity they want (couldn't help myself trying to one more time argue that this minor change wouldn't hurt :P).

jamonholmgren:
I'm going to put a pin in this to revisit later, as it's an interesting problem that needs a more elegant solution IMO.

jamonholmgren:
FYI, the new [mst-reference-pool](https://github.com/infinitered/mst-reference-pool) is the better solution I came up with.

jamonholmgren:
Closing in favor of mst-reference-pool.

PR #1451, opened 2020-01-10
(docs): fix link to async-actions docs

frontstall:


mweststrate:
Closing as this is the generated file, not the source. Will update in the source. Thanks for noticing!

Issue #734, opened 2018-03-26
in case of failure typecheck dumps out the whole snapshot


jayarjo:
When checking a snapshot for compliance to the root model (in case of failure) `typecheck` dumps out the whole snapshot and the whole type tree, making it nearly impossible to visually locate discrepancy. Could it output only the discrepancy between snapshot and model instead? Or at least output it at a separate line?

Bnaya:
I have the same pain point.
My local workaround, when big arrays are involved, is to catch the error from `create` and then run create call separately for each array item separately. that way it will throw again, but with granular error

maxgallo:
Hey @jayarjo, I think what you ask is happening already (or maybe you want more detailed information)
In the second error in the console log, **the first line of the call stack is the thing**, I suppose, you're looking for.

<img width="902" alt="screen shot 2018-03-31 at 12 32 27" src="https://user-images.githubusercontent.com/3715587/38162801-d4f16a58-34df-11e8-910e-793e988da76d.png">

jayarjo:
It is not a problem when snapshot is small. You should try a nested one, with hundreds of items in arrays somewhere down the tree.

maxgallo:
I don't have a direct experience with hundreds of object, I daily have to deal with dozens of objects max. In my experience usually the first few lines of the stack trace, after the big dump of the whole thing, are really useful.

mweststrate:
Problem is clear, PRs improving this are welcome

jaythomas:
> Problem is clear, PRs improving this are welcome

So the improvement would be to... display the model where the offense is happening versus the entire model that is type-checked? Or how about don't display any data blob and just show the offending properties? Throwing my hat in the ring...

Given the model:

```js
import { types as t } from 'mobx-state-tree'

export default t.model('statement', {
  id: t.string,
  label: t.string,
  transactions: t.array(
    t.model('transaction', {
      id: t.string,
      amount: t.maybeNull(t.number),
      completed_on: t.model('dateSignature', {
        // YYYY-MM-DD
        date: t.refinement('serverDateFormat', t.string, value => {
          return Boolean(value.match(/^[1-9]\d{3}-[01]\d-[0123]\d/))
        })
      })
    })
  )
})
```

And the server data from a network request:

```js
import statementModel from './models/statement'

const statement = statementModel.create({
  id: 'daf672da-9195-4804-b177-9d09c45d3b8e',
  label: ['October Week 1'],
  transactions: [
    {
[...]

PR #1474, opened 2020-02-23
Improve type checker error message output

jaythomas:
This is a proof-of-concept for the formatting proposal mentioned in #734.
It is not very robust and breaks all the tests that check against the error message, but you can see the test I added and if there is good feedback I'll put some greater effort into getting this PR-ready.

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/formatting.test.js
+++ b/packages/mobx-state-tree/__tests__/core/formatting.test.js
@@ -1,0 +1,68 @@
+const t = require("../../src").types
+
+const statementModel = t.model("statement", {
+    id: t.string,
+    label: t.string,
+    transactions: t.array(
+        t.model("transaction", {
+            id: t.string,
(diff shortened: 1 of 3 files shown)


cmdcolin:
I would love something like this

Our state tree model is very large and has lots of different subtypes and error messages have gotten very big. We actually have to manually catch all errors and truncate them because otherwise react-error-overlay takes 100% cpu and never finishes sometimes. I tried yarn linking this and I got a much more manageable error. I love it :)

Before (see...Show more (15Mb)!)
Before

After
After

cmdcolin:
PS Original screenshots in my above post were duplicates of each other but should illustrate it now just in case you saw it via email

jaythomas:
Thank you for the feedback. Unfortunately the project I was using mobx-state-tree for we switched over to using a custom state management solution as mobx's `ObservableArray` that MST uses in place of native arrays wasn't compatible with our component framework, Vue.js.

PR #1501, opened 2020-04-12
Add setImmediateWithFallback util

isaachinman:
@mweststrate Let me know if you want test coverage. Also, you specifically requested `setTimeout(fn, 1)`, but I think a value of zero would do just as well in this case, no?

### DIFF
--- a/packages/mobx-state-tree/src/core/flow.ts
+++ b/packages/mobx-state-tree/src/core/flow.ts
@@ -1,3 +1,5 @@
+import { setImmediateWithFallback } from "../utils"
+
 /**
  * @hidden
  */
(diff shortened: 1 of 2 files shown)


xaviergonz:
Looks good to me, just I'm not sure why settimeout is set to 1 instead of 0

mweststrate:
IIRC it doesn't matter, anything below 5 defaults to 5 in most browsers.

mweststrate:
LGTM, thanks for investigating this!

mweststrate:
Released in 3.16.0.

wscotten:
@mweststrate awesome!  Thank you so much!!!!

PR #1519, opened 2020-05-01
fix applySnapshot type checking

### LINKED ISSUE #1354: `applySnapshot` seems to preprocess the snapshot twice
**_Bug report_**

https://codesandbox.io/s/flamboyant-fire-4qoxr

I have a model with both a `preProcessSnapshot` and `postProcessSnapshot`.

I want to make some changes on it, but rollback the changes if something fails, so here's my process:
- Backup the model using `getSnapshot`
- Perform some modifications on the instance
- Oops, something fails...
- Rollback using `applySnapshot(self, backupSnapshot)`

I get at "Error while converting..." error, that seems to indicate the preProcessor ran twice.

My model is defined like this:
```js
const Model = types
  .model({
    id: types.string,
    realId: types.number,
    foo: types.string
  })
  .preProcessSnapshot(snapshot => ({
    ...snapshot,
    realId: snapshot.id,
    id: "some-hash"
  }))
  .postProcessSnapshot(snapshot => ({
    ...snapshot,
    realId: undefined,
    id: snapshot.realId
  }));
```

`getSnapshot` returns:
```js
{
  id: 123,
  realId: undefined,
  foo: 'bar',
}
```

Yet, applying that snapshot fails:
```
at path "/realId" value `"some-hash"` is not assignable to type: `number`
```

The only way for such an error to happen, would be for `preProcessSnapshot` to have run twice.

```js
// Snapshot
{
  id: 123,
  realId: undefined,
  foo: 'bar',
}

// After one preProcess
{
  id: 'some-hash',
  realId: 123,
  foo: 'bar',
}

// After two preProcess
{
  id: 'some-hash',
  realId: 'some-hash',
  foo: 'bar',
}
```

mishk0:
`typecheckInternal` should be called with an original snapshot. It seems to fix #1354


jamonholmgren:
@mishk0 It looks like the `typecheckInternal` was moved at some point. Does that make this PR unnecessary (and #1354 fixed)?

adamkovalsky:
@jamonholmgren this is still an issue. The problem is `ModelType.applySnapshot` runs the snapshot through `applySnapshotPreProcessor` and then calls `typecheckInternal`, which leads to `isValidSnapshot` calling `applySnapshotPreProcessor` on the already processed snapshot.

See:

https://github.com/mobxjs/mobx-state-tree/blob/8f3706d08e35f600d12a658ae1ecb19a12f85638/packages/mobx-state-tree/src/types/complex-types/model.ts#L678-L679

and

https://github.com/mobxjs/mobx-state-tree/blob/8f3706d08e35f600d12a658ae1ecb19a12f85638/packages/mobx-state-tree/src/types/complex-types/model.ts#L703

jamonholmgren:
@adamkovalsky Gotcha, thanks for the info! Would you want to submit a new PR without conflicts that fixes this issue? A regression test would be :100: too.

adamkovalsky:
@jamonholmgren I'm afraid I haven't the time to submit a new PR. @mishk0 ?

jamonholmgren:
@mishk0 Still interested in pushing this forward?

adamkovalsky:
@jamonholmgren as far as I know the original plan was to deprecate snapshot processors on models, and only rely on `types.snapshotProcessor`. Is that still the plan? In that case maybe this is not worth fixing.

jamonholmgren:
I went ahead and un-deprecated `pre/postProcessSnapshot` in #1986, so I think this should be shipped.

jamonholmgren:
This will still result in the preProcessSnapshot running twice; however:

1. It will not double-apply the snapshot before validation
2. It will only run twice in development mode (or if runtime type checking is turned on in prod, which is rare)

So I think this is an acceptable fix, and will ship in the next release of mobx-state-tree.

If someone wants to figure out a way to only run it once even in dev, feel free. Not a high priority right now.

jamonholmgren:
Released in v5.1.8: https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.1.8

geohuz:
Well, this is quite annoying I spend whole night to figure out why it's being ran twice, hope it can be fixed in development mode.

scytacki:
I believe this also means if `preProcessSnapshot` modifies its snapshot parameter in place, the modified snapshot will get passed to the second call to `preProcessSnapshot` and the result of this second call is what will actually be applied.  At least we are seeing this behavior when going from 5.1.5 to 5.1.8.  

In any case, `preProcessSnapshot` should not be modifying the snapshot in place, so the fix was easy. I'm just bringing this up in case someone else runs into it.

PR #1543, opened 2020-06-30
Add toGenerator utility function

fruitraccoon:
The `toGenerator` function is intended for wrapping a promise-returning function to create a generator-returning function. This can then be used to obtain a strongly-typed return value in async actions using `yield*` (rather than `yield`).

This approach has been discussed in [this Spectrum thread](https://spectrum.chat/mobx-state-tree/general/typescript-typing-for-async-actions-yield~2bcd1b6c-2449-4c56-987f-4483fc57e32b).

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/async.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/async.test.ts
@@ -363,8 +363,11 @@ test("flow typings", async () => {
  * https://stackoverflow.com/a/55541672/4289902
  */
 type IfAny<T, Y, N> = 0 extends 1 & T ? Y : N
-function ensureNotAny<T>(value: IfAny<T, never, T>) {}
-function ensureType<T>(value: T) {}
+
+/**
+ * Ensure that the type of the passed value is of the expected type, and is NOT the TypeScript `any` type
+ */
+function ensureNotAnyType<TExpected, TActual>(value: IfAny<TActual, never, TExpected>) {}
 
 test("yield* typings for toGeneratorFunction", async () => {
     const voidPromise = () => Promise.resolve()
@@ -379,16 +382,13 @@ test("yield* typings for toGeneratorFunction", async () => {
     const M = types.model({ x: 5 }).actions((self) => {
         function* testAction() {
             const voidResult = yield* voidGen()
-            ensureNotAny(voidResult)
-            ensureType<void>(voidResult)
+            ensureNotAnyType<void, typeof voidResult>(voidResult)
 
             const numberResult = yield* numberGen()
-            ensureNotAny(numberResult)
-            ensureType<number>(numberResult)
+            ensureNotAnyType<number, typeof numberResult>(numberResult)
 
             const stringResult = yield* stringWithArgsGen("input", true)
-            ensureNotAny(stringResult)
-            ensureType<string>(stringResult)
+            ensureNotAnyType<string, typeof stringResult>(stringResult)
 
             return stringResult
         }
@@ -401,8 +401,7 @@ test("yield* typings for toGeneratorFunction", async () => {
     const m = M.create()
 
     const result = await m.testAction()
-    ensureNotAny(result)
-    ensureType<string>(result)
+    ensureNotAnyType<string, typeof result>(result)
     expect(result).toBe("test-result")
 })
 
@@ -415,16 +414,13 @@ test("yield* typings for toGenerator", async () => {
     const M = types.model({ x: 5 }).actions((self) => {
         function* testAction() {
             const voidResult = yield* toGenerator(voidPromise())
-            ensureNotAny(voidResult)
-            ensureType<void>(voidResult)
+            ensureNotAnyType<void, typeof voidResult>(voidResult)
 
             const numberResult = yield* toGenerator(numberPromise())
-            ensureNotAny(numberResult)
-            ensureType<number>(numberResult)
+            ensureNotAnyType<number, typeof numberResult>(numberResult)
 
             const stringResult = yield* toGenerator(stringWithArgsPromise("input", true))
-            ensureNotAny(stringResult)
-            ensureType<string>(stringResult)
+            ensureNotAnyType<string, typeof stringResult>(stringResult)
 
             return stringResult
         }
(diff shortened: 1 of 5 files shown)


fruitraccoon:
Please note that I've not regenerated the docs (which causes a single test failure), as when I do there are many changes relating to the name of my fork project, rather than the real one.

Bnaya:
Thanks @fruitraccoon! i will review it at the weekend i believe

packages/mobx-state-tree/src/utils.ts
```diff
@@ -483,3 +483,30 @@ export function setImmediateWithFallback(fn: (...args: any[]) => void) {
...
+ * const someModel.actions(self => ({
+ *   someAction: flow(function*() {
+ *     // value is typed as number
+ *     const value = yield* getDataGen("input value");
+ *     ...
+ *   })
+ * }))
+ * ```
+ */
+export function toGenerator<R, Args extends any[]>(
```
Bnaya, line 506: Let's move it next to?
https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/core/flow.ts#L19-L28
feels contextually more sound

packages/mobx-state-tree/src/utils.ts
```diff
@@ -483,3 +483,30 @@ export function setImmediateWithFallback(fn: (...args: any[]) => void) {
         setTimeout(fn, 1)
     }
 }
+
+/**
```
Bnaya, line 487: Add experimental note, that this might change on minor releases

packages/mobx-state-tree/__tests__/core/async.test.ts
```diff
@@ -356,3 +357,51 @@ test("flow typings", async () => {
...
+
+        return {
+            testAction: flow(testAction)
+        }
+    })
+
+    const m = M.create()
+
+    const result = await m.testAction()
+    ensureNotAny(result)
```
Bnaya, line 404: Now that the repo is upgraded to ts 3.9, we may use 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments
for this kind of assertions

Bnaya:
Would you mind adding also:
```ts
/**
 * @experimental 
 * experimental api - might change on minor/patch releases
 * 
 * Convert a promise to a generator yielding that promise
 * This is intended to allow for usage of `yield*` in async actions to
 * retain the promise return type.
 *
 * Example:
 * ```ts
 * function getDataAsync(input: string): Promise<number> { ... }
 *
 * const someModel.actions(self => ({
 *   someAction: flow(function*() {
 *     // value is typed as number
 *     const value = yield* promiseToGenerator(getDataAsync("input value"));
 *     ...
 *   })
 * }))
 * ```
 */
function* promiseToGenerator<R>(p: Promise<R>): Generator<Promise<R>, R, R> {
  return yield p as any;
}
```

And a test 
It's really ok it won't happen on this PR:)

packages/mobx-state-tree/src/utils.ts
fruitraccoon: Makes sense - done.

packages/mobx-state-tree/src/utils.ts
fruitraccoon: Done

packages/mobx-state-tree/__tests__/core/async.test.ts
fruitraccoon: I don't think I can use `@ts-expect-error` in this case?  If I understand correctly, it's for when typescript wants to report an error, but in the context of the test it's the intended code.  

In this test I'm wanting the opposite, as what I'm testing for is that the type of the returned value is "not `any`", which never normally shows up as an error of course.  With the current code, if I change one of the `yield*` calls to just be `yield`, a typescript error appears (as intended).  

I've simplified the check from two functions to one, and added a comment.  Let me know if I've misunderstood though!

fruitraccoon:
No problem, I've added the extra function to convert a promise.  However I've renamed both it and the original function, as when I [look at MDN for generators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator):

> The Generator object is returned by a generator function ...

So my original function name choice was not ideal.  So now the original function is called `toGeneratorFunction` and the new one that converts a promise to a generator is called `toGenerator`.

But let me know if you prefer the original names, something shorter, or something different altogether.

Bnaya:
> No problem, I've added the extra function to convert a promise. However I've renamed both it and the original function, as when I [look at MDN for generators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator):
> 
> > The Generator object is returned by a generator function ...
> 
> So my original function name choice was not ideal. So now the original function is called `toGeneratorFunction` and the new one that converts a promise to a generator is called `toGenerator`.
> 
> But let me know if you prefer the original names, something shorter, or something different altogether.

I personally prefer verbose names (promiseToGenerator)
And one of the reason we mark these functions as experimental is so we can rename them later with out breaking change if w find the name not good.

So i think we can take out the PR from draft mode?

packages/mobx-state-tree/__tests__/core/async.test.ts
Bnaya: What you cloud do: 
```ts
// result return type is string. if there is no type error here, means we have wrong return type
// @ts-expect-error
const notAny: number =  result
;
const isString: string =  result;
```

fruitraccoon: I didn't think of that approach - nice!

I couldn't think of a way to have a "one-liner" to do both checks though (a la the `ensureNotAnyType` function).  Would you prefer to keep the tests as they are now, or switch them to the technique above?

Bnaya:
Merged! thank you!

Bnaya:
Released as v3.17.1, I'm trying to figure out how to update the website
The docs in the repo are updated:
https://github.com/mobxjs/mobx-state-tree/blob/10808b91a96a537b69d77904b2657d7a2b04ffb7/docs/API/index.md#togenerator

PR #1539, opened 2020-06-27
Deps upgrade

Bnaya:

PR #1545, opened 2020-07-05
re generate docs

Bnaya:
Local website with these docs changes is working property

### DIFF
--- a/docs/API/index.md
+++ b/docs/API/index.md
@@ -1,10 +1,10 @@
 ---
 id: "index"
-title: "mobx-state-tree"
+title: "mobx-state-tree - v3.16.0"
 sidebar_label: "Globals"
 ---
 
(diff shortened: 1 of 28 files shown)


docs/API/index.md
```diff
@@ -1,10 +1,10 @@
 ---
 id: "index"
-title: "mobx-state-tree"
+title: "mobx-state-tree - v3.16.0"
```
Bnaya, line 3: probably related to deps changes from:

https://github.com/mobxjs/mobx-state-tree/pull/1539/files#diff-8ee2343978836a779dc9f8d6b794c3b2L9116-L9148

Bnaya: I've tried to set 
https://typedoc.org/guides/options/#includeversion
to false, but it didn't help.
I think it's hard coded somehow with our theme

Issue #1425, opened 2019-11-18
Typescript declaration size


skokenes:
**_Question_**

* [x] I've checked documentation and searched for existing issues
* [ ] I tried the [spectrum channel](https://spectrum.chat/?t=dad48299-3dfc-4e10-b6da-9af1e39498a3)

I have a fairly complex set of models that form a big store. This store is shared between a few services, so I have the store defined in its own package that my other services import from.

I am trying to generate typescript declarations for the models in this store so that my other services that use this store or sub models can get the correct typings. I am not sure if I am doing something wrong but I am getting a declaration file that is 7.4mb and about 109K LOC. This is causing poor performance in my IDE as I make changes.

Looking at the declaration file, I can see that the resulting size is due to reproducing the same type definitions over and over wherever they are used in my model. For example, I have a model type called `Expression` that is used by many other models in my store. This definition gets a unique declaration in every model that is used, rather than reusing a single type declaration for `Expression`.

My questions are:
1) Is this expected behavior?
2) Is there a way to avoid this so that I get slim, faster compiled declarations?
3) Any other tips for how to architect or produce declarations for this type of project?

Thanks

mweststrate:
Make sure to use the recommended `export interface ITodo extends typeof Todo
.Type {}` for all of your model types, that might help.

k-g-a:
Make sure to specify concrete return types for views/actions which return other nodes, otherwise inferred return types are really huge.

skokenes:
Those solutions don't seem to address my problem unfortunately. See the following example repo:
https://github.com/skokenes/mst-typescript-dec-test

In src, I create a store called `ModelA` with a simple string. Then I create two stores that wrap that store:

- `ModelB`: this uses `ModelA` as a nested property the way one typically does in MST, with inferred typing
- `ModelC`: this uses `ModelA` as a nested property but with a hardcoded typing that references `ModelA`'s type

When I build the declarations file for these 3 models, `ModelB` creates a completely new type for the nested `ModelA`. `ModelC` does not, but manually hardcoding every type this way with MST is pretty tough.

Is there a reason that `ModelB` creates new types for nested complex types instead of reusing the existing typings? Can this be avoided? 

[...]

PR #1562, opened 2020-08-02
Add experimental union & compose type helpers

Bnaya:
Following lessons learned in
https://github.com/mobxjs/mobx-state-tree/issues/1425#issuecomment-558008891


mweststrate:
LGTM at first sight. Do you know any downsides of this approach? If you don't know about any, and not tests fail, I'd just make it the standard rather than creating an alternative API :)

I expect btw with the upcoming variadic tuples, quite some overloads of other api's can be cleaned up as well: https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#variadic-tuple-types

Bnaya:
I will give it more time in the weekend!

arodik:
@Bnaya that's awesome! Уou did great work! Are you going to finish it? I think that all MST devs will be grateful

Bnaya:
> @Bnaya that's awesome! Уou did great work! Are you going to finish it? I think that all MST devs will be grateful

I've found some issues with it, that i'm not sure I could make a version fit for general use.
But it's a simple wrapper function around regular union, you may just copy it and use it in your project

jamonholmgren:
@Bnaya Does that mean we should close this PR? Might make sense to put it in the documentation somewhere as a recipe.

Bnaya:
I've closed it for now

philllies:
@Bnaya could you elaborate on the issues you encountered?

PR #1570, opened 2020-08-24
fix IMSTMAP is not an ObservableMap

MoonBall:
fix https://github.com/mobxjs/mobx/issues/2422#issuecomment-674816446.

I use `conditional-type-checks` for test ts type. the code is as below.

```tsx
import { types } from "mobx-state-tree"
import { values } from "mobx"
import { Has, AssertTrue } from "conditional-type-checks"

const Todo = types.model("Todo", {
    id: types.identifierNumber,
    text: types.optional(types.string, "")
})

const TodoListStore = types.model("TodoList", {
    todos: types.optional(types.map(Todo), {})
})

const store = TodoListStore.create({
    todos: {
        "1001": {
            id: 1001,
            text: "Mow Lawn"
        },
        "1002": {
            id: 1002,
            text: "Pickup shopping"
        }
    }
})

const todoListValues = values(store.todos)
todoListValues.forEach((it) => {
    type doTest = AssertTrue<Has<typeof it, { id: number; text: string }>>
})

```

### DIFF
--- a/packages/mobx-state-tree/src/types/complex-types/map.ts
+++ b/packages/mobx-state-tree/src/types/complex-types/map.ts
@@ -64,7 +64,7 @@ export interface IMapType<IT extends IAnyType>
 }
 
 /** @hidden */
-export interface IMSTMap<IT extends IAnyType> {
+export interface IMSTMap<IT extends IAnyType> extends MSTMap<IT> {
     // bases on ObservableMap, but fine tuned to the auto snapshot conversion of MST
 
     clear(): void
@@ -151,7 +151,7 @@ export enum MapIdentifierMode {
     NO
 }
 
-class MSTMap<IT extends IAnyType> extends ObservableMap<string, any> {
+class MSTMap<IT extends IAnyType> extends ObservableMap<string, IT["Type"]> {
     constructor(initialData?: [string, any][] | IKeyValueMap<any> | Map<string, any> | undefined) {
         super(initialData, observable.ref.enhancer)
     }


MoonBall:
@mweststrate PTAL

mweststrate:
Sorry focusing on mobx 6 first. If someone else can review that'd be great

Op di 1 sep. 2020 03:50 schreef Gang Chen <notifications@github.com>:

jesse-savary (changes requested):
Pulled this down and tested it but unfortunately I don't think this solves the problem.

The following test fails:
```
const todoListValues = values(store.todos)
todoListValues.forEach((it) => {
  type doTest = AssertTrue<Has<typeof it, { id: number; text: string }>> |
    AssertFalse<IsAny<typeof it>>
})
```

Somehow `it` is being typed to any. I don't have time to dig deeper but @MoonBall I figured you'd want to chime in. Is this some kind of error on my end?

MoonBall:
Sorry for my late reply.
I rebased master, and there is an error as below.
image

The reason is the return value of `IMSTMap`'s `toJSON()` method is an object rather than an array that is the return value of `ObservableMap`'s `toJSON()` method. So `IMSTMap` is not incompatible with `ObservableMap`. 

Now my commit is out-of-date, I think there is no simple way to fix it 😰.

jamonholmgren:
If someone has a better idea to fix this, feel free to open a new PR or comment here with ideas. In the meantime, I'm going to close this. Thanks for your work on this, @MoonBall !

Issue #1506, opened 2020-04-15
Models with snapshotProcessor can't be added to maps


igaponov:
**_Bug report_**

* [x] I've checked documentation and searched for existing issues
* [x] I've made sure my project is based on the latest MST version
* [x] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

**Sandbox link or minimal reproduction code**
https://codesandbox.io/s/mobx-state-tree-todolist-i5idw

**Describe the expected behavior**
Maps should work with processed models without errors.

**Describe the observed behavior**
Adding a processed model throws the next exception:
```
[mobx-state-tree] Error while converting <Todo@<root>(id: 01)> to `Todo`:

    value of type Todo: <Todo@<root>(id: 01)> is not assignable to type: `Todo`, expected an instance of `Todo` or a snapshot like `snapshotProcessor({ id: identifier; title: string; finished: boolean? })` instead. (Note that a snapshot of the provided value is compatible with the targeted type)
```

mweststrate:
I think you linked the wrong sandbox? It doesn't seem to contain either a map or a snapshot processor

igaponov:
@mweststrate sorry, it didn't save my last changes. Now the link works.

mbest:
I'm getting a similar problem:

> Error: [mobx-state-tree] Error while converting <WebFilterSettingsBase@<root>> to `(WebFilterSettingsBase | undefined)`:
    value of type WebFilterSettingsBase: <WebFilterSettingsBase@<root>> is not assignable to type: `(WebFilterSettingsBase | undefined)`, expected an instance of `(WebFilterSettingsBase | undefined)` or a snapshot like `(snapshotProcessor({ allowedList: string[]?; blockedList: string[]?; hasSettings: boolean; isWebsiteFilterEnabled: boolean; useAllowedListOnly: boolean; hasAnyWebsites: boolean; isBlockedBrowsersEnabled: boolean }) | undefined?)` instead. (Note that a snapshot of the provided value is compatible with the targeted type)

This is from simply trying to assign to a `types.maybe(WebFilterSettings)` using `value = WebFilterSettings.create(...)`. Below is the model definition.

```ts
const WebFilterSettingsBase = t.model("WebFilterSettingsBase", {
    allowedList: t.array(t.string),
    blockedList: t.array(t.string),
    hasSettings: t.boolean,
    isWebsiteFilterEnabled: t.boolean,
[...]

PR #1581, opened 2020-09-14
Snapshot processor is assignable from fix

kidroca:
Related to #1506 

Debugging the sandbox example here: https://codesandbox.io/s/mobx-state-tree-todolist-i5idw

I found that the `isAsignableFrom` check during `validate` resolves to the `BaseType` but it should instead resolve to the `_subtype.isAsignabllFrom` to work correctly when assigning instances:

image

That's why I've added an override to `isAsignabllFrom` in `SnapshotProcessor.ts` which delegates to the subtype under the snapshot processor

It seems the bug appeared in the tests as well and it was covered by an exception expectation:
```ts
    expect(() => s.b.push(n)).toThrow("Error while converting")	
}
```

I've removed these since the fix addressed them and they appear incorrect - you should be able to use an instance created with a snapshot processor in place of the plain instance

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/snapshotProcessor.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/snapshotProcessor.test.ts
@@ -410,9 +410,6 @@ describe("snapshotProcessor", () => {
             detach(n)
             expect(s.a.length).toBe(0)
             expect(getSnapshot(n)).toEqual({ x: 1 })
-            if (process.env.NODE_ENV !== "production") {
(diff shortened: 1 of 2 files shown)


jamonholmgren (approved):
Looks good! Will release in next patch-level version.

jamonholmgren:
We will actually wait for next major version since this _could_ be a breaking change.

jamonholmgren:
I'll release in next major in the unlikely event that this breaks something.

jamonholmgren:
Released in https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.0.0

PR #1591, opened 2020-10-09
Get tutorial to work while no support for mobx6

iswara108:
Suggested to update the codesandbox references as well. At the moment they are broken.


iswara108:
For now, the tutorial is broken.

jamonholmgren:
Thank you @iswara108 !

PR #1593, opened 2020-10-14
Fix example packages

iswara108:
This is to fix some of the example packages.
Each commit here is a different fix.
The first commit is the most important since the "boxes" example is currently broken.
The rest is just to simplify the code a bit and update some packages.


jamonholmgren:
@iswara108 Would you mind resolving conflicts? I'll merge once you do. Thank you!

jamonholmgren (approved):
Thanks @iswara108 and @willrax !

PR #1609, opened 2020-11-18
Update the intro on the README and add a code example section

jamonholmgren:
One of my goals is to make the documentation more "friendly" and approachable. This is a first pass at that.

I've added an introductory paragraph that explains how MST fits into the overall ecosystem (as a Redux alternative) and what it is. I've also added a code example section, as there's nothing like seeing a bit of code to get a feel for a library.

Also adds Infinite Red to Thanks section as we're sponsoring maintenance going forward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,14 +70,20 @@ const rootStore = RootStore.create({
     tweets: [tweet]
 })
 
-// Ready to use in a React component or anywhere:
-const MyComponent = (props) => {
+// Ready to use in a React component, if that's your target.
+import { observer } from "mobx-react-lite"
+const MyComponent = observer((props) => {
     return <div>Hello, {rootStore.author.firstName}!</div>
-}
+})
+
+// Note: since this component is "observed", any changes to rootStore.author.firstName
+// will result in a re-render! If you're not using React, you can also "listen" to changes
+// using `onSnapshot`: https://mobx-state-tree.js.org/concepts/snapshots
 ```
 
 ## Thanks!
 
+-   [Michel Weststrate](https://twitter.com/mweststrate) for creating MobX, MobX-State-Tree, and MobX-React.
 -   [Infinite Red](https://infinite.red) for supporting ongoing maintenance on MST.
 -   [Mendix](https://mendix.com) for sponsoring and providing the opportunity to work on exploratory projects like MST.
 -   [Dan Abramov](https://twitter.com/dan_abramov)'s work on [Redux](http://redux.js.org) has strongly influenced the idea of snapshots and transactional actions in MST.
--- a/README.md
+++ b/README.md
@@ -2,14 +2,18 @@
 
 # mobx-state-tree
 
-_Opinionated, transactional, MobX powered state container combining the best features of the immutable and mutable world for an optimal DX_
-
 [![npm version](https://badge.fury.io/js/mobx-state-tree.svg)](https://badge.fury.io/js/mobx-state-tree)
 [![CircleCI](https://circleci.com/gh/mobxjs/mobx-state-tree.svg?style=svg)](https://circleci.com/gh/mobxjs/mobx-state-tree)
 [![Coverage Status](https://coveralls.io/repos/github/mobxjs/mobx-state-tree/badge.svg?branch=master)](https://coveralls.io/github/mobxjs/mobx-state-tree?branch=master)
-[![Join the community on Spectrum](https://withspectrum.github.io/badge/badge.svg)](https://spectrum.chat/mobx-state-tree)
+[![Join the community on Github Discussions](https://img.shields.io/badge/GitHub-Discussions-lightgrey)](https://github.com/mobxjs/mobx-state-tree/discussions)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,23 +2,83 @@
...
+})
+
+// Now instantiate the store!
+const rootStore = RootStore.create({
+    author: jamon,
+    tweets: [tweet]
+})
+
+// Ready to use in a React component or anywhere:
+const MyComponent = (props) => {
```
akaguny, line 74: MST is about manage data and reaction for this in our components. Not only replace for classes with construcrors.
this example not explain that because nothing will change if i change `rootStore.author.firstName`.
maybe more MVP example with one simple input for add tweets =)

jamonholmgren: Fair point -- I think that adding the `observer` from mobx-react-lite will work.

README.md
```diff
@@ -2,23 +2,83 @@
...
 [![CircleCI](https://circleci.com/gh/mobxjs/mobx-state-tree.svg?style=svg)](https://circleci.com/gh/mobxjs/mobx-state-tree)
 [![Coverage Status](https://coveralls.io/repos/github/mobxjs/mobx-state-tree/badge.svg?branch=master)](https://coveralls.io/github/mobxjs/mobx-state-tree?branch=master)
-[![Join the community on Spectrum](https://withspectrum.github.io/badge/badge.svg)](https://spectrum.chat/mobx-state-tree)
+[![Join the community on Github Discussions](https://img.shields.io/badge/GitHub-Discussions-lightgrey)](https://github.com/mobxjs/mobx-state-tree/discussions)
+
+## What is mobx-state-tree?
+
+Technically speaking, mobx-state-tree (also known as MST) is a state container system built on [MobX](https://github.com/mobxjs/mobx), a functional reactive state library.
+
+This may not mean much to you, and that’s okay. I’ll explain it like this: **MobX is a state management "engine", and MST makes it feel a lot more like Redux by giving it structure.** MST also offers better performance and much less boilerplate than Redux!
```
Romanior, line 14: As a person, having extensive experience in both I would say opposite: MST feels a lot not like Redux, which one of the things we choose it over Redux, it has different architecture, philosophy and mind models, agreed on the following sentence. Also none of your following examples show the similarity you mention.

I would rephrase it in a way:

`MST is MobX on steroids, MST gives structure, patterns and recipes, which are valuable in large team, but also useful is small one when you expect your code to scale rapidly. If we compare it to Redux, MST offers better performance and much less boilerplate code than Redux!`

Thank you for taking care of it.

iswara108:
How about:

MobX-state-tree is a full-scale state management system that combines the
benefits of an immutable state management system like [Redux](
https://redux.js.org/) and a mutable one like [MobX](https://mobx.js.org).

The end result is [transactionality](
https://mobx-state-tree.js.org/concepts/snapshots), [traceability](
https://mobx-state-tree.js.org/concepts/patches), [composition](
https://mobx-state-tree.js.org/concepts/actions), [discoverability,
co-location, and encapsulation](
https://mobx-state-tree.js.org/concepts/trees) - all out of the box,
scalable and without boilerplate.

And a reference to https://www.youtube.com/watch?v=rwqwwn_46kA

README.md
akaguny: now in documentation used onSnapshot callback 
image
if react developers is not primary auditory we can replace it on reaction that change text of dom node by vanillajs.

Can you make codesandbox or another runnable environment for example ?)

README.md
jamonholmgren: @Romanior I like your changes! I'll make some modifications accordingly.

README.md
```diff
@@ -2,23 +2,83 @@
 
 # mobx-state-tree
 
-_Opinionated, transactional, MobX powered state container combining the best features of the immutable and mutable world for an optimal DX_
-
 [![npm version](https://badge.fury.io/js/mobx-state-tree.svg)](https://badge.fury.io/js/mobx-state-tree)
 [![CircleCI](https://circleci.com/gh/mobxjs/mobx-state-tree.svg?style=svg)](https://circleci.com/gh/mobxjs/mobx-state-tree)
 [![Coverage Status](https://coveralls.io/repos/github/mobxjs/mobx-state-tree/badge.svg?branch=master)](https://coveralls.io/github/mobxjs/mobx-state-tree?branch=master)
-[![Join the community on Spectrum](https://withspectrum.github.io/badge/badge.svg)](https://spectrum.chat/mobx-state-tree)
+[![Join the community on Github Discussions](https://img.shields.io/badge/GitHub-Discussions-lightgrey)](https://github.com/mobxjs/mobx-state-tree/discussions)
```
jamonholmgren, line 8: Switch back to Spectrum

akaguny: but why?

jamonholmgren: @akaguny It was a core team decision, but we could revisit it if we had enough energy from the community behind switching. It was determined that Spectrum is working "well enough for now".

PR #1615, opened 2020-11-25
Refer to official TS performance wiki

elektronik2k5:


docs/tips/typescript.md
```diff
@@ -64,7 +64,9 @@ interface ITodoSnapshotIn extends SnapshotIn<typeof Todo> {} // => { title?: str
 interface ITodoSnapshotOut extends SnapshotOut<typeof Todo> {} // => { title: string }
 ```
 
-Note, it is important to use `interface` and not `type` when constructing those types! Although `type`s will work exactly the same, due to their nature they will be much more expensive for the compiler to typecheck.
+Note, it is important to use `interface` and not `type` when constructing those types! Although `type`s will work exactly the same, due to their nature they will be (much more expensive for the compiler to typecheck)[https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections].
+
+For further performace tips, read the (official TypeScript performance wiki)[https://github.com/microsoft/TypeScript/wiki/Performance].
```
jamonholmgren, line 69:
```suggestion
For further performance tips, read the [official TypeScript performance wiki](https://github.com/microsoft/TypeScript/wiki/Performance).
```

jamonholmgren (changes requested):
Just small tweaks needed.

docs/tips/typescript.md
elektronik2k5: Oops, I swapped the braces and brackets :sweat_smile:

jamonholmgren (approved):
Thank you @elektronik2k5 !

PR #1616, opened 2020-11-30
Point link to Patches to documentation

cmlarsen:
Previously the link would dump you on to the GH page.  This takes you directly to the docs.


jamonholmgren (approved):
Thanks @cmlarsen !

PR #1510, opened 2020-04-22
Change castToReference to castToReferenceSnapshot

Tucker-Eric:
Change `castToReference` to `castToReferenceSnapshot` for docs.

PR #1619, opened 2020-12-09
docs: castToReferenceSnapshot instant of castToReference

akaguny:
castToReference function not exist in the package.
https://mobx-state-tree.js.org/API/#casttoreferencesnapshot


akaguny:
close for #1510

Issue #1648, opened 2021-01-19
My model returns fields as null, if I created model fields with types.maybeNull and use model in types.union


KislyakovDS:
**_Bug report_**

* [x] I've checked documentation and searched for existing issues
* [x] I've made sure my project is based on the latest MST version
* [x] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

**Sandbox link or minimal reproduction code**
https://codesandbox.io/s/sad-dan-0nhji

**Describe the expected behavior**
I have a model with types.maybeNull field. Then I use it model in another model with types.union and types.late (circular dependencies). I create store with snapshot and expect store will contains all data, but it ignores some snapshot data and show fields as null.
I also asked this question in spectrum.chat
https://spectrum.chat/mobx-state-tree/general/i-dont-understand-behavior-of-types-maybenull-with-types-late~99969a6d-b59f-40be-98c9-8db9ab1a152d

Model created with snapshot:
```
const group = Group.create({
  name: "New group",
  priority: 1,
  filter: {
    filters: [
      {
        filters: [
          { field: "chat_channel", value: "some channel", operator: "=" },
          { field: "eq_utm_term", value: "11121", operator: "=" }
        ],
        condition: "and"
      },
      {
        filters: [
          { field: "tags", value: "some tag 1", operator: "=" },
          { field: "tags", value: "some tag 2", operator: "=" },
          { field: "tags", value: "some tag 3", operator: "=" }
        ],
        condition: "or"
      }
    ],
    condition: "or"
  }
});
```

I expected group instance will equal to my snapshot.
If in SingleFilter model remove types.maybeNull it will work as expected.

**Describe the observed behavior**
Group instance is equal to:
```
{
  name: "New group",
  priority: 1,
  filter: {
    field: null,
    operator: null,
    value: null,
  },
}
```

willrax:
I'd say we need to write a failing test for this behaviour then try to figure out what exactly is causing it.

coolsoftwaretyler:
Hey folks - looks like @willrax came up with a PR to close this issue out. Hopefully we can clean up CI for that PR and close this issue. 

For now, I'm going to drop a label here that it has a PR that would fix the problem.

coolsoftwaretyler:
I misunderstood that PR - it was just adding a failing test. I'm going to leave the label, but we still are open to an actual fix. 

[...]

PR #1652, opened 2021-01-21
fix up snapshot applying to maybeNull Array union

willrax:
Adds a failing test for issue #1648 

Next steps are to figure out whats up. I don't think this is a great test but it fails in the right way for now. We'll improve it as we figure out exactly whats broken here.

Might not even be the right place to add the test!

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/union.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/union.test.ts
@@ -180,6 +180,58 @@ test("961 - apply snapshot to union should not throw when union keeps models wit
     applySnapshot(u, getSnapshot(Bar.create()))
 })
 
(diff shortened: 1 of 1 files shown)


jamonholmgren:
Good call. Now we need a PR to fix the failing test!

coolsoftwaretyler:
Hey @willrax - are you still interested in bringing this MST PR to conclusion? If so, I'm happy to help you get CI passing. If not, let us know and we can perhaps find a way to take this to the finish line. 

Really appreciate your initial effort here, once way or another!

coolsoftwaretyler:
Hey @willrax - thanks again for putting together this PR. It's a useful step towards resolving the issue, but I'm trying to clean up and close out some stale PRs, so I'm going to close this for now.

I've added this PR to the issue for posterity, so folks will have a good starting point if anyone wants to pick it up.

PR #1656, opened 2021-01-28
mobx 6.1.0 compatibility

### LINKED ISSUE #1653: Error with mobx 6.1.0
**_Bug report_**

no error

i got this error after updating to Mobx 6.1
`Error: [MobX] Cannot apply '@action' to 'ComplexType.prototype.create':
The field is already decorated with '@action'.
Re-decorating fields is not allowed.
Use '@override' decorator for methods overriden by subclass.`

i think this error is caused by MST, no?

urugator:
Fixes #1653

Replaced `makeObservable` by manual creation of actions/computed - a bit inconvinient DX wise, but it should perform a lot better.

### DIFF
--- a/packages/mobx-state-tree/src/core/node/object-node.ts
+++ b/packages/mobx-state-tree/src/core/node/object-node.ts
@@ -1,5 +1,5 @@
 // noinspection ES6UnusedImports
-import { action, computed, reaction, _allowStateChangesInsideComputed, makeObservable } from "mobx"
+import { action, computed, IComputedValue, reaction, _allowStateChangesInsideComputed } from "mobx"
 import {
     addHiddenFinalProp,
     ComplexType,
@@ -15,7 +15,6 @@ import {
     IJsonPatch,
     IMiddleware,
     IMiddlewareHandler,
-    invalidateComputed,
     IReversibleJsonPatch,
     NodeLifeCycle,
     resolveNodeByPathParts,
@@ -117,6 +116,7 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
     private _initialSnapshot: C
     private _cachedInitialSnapshot?: S
     private _cachedInitialSnapshotCreated = false
+    private _snapshotComputed: IComputedValue<S>
 
     constructor(
         complexType: ComplexType<C, S, T>,
@@ -126,7 +126,7 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
         initialValue: C
     ) {
         super(complexType, parent, subpath, environment)
-        makeObservable(this)
+        this._snapshotComputed = computed<S>(() => freeze(this.getSnapshot()))
 
         this.unbox = this.unbox.bind(this)
 
@@ -177,7 +177,6 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
         }
     }
 
-    @action
     createObservableInstance(): void {
         if (devMode()) {
             if (this.state !== NodeLifeCycle.INITIALIZING) {
@@ -232,7 +231,7 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
 
         // NOTE: we need to touch snapshot, because non-observable
         // "_observableInstanceState" field was touched
-        invalidateComputed(this, "snapshot")
+        ;(this._snapshotComputed as any).trackAndCompute()
 
         if (this.isRoot) this._addSnapshotReaction()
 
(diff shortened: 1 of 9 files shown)


packages/mobx-state-tree/src/core/node/object-node.ts
```diff
@@ -344,9 +343,8 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
     private _snapshotUponDeath?: S
 
     // advantage of using computed for a snapshot is that nicely respects transactions etc.
-    @computed
     get snapshot(): S {
-        return freeze(this.getSnapshot())
+        return this._snapshotComputed.get()
```
EmilTholin, line 347: This is a great PR @urugator!

I'm don't have a great understanding of the MST codebase, so excuse me if my comments are silly.
Could it be of value to use `makeObservable(this, { snapshot: computed });` in the constructor instead? [`makeObservable` seems to be pretty cheap with given annotations](https://github.com/mobxjs/mobx/blob/28f8a11d8b94f1aca2eec4ae9c5f45c5ea2f4362/packages/mobx/src/api/makeObservable.ts#L26).
If not, maybe we should remove the comment above the method?

packages/mobx-state-tree/src/core/type/type.ts
```diff
@@ -520,6 +518,7 @@ export abstract class SimpleType<C, S, T> extends BaseType<C, S, T, ScalarNode<C
         return null
     }
 }
+BaseType.prototype.create = action(BaseType.prototype.create)
```
EmilTholin, line 521: Maybe this row should be placed directly below the `BaseType` class?

urugator: Is it not? These classes are so long I had a hard time locating where they end :)

EmilTholin: This is below `SimpleType`, I believe! The `BaseType` class definition ends at line ~370.

packages/mobx-state-tree/src/core/node/object-node.ts
urugator: Every time a constructor is called (which is quite often a suppose) it has to:
[create observable object administration](https://github.com/mobxjs/mobx/blob/28f8a11d8b94f1aca2eec4ae9c5f45c5ea2f4362/packages/mobx/src/types/observableobject.ts#L640-L668)
`Reflect.ownKeys` on passed object 
[find closest getter in proto chain](https://github.com/mobxjs/mobx/blob/28f8a11d8b94f1aca2eec4ae9c5f45c5ea2f4362/packages/mobx/src/types/computedannotation.ts#L21-L37)
[define new computed property](https://github.com/mobxjs/mobx/blob/28f8a11d8b94f1aca2eec4ae9c5f45c5ea2f4362/packages/mobx/src/types/observableobject.ts#L439-L491)
VS 
creation of single computed value

Your call :)

> If not, then maybe we should remove the comment above the method?

Remove? or move above the `computed` creation? I left it there as there is similar comment right after near `getSnapshot()`. It seemed better to keep it near the exposed part as that's what consumers see first.

packages/mobx-state-tree/src/core/type/type.ts
urugator: Good call, should be fixed, thanks!

packages/mobx-state-tree/src/core/node/object-node.ts
EmilTholin: Yeah, creation of a single computed value per instance is definitely cheaper, for sure. What I was trying to get at was the the `snapshot` getter will technically not be a computed anymore, right? Maybe I'm nitpicking.

urugator: Depens how you look at it, the getter is always just a getter that delegates to an underlying computed value (normally exists on ObservableObjectAdministration).
However - and I can be wrong about this as I am not very familar with MST - I think the gist of the comment is to point that the "snapshot" retrieved via this getter will be *computed* - cached, reactive etc.
Obviously you could remove the getter and expose the computed directly, but that would mean to change the API.

EmilTholin:
This change released in `5.0.1` decreased instantiation time by 20% in some production benchmarks. We are lucky to have a MobX pro like you @urugator here!

jamonholmgren:
Released as 5.0.1 -- and based on some preliminary benchmarks, this speeds up instantiation by about 20%! If true, this is very cool.

If anyone notices any problems with 5.0.1, please file a new issue and mention #1656 (this PR) in your description.

https://github.com/mobxjs/mobx-state-tree/releases/tags/v5.0.1

Issue #726, opened 2018-03-22
Pass stores to components via spread operator?


jayarjo:
Is there any reason to why we can't pass store to the component via spread operator?

So instead of:
```jsx
<Component store={store} />
```
do:
```jsx
<Component {...store} />
```
In fact I've actually tried and it injects direct properties only. Would be great to have views and actions in there too.

robinfehr:
one of the reasons for example is that getters would get invoked immediately when youd use the spread operator. you'd therefore only have the plain value in the component instead. observing these wouldn't be possible. but yes one could wrap them again into an observable...

i guess this question can easily lead quite deep into some design decisions which were made in the beginning and could be related to mobx. @mweststrate ping.

pelotom:
I was really hoping this was something MST could do, since MobX can't. What seems to invariably happen if you can't spread a model onto a component's props, is that your components devolve into having a single prop which holds the model object, e.g.

```jsx
class Counter extends React.Component {
  render() {
    return (
      <div>
        <button onClick={this.props.model.onIncrement}>increment</button>
        counter: {this.props.model.count}
      </div>
    );
  }
}
```

It's bad enough having to write `this.props` in front of everything, but this makes the prefix effectively `this.props.model`. Of course, it's possible to destructure everything out,

```ts
const { count, onIncrement } = this.props.model;
```

it's just slightly more annoying 🙂

andraz:
`<Component {...store} />` spreads all props by triggering all getters: `<Component prop1="prop1" prop2="prop2" />`

`<Component {...{store}} />` does equal `<Component store={store} />`

PR #1661, opened 2021-02-15
Warn against using spread syntax with React

davetapley:
This caught me out too, it's a not uncommon pattern in React, so it's worth calling it out for anyone reading the docs.
https://github.com/mobxjs/mobx-state-tree/issues/726


jamonholmgren (approved):

jamonholmgren:
Thanks @davetapley ! Good addition.

PR #1688, opened 2021-04-11
Confusing wording on getTodosWhereDoneIs notice in getting-started

eccentric-j:
Currently on:

https://mobx-state-tree.js.org/intro/getting-started#model-views

> Notice that the getTodosWhereDoneIs view can also be used outside of its model, for example it can be used inside views.

Doesn't make sense to me. Is it trying to say that the view function can be called within other views and View components? Does this mean the get pendingCount() and get completedCount cannot be called in other views?


jamonholmgren (approved):
Makes sense to me, thanks @eccentric-j !

Issue #1607, opened 2020-11-17
Dynamic Import / Lazy Loading of Stores


clgeoio:
**_Feature request_**

**Is your feature request related to a problem? Please describe.**
When using MST, I have run in to issues with tools like `bundlesize` as having a common root store imports references to other stores, which can bring big packages along for the ride.

E.g. I want to add some code which uses compresses images and uploads them to my server. The method lives in the PhotoStore and the package that I am using to do this is 50kb.
With my RootStore as such, now any where else in code that I reference the root store brings the 50kb package too (e.g. the user account password reset page)

```
import { UserStore } from './user';
import { PhotoStore } from './photos';

const Root = types
  .model('Root', {
    user: types.optional(UserStore, {}),
    photos: types.optional(PhotoStore, {}),
  })

const createRootStore = (
  initialState: IRootSnapshotIn,
  dependencies: IDependencies,
) => Root.create(initialState, dependencies);

```

**Describe the solution you'd like**
I would like to be able to lazy import a store so my root store can be smaller in size and users will not get shipped code they do not use. Ideally this will work similar to other code splitting strategies in webpack like [dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports)

e.g.
```
const Root = types
  .model('Root', {
    user: types.lazy(() => import('./user'), { resolveComponent: (module) => module.UserStore }, {}),
    photos:  types.lazy(() => import('./photos'), { resolveComponent: (module) => module.PhotoStore }, {})
  })

const createRootStore = (
  initialState: IRootSnapshotIn,
  dependencies: IDependencies,
) => Root.create(initialState, dependencies);

....

const root = createRootStore(...);
// root.photos == undefined or "late"
root.photos.load()

// root.photos == PhotoStore
```

**Describe alternatives you've considered**
If a route is entirely independent from existing stores, it can live in its own tree and not load the main RootStore at all.

Avoid the use of references, and instead use plain strings and do the lookups manually. There wouldn't need to be a RootStore at all anymore.

**Additional context**
A PR was closed a while back which seemed to [offer this functionality](https://github.com/mobxjs/mobx-state-tree/pull/247)

**Are you willing to (attempt) a PR?**
* [x] Yes
* [] No

clgeoio:
[...]

PR #1722, opened 2021-05-29
Add lazy type for lazy importing of models

clgeoio:
This PR is very much based on: https://github.com/mobxjs/mobx-state-tree/pull/247. 
It adds the `lazy` type, which can be used to dynamically import a model, it relates to the issue https://github.com/mobxjs/mobx-state-tree/issues/1607

This PR still has a couple of todos:
- [x] Write out the documentation (I just copied it from late to make the test happy)
- [ ] Look into the `as unknown as T` casting that is happening in the lazy function. Could there be a better type to use?
- [x] Look into the `any` that is being used in the `shouldLoadPredicate`. It'd be awesome if this could somehow infer the parent model
- [ ] Add some error handling and associated tests


clgeoio:
I'm still thinking about how to handle errors (e.g. if the `loadType()` promise rejects). 
My ideas so far are:
- Have `onLoadError: (parent) => ` option in the `lazy` constructor so that the presence of an error can be communicated to the place where loading occurred.
- Make the `U` generic conform to an interface with `hasError: boolean` so this can be set in via patch.
- Expose the state of the `loadType` promise (`resolved`, `rejected`) 

I expect that a root store may have a number of `lazy` types and to differentiate which stores loaded successfully may be nice

iMakedonsky:
Awesome work!
Can't wait for it to be included in main repo

jamonholmgren:
@mweststrate Do you have thoughts on this one?

packages/mobx-state-tree/src/types/index.ts
```diff
@@ -54,5 +55,6 @@ export const types = {
     late,
     undefined: undefinedType,
     null: nullType,
-    snapshotProcessor
+    snapshotProcessor,
+    lazy
```
diegochavez, line 59: I think should be nice to move this after late to keep consistency?.

jamonholmgren (approved):
I'm okay with merging this. It's self-contained and seems well-written.

coolsoftwaretyler:
Hey @jamonholmgren - last comment you made on this PR was that it's ready to merge. Should we do that? I'm not familiar with our overall release strategy. Do we merge this kind of thing directly into the `master` branch, or does it get on some separate release branch?

I'm guessing this is the PR we had that Twitter comment about.

coolsoftwaretyler:
We have a related MR adding more types in https://github.com/mobxjs/mobx-state-tree/issues/1960. I'm guessing we may get a conflict or two in docs files between this one and that one, but I'm hoping we can merge these in together and ship three new types total in the next minor version.

coolsoftwaretyler:
Spoke with Jamon yesterday and we plan to ship an alpha minor version sometime next week. I'm going to merge this in now and target that release.

Thanks, @clgeoio! Sorry it took us so long here.

PR #1724, opened 2021-06-02
Fix code example for simulated inheritance

cmdcolin:
Hi there!
I saw that potentially the example for doing simulated inheritance was wrong, and doesn't have a parentheses to call the self.surface function

This was also just maybe worth clarifying via some comments to be clear that self.surface and superSurface are functions that need calling instead of being getters

When currently run, the volume returns NaN, so this fixes it (returns 64 if width is 4)

### DIFF
--- a/docs/tips/inheritance.md
+++ b/docs/tips/inheritance.md
@@ -17,6 +17,7 @@ const Square = types
         }
     )
     .views(self => ({
+        // note: this is not a getter! this is just a function that is evaluated
         surface() {
             return self.width * self.width
         }
@@ -26,7 +27,9 @@ const Square = types
 const Box = Square
     .named("Box")
     .views(self => {
-        // save the base implementation of surface
+        // save the base implementation of surface, again, this is a function.
+        // if it was a getter, the getter would be evaluated only once here
+        // instead of being able to evaluate dynamically at time-of-use
         const superSurface = self.surface
 
         return {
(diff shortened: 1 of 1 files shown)


cmdcolin:
Here is a codesandbox with both types.compose and types.union examples https://codesandbox.io/s/blue-dawn-wrgvj?file=/src/App.js

I think the compose example is generally a bit more like what someone would expect from inheritance, so I could add that example to the docs too, as the compose example in the docs does not have an example of overriding the surface function

makarov-roman:
@cmdcolin Is there any reason we can't use getter here?

cmdcolin:
@makarov-roman good question, the reason is "superSurface" gets evaluated only once if it is a getter here, so then the results are reported incorrectly

See https://codesandbox.io/s/blue-dawn-wrgvj?file=/src/App.js

We tried to use the derived getters plus changing runtime behavior, as in the DerivedWithGetters example, in our own codebase but started doing weird workarounds so I reevaluated this docs page and thought this would help clarify

jamonholmgren (approved):
This is probably the best way to approach this, yes.

I can't help but think it's a little _too_ contrived. Like, do we even need to talk about this at all?

But since it's there, this is a superior way to handle it. Thanks @cmdcolin !

cmdcolin:
do you have an alternative that you would recommend? I guess the only thing I see is using only composition but sometimes getting some "inherited" behavior is nice

cmdcolin:
(thanks for review+merger also:))

PR #1755, opened 2021-08-10
docs - Fixed typo in faq

colbygarland:
There is a small typo in the FAQ section (loose instead of lose)


jamonholmgren (approved):
Good catch, thanks!

PR #1784, opened 2021-09-15
Fix array reconciliation with snapshotPreprocessor

### LINKED ISSUE #1776: Arrays are not reconciled correctly with snapshotProcessor instances
**_Bug report_**

https://codesandbox.io/s/wispy-tree-5n1e7

Updating `modelsWithSP` with the same snapshot (as in the provided example) should not trigger the `autorun` reaction.

Every time `modelsWithSP` is updated with the identical snapshot, the `autorun` reaction is triggered. This is because the array reconciliation manually compares the identifier of array items, without taking snapshot processors into consideration.

This can be seen in the array's `areSame` function, which assumes the identifier will be present under the same property in the `newValue` input (which is the unprocessed snapshot):
https://github.com/mobxjs/mobx-state-tree/blob/fb4f3b209b52cee092a640209502a25facf1249a/packages/mobx-state-tree/src/types/complex-types/array.ts#L485

adamkovalsky:
Adds an `isMatchingSnapshotId` method to `ComplexType`, to be used instead of manually comparing identifier values between an existing node and a new snapshot.

Overrides the `isMatchingSnapshotId` method with the `SnapshotProcessor` type to run the preprocessor before attempting to get the identifier value from the new snapshot.

Patches the `is` method on the sub-type of the snapshotProcessor so that the specified snapshot is preprocessed before validation.

Closes #1776

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/snapshotProcessor.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/snapshotProcessor.test.ts
@@ -1,4 +1,4 @@
-import { types, getSnapshot, unprotect, cast, detach, clone } from "../../src"
+import { types, getSnapshot, unprotect, cast, detach, clone, SnapshotIn, getNodeId } from "../../src"
 
 describe("snapshotProcessor", () => {
     describe("over a model type", () => {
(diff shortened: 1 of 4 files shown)


Alaev (approved):

jamonholmgren:
This looks really good to me, @adamkovalsky. Do we need to update any documentation around this? Would you characterize this as a patch-level bugfix, as it appears to be to me?

adamkovalsky:
@jamonholmgren I don't think any documentation needs to be updated, this just brings the functionality in line with expected behavior. I doubt anyone specifically depends on the previously broken behavior. A patch-level bugfix sounds right to me!

jamonholmgren (approved):

jamonholmgren:
Released in 5.0.3: https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.0.3

PR #1785, opened 2021-09-17
docs: make the maybe type description explicit about undefined being used to represent nullability

GR34SE:
I have added an explicit description to the `maybe` type about the value `undefined` being used to represent nullability, so it matches https://mobx-state-tree.js.org/API/index#maybe and the typescript definitions.


elektronik2k5 (approved):

jamonholmgren (approved):
Thanks -- this make sense to me.

PR #1786, opened 2021-09-18
Update dev dependencies

jamonholmgren:
In this PR, I update all dev dependencies (and a few dependencies in the secondary packages).

I was not able to fully update:

* TypeScript (v4), because of a couple test failures
* Jest (v27 doesn't play well with us)
* TypeDoc (v0.20 doesn't play well with us)

Otherwise, tests pass, so I think we're good to merge.


jamonholmgren:
Released in 5.0.3: https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.0.3

PR #1792, opened 2021-09-20
Fix union reconciliation with snapshotPreprocessor

### LINKED ISSUE #1777: snapshotProcessor preProcessor wrongly called with undefined when wrapped with maybe
**_Bug report_**

https://codesandbox.io/s/admiring-allen-i1jsq

The pre-processor should never be called with `undefined`, even though the type is wrapped with `types.maybe` when being used in the store, the model itself should never get `undefined` as input.

The pre-processor gets called with `undefined` as the snapshot causing an error to be thrown.

### LINKED ISSUE #1791: Reconciliation of a snapshotPreprocessor wrapped in a union can fail
**_Bug report_**

https://codesandbox.io/s/weathered-bush-cy907

**Describe the issue**
Reconciliation of a `snapshotProcessor` type which is wrapped in a union (or a `types.maybe` for simplicity) fails because it circumvents the preprocessor and attempts to reconcile the underlying model directly.

This issue was unearthed by PR #1784, and is related to the fix from PR #1047 which shortcuts the type checking of union types to the "current" type when possible, except that the current type is then wrongly reconciled using the underlying model rather than the `snapshotProcessor`. Prior to PR #1784, the type's `is` method wouldn't consider the snapshot to be valid input and skip the shortcut, but that just meant that the fix from #1047 wouldn't work correctly when the types in the union were preprocessed.

adamkovalsky:
Adds a `getReconcilationType` method to `BaseNode`, which returns the type which should be used for reconcilating the node. This is normally the node's `type`, but in the case of `snapshotProcessor` it is overridden to be the `snapshotProcessor` type. This is necessary so that the union type `reconcile` can correctly reconcile the "current" node first if it's possible. This fixes #1791.

Fixes the `snapshotProcessor` type's `isValidSnapshot` and `is` methods to return a validation error and not to throw if the specified preprocessor fails, which can happen when checking if an arbitrary snapshot is valid input (e.g. when determining a union type). We can't expect preprocessors to accept any arbitrary input, so we must assume they can throw when provided with invalid snapshots. This also fixes #1777.

Fixes the array's `areSame` utility to first check that the provided value is a valid snapshot before calling `isMatchingSnapshotId`, to avoid the preprocessor throwing an error if the provided snapshot is invalid.

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/snapshotProcessor.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/snapshotProcessor.test.ts
@@ -646,4 +646,107 @@ describe("snapshotProcessor", () => {
             expect(getNodeId(store.items[0])).toBe(oldNodeId)
         })
     })
+
+    describe("single node reconcilication", () => {
+        test("model with transformed property is reconciled", () => {
+            const SP = types.snapshotProcessor(
+                types.model({
+                    id: types.identifier,
+                    x: types.number
+                }),
+                {
+                    preProcessor(sn: { id: string; y: number }) {
+                        return { id: sn.id, x: sn.y }
+                    }
+                }
+            )
+            const Store = types.model({ item: SP }).actions((self) => ({
+                setItem(item: SnapshotIn<typeof SP>) {
+                    self.item = cast(item)
+                }
+            }));
+            const store = Store.create({ item: { id: "1", y: 0 } });
+            const oldNodeId = getNodeId(store.item)
+            store.setItem({ id: "1", y: 1 })
+            expect(getNodeId(store.item)).toBe(oldNodeId)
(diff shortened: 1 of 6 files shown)


adamkovalsky:
@jamonholmgren I'm not sure why this fails to build on CircleCI, locally it seems to build correctly and all tests pass.

jamonholmgren:
Merged in your change to CI, @adamkovalsky, hopefully that'll do it. Will review this in the meantime!

jamonholmgren (approved):
Thanks @adamkovalsky ! Appreciate the great PRs you've been sending.

adamkovalsky:
Thank you @jamonholmgren, any plans on a patch release any time soon? This issue keeps creeping up on us.

PR #1795, opened 2021-09-23
Fix grammatical error in types documentation

MichaelKilbane:
Should be "than" since it's a comparison.


jamonholmgren:
Thank you!!

PR #1796, opened 2021-09-23
Add includeHooks flag to UndoManager middleware

andrezzoid:
It adds the possibility to receive an `includeHooks` flag in the environment data so the UndoManager can record any changes made inside lifecycle hooks.


jamonholmgren (approved):
This makes a lot of sense. Nice work @andrezzoid!

jamonholmgren:
Released in https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.1.0

PR #1802, opened 2021-09-29
refactor(mobx-state-tree): use reduce instead of forEach

rluvaton:


rluvaton:
It turns out to look more complicated than before 🤨, WDYT?

I changed `.forEach` to `reduce` as this is a classical `.reduce` usage

jamonholmgren (approved):

jamonholmgren:
Nice. I like reduce.

jamonholmgren:
Released in https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.1.0

PR #1822, opened 2021-10-27
Prevent snapshot being preprocessed twice

adamkovalsky:
Fixes a regression introduced in https://github.com/mobxjs/mobx-state-tree/pull/1792 where a snapshot preprocessor gets called on an already preprocessed snapshot, resulting in potential errors.

This issue was occurring because `isMatchingSnapshotId` on the underlying model was proxied to run the preprocessor, but it was being called by `reconcile` with an already preprocessed snapshot.

This fix removes the proxying of `is` and `isMatchingSnapshotId` on snapshot processor's underlying type, while fixing the array's `areSame` utility to rely on `getReconciliationType` to get the underlying node's type. This is necessary because the node stored in the array is the underlying model, while the child-type is the `snapshotProcessor` which is needed to correctly match unprocessed snapshots to processed nodes.

The cast to `any` in `areSame` is needed because the `SnapshotProcessor` type doesn't inherit from `ComplexType`, so there's no easy way to get proper type safety on the call, but we can safely assume that the type is either a `ComplexType` or a `SnapshotProcessor` if the node is an `ObjectNode`.

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/snapshotProcessor.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/snapshotProcessor.test.ts
@@ -609,6 +609,10 @@ describe("snapshotProcessor", () => {
                 }),
                 {
                     preProcessor(sn: { id: string; y: number }) {
+                        if ("x" in sn) {
+                            // Ensure snapshot don't run through preprocessor twice
+                            throw new Error("sn has already been preprocessed")
+                        }
                         return { id: sn.id, x: sn.y }
                     }
                 }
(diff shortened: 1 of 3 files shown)


bengry:
Any news on this by any chance? I'm keeping off of 5.0.4 because of a bug this PR fixes (see https://github.com/mobxjs/mobx-state-tree/discussions/1818#discussioncomment-1543830) and it seems to be getting stale.

jamonholmgren (approved):
Looks good, cutting a new patch release!

jamonholmgren:
Released as 5.0.5: https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.0.5

Issue #1846, opened 2021-12-23
Documentation for volatile does't mention that volatile expects real variables


selvan:
While wrting code, moving a variable inside ```model``` worked, not when moved to ```volatile```.  After spending good chunk of time stumbed upon [the comment](https://github.com/mobxjs/mobx-state-tree/issues/1232#issuecomment-475750805) that says  ```basically volatile expects real variables, (a real string, a real array, etc), so the way to create a "real variable" out of a type is to call create on the type.```

Adding this important piece of information to the [documentation](https://mobx-state-tree.js.org/concepts/volatiles) would help others in the same situation.

coolsoftwaretyler:
Oh interesting, this issue never got closed out when the PR was merged. Closing now. Thanks, @selvan!

PR #1847, opened 2021-12-23
Documentation update - volatiles.md

selvan:
Fix: volatiles.md doesn't mention about volatile expects real variables

### DIFF
--- a/docs/concepts/volatiles.md
+++ b/docs/concepts/volatiles.md
@@ -135,4 +135,4 @@ The object that is returned from the `volatile` initializer function can contain
 5.  Volatile props are preserved during the lifecycle of an instance. See also [reconciliation](reconciliation)
 6.  Changes in volatile props won't show up in the patch or snapshot stream
 7.  It is currently not supported to define getters / setters in the object returned by `volatile`
-8.  volatile expects real variables, (a real string, a real array, etc). Calling `create` on the types, will generate real variables.
+8.  Volatile props expect real variables, (a real string, a real array, etc). Calling `create` on the types, will generate real variables.
--- a/docs/concepts/volatiles.md
+++ b/docs/concepts/volatiles.md
@@ -135,3 +135,4 @@ The object that is returned from the `volatile` initializer function can contain
 5.  Volatile props are preserved during the lifecycle of an instance. See also [reconciliation](reconciliation)
 6.  Changes in volatile props won't show up in the patch or snapshot stream
 7.  It is currently not supported to define getters / setters in the object returned by `volatile`
+8.  volatile expects real variables, (a real string, a real array, etc). Calling `create` on the types, will generate real variables.


selvan:
PR for issue  [#1846] (https://github.com/mobxjs/mobx-state-tree/issues/1846)

docs/concepts/volatiles.md
```diff
@@ -135,3 +135,4 @@ The object that is returned from the `volatile` initializer function can contain
 5.  Volatile props are preserved during the lifecycle of an instance. See also [reconciliation](reconciliation)
 6.  Changes in volatile props won't show up in the patch or snapshot stream
 7.  It is currently not supported to define getters / setters in the object returned by `volatile`
+8.  Volatile props expect real variables, (a real string, a real array, etc). Calling `create` on the types, will generate real variables.
```
elektronik2k5, line 138: Thanks for the PR! Personally, I don't understand what "real" means in this context. Real as opposed to what? And how are they real? I actually have more questions now than before :)

selvan: Yep, agree. How about ```
Volatile props expect data, not types. They are not directly associated with any particular type, and they can hold values of all types.
```
Is it clear?, Kindly feel free to suggest. Thanks.

Comment (elektronik2k5) [reply docs/concepts/volatiles.md:138]: How about "Volatile prop values aren't limited to MST's types and can be assigned any value. This includes [JS primitives](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) such as `string`, `number`, `Symbol`, [Object types](https://developer.mozilla.org/en-US/docs/Glossary/Object) such as `Function`, POJOs, classes - and platform API's such as `localStorage`, `window.fetch` and basically anything you want."

Comment (selvan) [reply docs/concepts/volatiles.md:138]: Looks good,  Did you intend ```Volatile prop values aren't limited to values of MST's types``` instead of ```Volatile prop values aren't limited to MST's types``` ?

Volatile props doesn't care about MST's type or any other type. Volatile props are like javascript variable that hold values of all types.

We are trying to make things clear for us and others. Kindly suggest. Thanks.

Comment (elektronik2k5) [reply docs/concepts/volatiles.md:138]: Yeah, that's better :)

**Comment by selvan** [comment]
Have updated the PR @elektronik2k5

File: docs/concepts/volatiles.md [hunk 2]
```diff
@@ -135,3 +135,4 @@ The object that is returned from the `volatile` initializer function can contain
 5.  Volatile props are preserved during the lifecycle of an instance. See also [reconciliation](reconciliation)
 6.  Changes in volatile props won't show up in the patch or snapshot stream
 7.  It is currently not supported to define getters / setters in the object returned by `volatile`
+8.  Volatile prop values aren't limited to values of MST's types and can be assigned any value. This includes [JS primitives](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) such as string, number, Symbol, [Object](https://developer.mozilla.org/en-US/docs/Glossary/Object) types such as Function, POJOs, classes - and platform API's such as localStorage, window.fetch and basically anything you want.
```
elektronik2k5, line 138:
```suggestion
8.  Volatile prop values aren't limited to values of MST's types and can be assigned any value. This includes [JS primitives](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) such as `string`, `number`, `Symbol`, [Object types](https://developer.mozilla.org/en-US/docs/Glossary/Object) such as `Function`, POJOs, classes - and platform API's like `localStorage`, `window.fetch` and basically anything you want.
```

elektronik2k5 (approved):

jamonholmgren:
Thanks @selvan and @elektronik2k5 !

Issue #1817, opened 2021-10-21
Support PromiseLike with `flow()`


kav:
**_Feature request_**

**Is your feature request related to a problem? Please describe.**
When using supabase with MST and flow the flow type seems to get hung up on supabase's use of PromiseLike rather than Promise for their postgrest-js library
https://github.com/supabase/postgrest-js/blob/06946baf852b42f04b55e10ec900216f5ceeada7/src/lib/types.ts#L51

**Describe the solution you'd like**
Replace the use of Promise in flow() with PromiseLike, or perhaps add an alternative flow function? Maybe consider extending the generic to include the promise type? Not really sure what's best here happy to take any suggestions.

**Describe alternatives you've considered**

Wrap the PromiseLike in a Promise or patch the package directly. Encourage supabase to move from PromiseLike to the full Promise interface.

**Additional context**

**Are you willing to (attempt) a PR?**
* [x] Yes
* [ ] No

10000multiplier:
@kav Sir, I beg you to release this PR. 
![image](https://user-images.githubusercontent.com/57633368/147803934-a53903e1-b8d0-42af-93cf-6d4573b67784.png)

10000multiplier:
@kav just tried applying your fix with patch-package and still getting hung on how to make it work with supabase.

Would you be so kind as to post a little code snippet of what works for you?

kav:
Here's an example of how we were using it in an action:
```typescript
getCategories: flow(function* () {
  const { supabase } = self.environment;
  const { data, error }: { data: CategorySnapshot[]; error: Error } = yield supabase
    .from('categories')
    .select()
    .order('sortkey');
  if (error) throw error;
  // save or return or what have you
  ...
```

jamonholmgren:
Fixed in 5.1.1 -- thanks @kav!

https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.1.1

djMax:
This really seems like it should've been a major version bump. It breaks existing implementations quite soundly.

EmilTholin:
I'm not sure the consequences of this change has been thought through all the way. `PromiseLike` only has `then`, [so code like this now throws an error](https://codesandbox.io/s/heuristic-maxwell-zy7c9?file=/src/index.ts):

```ts
import { flow, types } from "mobx-state-tree";

const getNum = (): Promise<number> => {
  return new Promise((resolve, reject) => {
    if (Math.random() < 0.5) {
[...]

PR #1850, opened 2022-01-04
Switching Promise to PromiseLike

kav:
As described in #1817 switching Promise to PromiseLike, in my case to support Supabase's use of the PromiseLike interface.


10000multiplier:
I highly support this commit. Thank you @kav

jamonholmgren:
Thank you @kav !

PR #1887, opened 2022-03-17
Fix: Don't override reconciliation type for unions in snapshotProcessors

### LINKED ISSUE #1849: A union wrapped in a snapshotProcessor causes an infinite recursion when reconciling
**_Bug report_**

https://codesandbox.io/s/hardcore-kare-66kdw?file=/src/index.ts

No error should occur.

A `Maximum call stack size exceeded` error is thrown. This happens because of the way `snapshotProcessor` is implemented, "hijacking" the node's underlying type instead of being an actual type. When a `union` is wrapped in a `snapshotProcessor` and a snapshot is applied, the `union` is triggering the `snapshotProcessor` to check if it's a valid value, and the `snapshotProcessor` in turn triggers the `union` to validate that it can accept the snapshot after it's been processed.

Tucker-Eric:
closes #1849

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/snapshotProcessor.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/snapshotProcessor.test.ts
@@ -782,4 +782,27 @@ describe("snapshotProcessor", () => {
         store.setItem(undefined)
         expect(store.item).toBeUndefined()
     })
+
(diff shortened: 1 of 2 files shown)


jamonholmgren (approved):

jamonholmgren:
Thanks for the PR @Tucker-Eric! Sorry it's taken me so long to merge this. I'll cut a new release.

PR #1911, opened 2022-05-18
chore: added pnpm installation script

cisolarix:


EmilTholin:
Hi @cisolarix!

Thank you for the PR! I think it's safe to say that if you are using pnpm you already know how installing dependencies works in relation to the npm and yarn examples, so I think this addition is not needed.

PR #1932, opened 2022-07-22
Suggest mobx-store-provider for React

davetapley:
Thanks for a great package @jonbnewman ! 🙏🏻


EmilTholin:
Thank you for the PR @davetapley. The library looks promising and nice, but I don't think it should be officially recommended in the documentation.

PR #1936, opened 2022-07-26
Accept readonly arrays as input for array type

### LINKED ISSUE #1676: Array type cannot be instantiated with a read-only array
For example, see the following simple test:
```ts
const ArrayType = types.array(types.string);
const data = ["foo", "bar"] as const;
ArrayType.create(data);
```

This fails with the error: `The type 'readonly ["foo", "bar"]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.ts(2345)`

_Originally posted by @mbest in https://github.com/mobxjs/mobx-state-tree/issues/1314#issuecomment-617906533_

adamkovalsky:
Fixes #1676.

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/array.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/array.test.ts
@@ -561,7 +561,7 @@ test("assigning filtered instances works", () => {
     expect(notDone.every((t) => isAlive(t))).toBe(true)
 })
 
-test("#1676 should accept read-only arrays", () => {
+test("#1676 - should accept read-only arrays", () => {
     const ArrayType = types.array(types.string)
     const data = ["foo", "bar"] as const
     const instance = ArrayType.create(data)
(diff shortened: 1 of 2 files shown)


packages/mobx-state-tree/__tests__/core/array.test.ts
```diff
@@ -560,3 +560,10 @@ test("assigning filtered instances works", () => {
     expect(done.every((t) => !isAlive(t))).toBe(true)
     expect(notDone.every((t) => isAlive(t))).toBe(true)
 })
+
+test("#1676 should accept read-only arrays", () => {
```
EmilTholin, line 564: Awesome, thank you @adamkovalsky!

Mega nitpick: All other issue-fixing tests are denoted by `#issue_num - description`. Could we format this one like this as well?

EmilTholin (approved):

packages/mobx-state-tree/__tests__/core/array.test.ts
adamkovalsky: No problem, done!

jamonholmgren (approved):
Nice, thank you @adamkovalsky!

jamonholmgren:
Released as 5.1.6.

https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.1.6

PR #1952, opened 2022-08-25
call lifecycle hooks after children are created

### LINKED ISSUE #1680: Child objects containing a reference maybe undefined in afterCreate() or afterAttach()
**_Bug report_**

* [x] Fork [this code sandbox](https://codesandbox.io/s/festive-rain-7rtp1) to reproduce.  (execute unit tests)

[https://codesandbox.io/s/festive-rain-7rtp1](https://codesandbox.io/s/festive-rain-7rtp1)

=> Execute the jest unit tests. Second unit test "load DataModel from rootSnapshot" fails

I have a model with 4 levels of data in the following format:

```
export const ItemModel = types.model("ItemModel").props({
  id: types.identifier,
  name: types.string
});
export const ItemContainerModel = types.model("ItemContainerModel").props({
  item: types.safeReference(ItemModel)
});
export const ObjectModel = types.model("ObjectModel").props({
  itemContainer: types.maybeNull(ItemContainerModel)
});
export const DataModel = types
  .model("Data")
  .props({
    item1: ItemModel,
    object1: types.maybe(ObjectModel),
    object2: types.maybe(ObjectModel)
  })
  .actions((self) => {
    return {
      afterCreate() {
        // here self.object2?.itemContainer is DEFINED while self.object1?.itemContainer is UNDEFINED
      }
    };
  });
export const RootModel = types.model("RootModel").props({
  data: DataModel
});
```

I expect in `DataModel.afterCreate()` that  `self.object1.itemContainer` to be defined, like `self.object2.itemContainer`

In `DataModel.afterCreate()` `self.object1.itemContainer` is NOT defined. whereas  `self.object2.itemContainer` IS defined.
I suspected an issue between "late loading" and "references resolving" but I have tried several uses of `types.late()`, and nothing solved it.

If I replace `item: types.safeReference(ItemModel)` with `item: types.string`, it behaves like it should of course (eg: `self.object1.itemContainer` is defined), but `item` is no more a reference ;-) I need it to be a reference.

Please note that `object1` and `object2` are 2 objects with exactly same content, same model.
If I add `object3` with same content, `object1.itemContainer ` is still undefined, but not `object2.itemContainer` nor `object3.itemContainer`.

What is strange is that the concerned object (`root.data.object1` in the second unit test in codesandbox), is well defined AFTER store has been fully created: (after `const root = RootModel.create(rootSnapshot)`call)

Fork [this code sandbox](https://codesandbox.io/s/festive-rain-7rtp1) to reproduce.  (see unit tests)

Note that the issue does not happen if when there is only 3 levels (`data > itemContainer > object1` instead of 4 levels in `root > data > itemContainer > object1`) (See 1st unit test `load DataModel from rootSnapshot.data`)

I'm blocked since days with this issue ;-)

thxxxx in advance !!!

**Edit:**
Same behavior with `afterAttach()`.
To reproduce in [codesandbox](https://codesandbox.io/s/festive-rain-7rtp1), simply replace `afterCreate()` with `afterAttach()` in unit tests.

### LINKED ISSUE #1951: lifecycle hooks cannot always access their children
**_Bug report_**

```javascript
test.only("lifecycle hooks can access their children", () => {
    const events: string[] = []
    function listener(e: string) { events.push(e) }

    const Child = types
        .model("Todo", {
            title: ""
        })
        .actions(self => ({
            afterCreate() {
                listener("new child: " + self.title)
            }
        }))

    const Parent = types
        .model("Parent", {
            child: Child
        })
        .actions(self => ({
            afterCreate() {
                // **This the key line**: it is trying to access the child
                listener("new parent, child.title: " + self.child?.title)
            }
        }))

    const Store = types.model("Store", {
        parent: types.maybe(Parent)
    })

    const store = Store.create({
        parent: {
            child: { title: "Junior" }
        }
    });
    // As expected no hooks are called.
    // The `parent` is not accessed it is just loaded.
    events.push("-")

    // Simple access does a sensible thing
    const parent = store.parent;
    expect(events).toEqual([
        "-",
        "new child: Junior",
        "new parent, child.title: Junior",
    ])

    // Clear the events and make a new store
    events.length = 0;
    const store2 = Store.create({
        parent: {
            child: { title: "Junior" }
        }
    });
    events.push("-")

    // Using resolvePath fails to do the right thing. 
    const child = resolvePath(store2, "/parent/child");
    expect(events).toEqual([
        "-",
        "new parent, child.title: Junior", // <-- fails with "new parent, child.title: undefined"
        "new child: Junior"
    ])
})
```

Parents should be able to access their children in afterCreate. This should be the case regardless of how the model is created or accessed.

In some cases, for example `resolvePath` and `applyPatch` the children are undefined in afterCreate. This happens when the parent and child are created using a snapshot, and then the child is accessed or modified with a path before the parent is accessed

I believe this is a simplified version of what was reported in #1680

The issue has to do with how `ObjectNode#createObservableInstance` finishes creating the parent first and calls the parent's lifecycle methods before the child has been fully created. I think the right solution is to wait to call the lifecycle hooks until both the parent and child are fully created.

scytacki:
Previously, in some cases, the parent lifecycle hooks would be called before the children were created.

Fixes #1951
Maybe fixes #1680

Note: This will change the order of the hooks in some cases. You can see this change when references are accessed by looking at the changes in `object.test.ts`.  The order should be changed in other cases too. I believe the order is still consistent with the intent. Ie afterCreate is always called before afterAttach for a node. And afterAttach should not be called until the parent is created. If you'd like I could add additional tests illustrating this order change.

### DIFF
--- a/packages/mobx-state-tree/src/core/node/object-node.ts
+++ b/packages/mobx-state-tree/src/core/node/object-node.ts
@@ -171,13 +171,13 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
         }
     }
 
-    createObservableInstanceIfNeeded(runHooks = true): void {
+    createObservableInstanceIfNeeded(fireHooks = true): void {
         if (this._observableInstanceState === ObservableInstanceLifecycle.UNINITIALIZED) {
-            this.createObservableInstance(runHooks)
+            this.createObservableInstance(fireHooks)
         }
     }
 
-    createObservableInstance(runHooks = true): void {
+    createObservableInstance(fireHooks = true): void {
         if (devMode()) {
             if (this.state !== NodeLifeCycle.INITIALIZING) {
                 // istanbul ignore next
@@ -208,6 +208,7 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
 
         // initialize the uninitialized parent chain from parent to child
         for (const p of parentChain) {
+            // delay firing hooks until after all parents have been created
             p.createObservableInstanceIfNeeded(false)
         }
 
@@ -239,16 +240,16 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
 
         this.state = NodeLifeCycle.CREATED
 
-        if (runHooks) {
+        if (fireHooks) {
             this.fireHook(Hook.afterCreate)
             // Note that the parent might not be finalized at this point
-            // so afterAttach won't be called in that case
+            // so afterAttach won't be called until later in that case
             this.finalizeCreation()
 
-            // run the hooks of the parents that we created
+            // fire the hooks of the parents that we created
             for (const p of parentChain.reverse()) {
                 p.fireHook(Hook.afterCreate)
-                // This will call afterAttach on the child
+                // This will call afterAttach on the child if necessary
                 p.finalizeCreation()
             }
         }
--- a/packages/mobx-state-tree/src/core/node/object-node.ts
+++ b/packages/mobx-state-tree/src/core/node/object-node.ts
@@ -171,13 +171,13 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
         }
     }
 
-    createObservableInstanceIfNeeded(): void {
+    createObservableInstanceIfNeeded(runHooks = true): void {
         if (this._observableInstanceState === ObservableInstanceLifecycle.UNINITIALIZED) {
-            this.createObservableInstance()
+            this.createObservableInstance(runHooks)
         }
     }
 
-    createObservableInstance(): void {
+    createObservableInstance(runHooks = true): void {
         if (devMode()) {
             if (this.state !== NodeLifeCycle.INITIALIZING) {
                 // istanbul ignore next
@@ -208,7 +208,7 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
 
         // initialize the uninitialized parent chain from parent to child
         for (const p of parentChain) {
-            p.createObservableInstanceIfNeeded()
+            p.createObservableInstanceIfNeeded(false)
         }
 
         const type = this.type
@@ -238,9 +238,20 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
         this._childNodes = EMPTY_OBJECT
 
         this.state = NodeLifeCycle.CREATED
-        this.fireHook(Hook.afterCreate)
 
-        this.finalizeCreation()
+        if (runHooks) {
+            this.fireHook(Hook.afterCreate)
+            // Note that the parent might not be finalized at this point
+            // so afterAttach won't be called in that case
+            this.finalizeCreation()
+
+            // run the hooks of the parents that we created
+            for (const p of parentChain.reverse()) {
+                p.fireHook(Hook.afterCreate)
+                // This will call afterAttach on the child
+                p.finalizeCreation()
+            }
+        }
     }
 
     get root(): AnyObjectNode {
--- a/packages/mobx-state-tree/__tests__/core/hooks.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/hooks.test.ts
@@ -9,7 +9,9 @@ import {
     onSnapshot,
     isAlive,
     hasParent,
-    cast
+    cast,
+    resolvePath,
+    getParent
 } from "../../src"
 
 function createTestStore(listener: (s: string) => void) {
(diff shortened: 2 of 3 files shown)


packages/mobx-state-tree/src/core/node/object-node.ts
```diff
@@ -208,7 +208,7 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
 
         // initialize the uninitialized parent chain from parent to child
         for (const p of parentChain) {
-            p.createObservableInstanceIfNeeded()
+            p.createObservableInstanceIfNeeded(false)
```
scytacki, line 211: I don't think `...IfNeeded` is required here. We know these parents are not observable from the loop above. But I left it that way to minimize the number of changes. 

Perhaps there is a case where calling createObservableInstance on a parent triggers createObservableInstance on a child before this loop calls createObservableInstance itself.

packages/mobx-state-tree/src/core/node/object-node.ts
```diff
@@ -238,9 +238,20 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
...
 
-        this.finalizeCreation()
+        if (runHooks) {
+            this.fireHook(Hook.afterCreate)
+            // Note that the parent might not be finalized at this point
+            // so afterAttach won't be called in that case
+            this.finalizeCreation()
+
+            // run the hooks of the parents that we created
+            for (const p of parentChain.reverse()) {
```
scytacki, line 249: This isn't very efficient. But neither is `parentChain.unshift(parent)` above. I'd assume the depth of the tree is small enough in all cases, that there is no need to optimize this.

jamonholmgren: Will this possibly fire the parent afterCreates multiple times?

jamonholmgren:
@scytacki Thanks for the PR! I'm not as familiar with this part of the code base and will need to review it with the other core team members.

packages/mobx-state-tree/src/core/node/object-node.ts
scytacki: I don't believe it can fire more than once because the use of the `_observableInstanceState`. The `parentChain` only includes the parents with an `_observableInstanceState` of `UNINITIALIZED`. See line 203 above. 

The current node's `_observableInstanceState` is set to `CREATING` immediately when `createObservableInstance` is called. So the current node will not be included in any `parentChain`s of its children.

Additionally parents in the `parentChain` have their observable instance created for the first time on line 212 (It was at this point in the old code that the `afterCreate` hook was called on the parents before the children). So these parents will now have an `_observableInstanceState` of `CREATING`. Therefore any future `parentChain`s will not include these objects and thus their `afterCreate` will not be called a second time.

scytacki:
@jamonholmgren have you had a chance to look at this more yet?  We've been using it successfully in a large MST codebase without issue for about 3 months now.

jamonholmgren:
@scytacki I appreciate you pushing this forward.

What is your response to this comment? https://github.com/mobxjs/mobx-state-tree/issues/1951#issuecomment-1284401374

I don't have a strong opinion on this, but I want to make sure whatever we do (especially if breaking changes are involved) is commonly acceptable to the community.

scytacki:
@jamonholmgren thanks for pointing out that comment I missed it before. I've now responded.

I'm not sure what the project considers a breaking change. The behavior being changed (fixed) here is not something that was documented. So if the "contract" is to maintain what is documented, then this PR is not a breaking change. If the "contract" is to not change any behavior whether it is documented or not, then yes this is a breaking change. 

The ways this changes the behavior are:
1. The order of the hooks are called in a nested tree is no longer the same regardless of how a deep child node is accessed. In the tests of this PR it shows how the order changes if a reference to a deep child node is accessed directly instead of traversing the children from the root. The order after this PR is still consistent with the documentation though, and I'd argue it is also consistent with the reasonable expectation of a developer.
2. afterCreate will always be able to access children. Perhaps a developer was using the fact that afterCreate previously would fail in some cases, and this developer was doing something special when it failed. However this failure was not documented, so maintaining this failure doesn't seem like something that needs to be done.

If you would rather maintain the current occasional failure of `afterCreate` and the order of the existing hooks, I could make a new PR that adds a new hook instead. I'm not sure what the name should be but perhaps something like `afterChildrenAvailable`. As a user of MST this wouldn't be my preference because it would be more complex to remember (and explain to others) about these 2 hooks.

coolsoftwaretyler:
I think since we have another breaking change over in https://github.com/mobxjs/mobx-state-tree/pull/2039, we should give this one the same label and treat it as such.

I wish the exact order of operations had been documented, which would make the question clearer. But as it is, I know from experience that "minor" updates which change lifecycle updates in other packages have given me some nasty errors that were very hard to troubleshoot in production. I would prefer for MobX-State-Tree not to do that to our users, and if we have other breaking changes slated for release at some point, I think we should leverage that opportunity to bundle this change with it.

The added tests here are awesome, and will help answer future questions about this behavior and if/when we break again. Would also love to make sure we incorporate this into the updated documentation.

scytacki:
We'd love to stop using our fork of MST and go back to the official version, so I might see if I can rework the PR to preserve the order the hooks while still fixing the issue. 

But even if I did that, would this be considered a breaking a change because of my 2nd point above?
> afterCreate will always be able to access children. Perhaps a developer was using the fact that afterCreate previously would fail in some cases, and this developer was doing something special when it failed.

coolsoftwaretyler:
Hey @scytacki - @jamonholmgren and I are going to be meeting next week and I can ask him his opinion on the breakage. If you were to touch up this PR, I think we would end up in two possible states: 

1. We all agree that the change doesn't break (assuming it preserves the order of operations overall), but we may be ready to field some questions if people were handling the prior failure states in a special way. In this scenario, we'd be able to ship relatively quickly and get you off of your fork.
2. We come to the conclusion that this modification *is* a breaking change, and should be handled as such. As I mentioned before, we have a handful of incoming breaking changes in open PRs, and I think we have some bigger ones planned for early 2024. We could still merge this change (or rebase) and y'all could [point your dependencies at the specific GitHub commit](https://docs.npmjs.com/cli/v9/configuring-npm/package-json?v=true#git-urls-as-dependencies) for the interim, which would get you off the fork, and we'd be on a path towards v6, where you could get back in the main deployments from MST.

I think either scenario is reason enough to try and get this PR back in working order. If you're not interested in the longer term play here, I totally understand, and Jamon and I could try to source someone else to do the cleanup (or do it ourselves).

One way or another, I think this PR is important, and I'd like to end up merging it in. Would be stoked to have you continue helping out on it, but I am just happy you've been engaged as it is, and we can take on the responsibility of getting this thing closed out in the coming months (fingers crossed for sooner).

scytacki:
Thanks @coolsoftwaretyler. I'll wait to see what you and Jamon decide. 
Preserving the older of hooks might add more code complexity that might not be worth it in the long run, but I'll take a closer look if you decide this would make the PR mergable and releasable this year.

As far as I know the PR is still in working order, so other than trying to preserve the order of the hooks I'm not sure what else I could help with. If there is something else it needs, let me know and I can try to find the time to address that.

coolsoftwaretyler:
Hey @scytacki - Jamon and I spoke yesterday and feel good about merging this in and releasing with the next minor/patch (I think it'll technically be a minor release).

Since this'll be my first time running through the MST release process, I'm going to probably start it as an alpha candidate on the minor version. I plan to do that sometime this coming week.

For now, I'm going to merge this PR, and if you wanna keep an eye out for `5.2.0-alpha.1` [on npm](https://www.npmjs.com/package/mobx-state-tree), that'll have this in there for ya, along with some other updates, fixes, etc.

I'll also be posting in discussions here about it to get feedback from folks. Hopefully with this PR and the other one you sent along, we can get y'all off your fork + improve MST reliability for all.

Really appreciate your time, and that you've stuck around for the long turnaround time on this.

scytacki:
Yay! Thanks for looking at it and merging it in.

I forgot that we have one more addition to MST in our fork so we can't get off it yet, but hopefully we'll get a PR submitted for that soon.

Issue #621, opened 2018-01-24
Middleware cannot abort asynchronous action


d4rky-pl:
Due to the way asynchronous actions are being called, there's no way to skip all asynchronous actions without also skipping synchronous ones.

As [the documentation states](https://github.com/mobxjs/mobx-state-tree/blob/master/docs/middleware.md)

> If next(call) is not invoked by your middleware, the action will be aborted and not actually executed.

and

> Note: for middleware, it is extremely important that next(action) is called for asynchronous actions (call.type !== "action"), otherwise the generator will remain in an unfinished state forever

Unfortunately during the `action` step we have no way of detecting if the action is asynchronous and on `flow_spawn` we cannot abort by skipping the `next(action)` because it makes the action hang indefinitely.

robinfehr:
brainstorming / having a look at the code properly again needed:
- maybe the call/ action needs to contain the information
- maybe decorators could help
- maybe even something different than flow

mweststrate:
Could we introduce a global flag, so that we know in middleware whether a
flow spawned? Eg

```
function middleware(call, next) {
    const res = next(call)
    if (mst.flowSpawned === true) {
      res.then(something)
   }
   return res
}
```

Op wo 21 mrt. 2018 om 14:09 schreef Rob <notifications@github.com>:

robinfehr:
@mweststrate such global flag would only be a closure during the runtime of the flow right?
if yes that would be aligned with the decoratorFactory idea we had.

mweststrate:
Yes. Undefined immediately after running middleware handlers.

Op wo 21 mrt. 2018 17:32 schreef Rob <notifications@github.com>:

d4rky-pl:
I'd be ok with either A) an `async` flag on action or B) extra event (`flow_start`) that happens before generator is spawned. Both would fix the issue perfectly :)

jayarjo:
Was this ever resolved in any way?

jamonholmgren:
Fixed in v5.1.8: https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.1.8

PR #1969, opened 2022-10-05
fix abort flow cause error

kaven276:
in packages/mobx-state-tree/src/core/flow.ts,
before this fix, when middleware call abort(), 
```typescript
     function onFulfilled(res: any) {
                let ret
                try {
                    // prettier-ignore
                    const cancelError: any = wrap((r: any) => { ret = gen.next(r) }, "flow_resume", res)
                    if (cancelError instanceof Error) {
                        ret = gen.throw(cancelError);
                    }
                } catch (e) {
                    // prettier-ignore
                    setImmediateWithFallback(() => {
                        wrap((r: any) => { reject(e) }, "flow_throw", e)
                    })
                    return
                }
                next(ret)
                return
            }

      function next(ret: any) {
                if (ret.done) {
```
gen.next(r) will not be called, and variable "ret" must be undefined, so in function next, ret.done will raise access undefined's property exception.

I design when abort(Error), call `ret = gen.throw(cancelError)`, so ret is normal generator step's result (have .done/.value), it fix the undefined error ` TypeError: Cannot read properties of undefined (reading 'done') return;` , it also cause flow to fin (before, gen is just hung forever)。

see https://github.com/mobxjs/mobx-state-tree/issues/621

### DIFF
--- a/packages/mobx-state-tree/src/core/flow.ts
+++ b/packages/mobx-state-tree/src/core/flow.ts
@@ -120,7 +120,7 @@ export function createFlowSpawner(name: string, generator: Function) {
 
         function wrap(fn: any, type: IMiddlewareEventType, arg: any) {
             fn.$mst_middleware = (spawner as any).$mst_middleware // pick up any middleware attached to the flow
-            runWithActionContext(
+            return runWithActionContext(
                 {
                     ...contextBase,
                     type,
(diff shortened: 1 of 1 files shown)


jamonholmgren:
Thanks for the PR, @kaven276 ! Would you be able to add a regression test?

jamonholmgren (approved):
I'll release without a regression test, as this seems pretty straightforward.

jamonholmgren:
Fixed in v5.1.8: https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.1.8

PR #1986, opened 2022-11-28
Removes deprecation notice for preProcessSnapshot & postProcessSnapshot

jamonholmgren:
`preProcessSnapshot` and `postProcessSnapshot` were deprecated in #1165, but have proven to be very popular in normal usage still (over `types.snapshotProcessor`, the replacement), and developers seem to prefer the ergonomics of them. So this PR un-deprecates them, as we don't intend to remove them.

(There is a bunch of (unavoidable, mostly) prettier cleanup in this file, but I've noted the specific two lines that have been removed.)

cc @xaviergonz if you're interested

### DIFF
--- a/packages/mobx-state-tree/src/types/complex-types/model.ts
+++ b/packages/mobx-state-tree/src/types/complex-types/model.ts
@@ -78,23 +78,22 @@ export interface ModelPropertiesDeclaration {
  *
  * @hidden
  */
-export type ModelPropertiesDeclarationToProperties<
-    T extends ModelPropertiesDeclaration
-> = T extends { [k: string]: IAnyType } // optimization to reduce nesting
-    ? T
-    : {
-          [K in keyof T]: T[K] extends IAnyType // keep IAnyType check on the top to reduce nesting
-              ? T[K]
-              : T[K] extends string
-              ? IType<string | undefined, string, string>
-              : T[K] extends number
-              ? IType<number | undefined, number, number>
-              : T[K] extends boolean
-              ? IType<boolean | undefined, boolean, boolean>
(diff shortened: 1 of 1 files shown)


packages/mobx-state-tree/src/types/complex-types/model.ts
```diff
@@ -212,12 +209,10 @@ export interface IModelType<
         fn: (self: Instance<this>) => { actions?: A; views?: V; state?: VS }
     ): IModelType<PROPS, OTHERS & A & V & VS, CustomC, CustomS>
 
-    /** @deprecated See `types.snapshotProcessor` */
     preProcessSnapshot<NewC = ModelCreationType2<PROPS, CustomC>>(
         fn: (snapshot: NewC) => ModelCreationType2<PROPS, CustomC>
     ): IModelType<PROPS, OTHERS, NewC, CustomS>
 
-    /** @deprecated See `types.snapshotProcessor` */
```
jamonholmgren, line 220: This line and the previous similar line on L215 are the only two relevant changes to this PR.

xaviergonz:
the good thing of types.snapshotProcessor vs the old ones was the superior TS typing, so the deprecation was added so people could slowly move towards that improvement, but if you feel that's not a good reason to deprecate it then sure, undeprecate it.

jamonholmgren:
Helpful to know, @xaviergonz. When (if?) we do a docs cleanup pass, this would be a good thing to note.

jamonholmgren:
Released as 5.1.7: https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.1.7

PR #2044, opened 2023-07-17
Test/add more test coverage for types model

coolsoftwaretyler:
I wrote this PR mostly to learn more about MobX-State-Tree internals - it adds some additional tests around `types.model`.

I personally learned a lot doing this, and I think the test coverage will be good! My intention is to keep doing things like this over the course of the next six months. Should accomplish two goals:

1. Get me more familiar with the library
2. The more test coverage we have about specific behaviors, the clearer it will be to us when PRs "break" the API. In an ideal state, we'd have appropriate tests to exercise the entire API surface, so that a passing test suite ensures us a changeset isn't breaking, and a failing test suite indicates we may have broken the API contract.

I also added the draft to a blog post. I intend to publish that on my [blog](https://coolsoftware.dev/blog/), but I wanted to include it in this PR for now to get some early feedback on it. I added it as a standalone commit, so we can rebase and drop it pretty easily.

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/model.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/model.test.ts
@@ -1,4 +1,4 @@
-import { applySnapshot, types } from "../../src"
+import { applySnapshot, getSnapshot, types } from "../../src"
 
 test("it should call preProcessSnapshot with the correct argument", () => {
     const onSnapshot = jest.fn((snapshot: any) => {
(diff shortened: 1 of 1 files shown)


packages/mobx-state-tree/__tests__/core/model.md
```diff
@@ -0,0 +1,129 @@
...
+
+I hope you're feeling pretty good about the high level overview so far. This helped me to map out what's going on with my models in MobX-State-Tree. Let's see it in action! I want to write some tests around the behavior I've described.
+
+### Testing names
+
+1. Providing a string as the first argument should set it as the model's name.
+2. Providing an empty string as the first argument should set it as the model's name.
+3. Providing a non-string argument as the first argument should set the model's name as 'AnonymousModel'.
+
+### Testing properties
```
coolsoftwaretyler, line 99: When I pull this out to my own blog, I'll add a link to this PR or the test file I updated.

packages/mobx-state-tree/__tests__/core/model.test.ts
```diff
@@ -17,3 +17,152 @@ test("it should call preProcessSnapshot with the correct argument", () => {
...
+    describe("when we provide an invalid name value, but a valid property object", () => {
+        test("the model will be named AnonymousModel", () => {
+            const Model = types.model(null as any, {
+                prop1: "prop1",
+                prop2: 2
+            })
+
+            expect(Model.name).toBe("AnonymousModel")
+        })
+        test("the model will have no properties", () => {
```
coolsoftwaretyler, line 133: @jamonholmgren - to me, this is possibly unintended behavior? It's probably not a huge deal. For folks using TypeScript, they'll get a type error if they try to:

```ts
const Model = types.Model(null, { prop1: "p1"}) // Redline on the first argument.
```

But for JavaScript consumer, this same code won't give them any TS hints, nor will it error at runtime. But the properties will end up empty.

I think this is a reasonable failure mode - nice that it won't actually toss an exception, but maybe confusing, since we don't even log anything in dev mode.

jamonholmgren: I think we should throw an error for sure; unintended behavior. I see this as a bugfix and not a breaking change.

Where this could come into play is if you end up constructing the model name using a variable and accidentally not getting anything.

```js
let modelName = undefined
if (someCondition) {
  modelName = "HelloWorld"
}

const Model = types.Model(modelName, { prop1: "p1"})
```

If `someCond` is truthy, it'll work as expected. If it's falsy, you'll just get really weird, confusing problems.

jamonholmgren (approved):
Excellent!

packages/mobx-state-tree/__tests__/core/model.test.ts
coolsoftwaretyler: Yeah I agree. I'm going to file this as an issue. Should be an easy fix - I'll tag it as such and save it to see if any new contributors want to pick it up sometime since it's not urgent.

coolsoftwaretyler:
Blog post here: https://coolsoftware.dev/blog/what-happens-when-you-create-an-mst-model/

packages/mobx-state-tree/__tests__/core/model.test.ts
coolsoftwaretyler: Opened an issue: https://github.com/mobxjs/mobx-state-tree/issues/2045

PR #2048, opened 2023-07-21
fix: throw error when creating a model with faulty name and valid pro…

nithinssabu:
…perties


packages/mobx-state-tree/__tests__/core/model.test.ts
```diff
@@ -122,23 +122,34 @@ describe("Edge case behavior", () => {
...
+                expect(modelSnapshot).toEqual({})
+            })
+        } else {
+            test("it should complain about invalid name", () => {
+                expect(() => {
+                    types.model(null as any, {
+                        prop1: "prop1",
+                        prop2: 2
+                    })
+                }).toThrowError("Name should be a string value")
```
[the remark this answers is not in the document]
coolsoftwaretyler: **suggestion:** @nithinssabu - I'd like to test the full error here to make sure we are asserting that we get the `fail()` output as expected (it adds a `[mobx-state-tree]` prefix). This is a small blocker, assuming we make the other blocking change anyway.

packages/mobx-state-tree/__tests__/core/model.test.ts
```diff
@@ -120,25 +120,43 @@ describe("Edge case behavior", () => {
             const modelSnapshot = getSnapshot(Model.create())
             expect(modelSnapshot).toEqual({})
         })
+        if (process.env.NODE_ENV !== "production") {
```
coolsoftwaretyler, line 123: **suggestion:** @nithinssabu - I added a test here to ensure we aren't throwing new errors to people who might be doing some edge-case behavior like instantiating empty models entirely.

Will you consider taking a look for other scenarios, like when a user instantiates a model *only* with a valid object, but no string, and make sure we aren't throwing new errors when we shouldn't?

My intention is mostly to make sure we have some test coverage around: `if (devMode() && typeof args[0] !== "string" && args[1])` which ensures the condition operates as you expect it/will let us know if anyone accidentally modifies it in a way that will break.

If you don't have time or interest to make this change, this is not a blocker for me. I'll add some additional tests later as well.

coolsoftwaretyler (changes requested):
Thanks for the PR! It looks great. I identified one small blocking change on the error message, added some suggestions, and a few thoughts of extra tests we can add, but I don't want to block on those.

packages/mobx-state-tree/src/types/complex-types/model.ts
```diff
@@ -736,6 +736,10 @@ export function model<P extends ModelPropertiesDeclaration = {}>(
  * See the [model type](/concepts/trees#creating-models) description or the [getting started](intro/getting-started.md#getting-started-1) tutorial.
  */
 export function model(...args: any[]): any {
+    if (devMode() && typeof args[0] !== "string" && args[1]) {
+        throw fail("Name should be a string value")
```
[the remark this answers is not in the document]
nithinssabu: Yes, this is much better. I will change it.

All the other suggestions makes sense to me as well and I have the time and interest to make those changes :)

packages/mobx-state-tree/__tests__/core/model.test.ts
nithinssabu: @coolsoftwaretyler Since the behaviour that it should not throw an error when we provide no arguments to the function is valid for both dev and production, do we need the condition `if (process.env.NODE_ENV !== "production")` for this test case?

Thinking about it further, isn't it implicity tested by other cases under the same `describe` block? When we test that the `model will be named AnonymoustModel` and `the model will have no properties`, we are testing that a model is successfully created and it does not throw an error, right?

coolsoftwaretyler: @nithinssabu - you are correct, but one thing I'm hoping to improve in MST is to have a higher quantity of tests at a more granular level of specificity. I want the test suite to function as both a set of automated tests, and a way for us to communicate intent. 

So in this case, I want to specifically communicate externally that in development mode, calling `types.model()` does not throw an error.

coolsoftwaretyler (approved):
Looks good! I'm going to squash, reword the commit, and merge. Thanks for the help @nithinssabu!

PR #2049, opened 2023-07-24
give complex types debug names

BrianHung:
https://github.com/mobxjs/mobx-state-tree/discussions/2011

This PR gives complex types (array, map, and object) [debug names](https://mobx.js.org/observable-state.html#options-). The human-readable names appear in MobX error messages and in the [trace utility](https://mobx.js.org/understanding-reactivity.html). MST already has a method called `describe` on each complex type that we can reuse for the MobX names.

PR #2050, opened 2023-07-25
pass parent to default value function in t.optional

BrianHung:
This PR allows the default value function in [t.optional](https://mobx-state-tree.js.org/API/#optional) to access the parent context via an argument.

This is useful when you want the default value to be a function of the parent state. For example, if we had a user model and wanted to deterministically generate a unique color based on their userName, we could do the following:

```ts
import { types as t } from "mobx-state-tree" 
const User = t.model("User", {
  id: t.identifier,
  userName: t.string,
  color: t.optional(t.string, (snapshot) => hashStringToColor(snapshot.userName))
})

const user = User.create({ id: uuid(), userName: "brian" });
console.log("color", user.color);
```

Right now, there isn't a way to generate the default value based on some state node.

Test cases would probably look like the situation above, plus an additional test case checking that the argument is null when the type is not a direct child of a model type.

```ts
const color = t.optional(t.string, snapshot => {
  console.log("snapshot", snapshot == null)
  return "#000";
});

color.create()
```

### DIFF
--- a/packages/mobx-state-tree/src/types/utility-types/optional.ts
+++ b/packages/mobx-state-tree/src/types/utility-types/optional.ts
@@ -16,7 +16,7 @@ import {
     devMode
 } from "../../internal"
 
-type IFunctionReturn<T> = () => T
+type IFunctionReturn<T> = (parent?: AnyObjectNode) => T
 
 type IOptionalValue<C, T> = C | IFunctionReturn<C | T>
 
--- a/packages/mobx-state-tree/src/types/utility-types/optional.ts
+++ b/packages/mobx-state-tree/src/types/utility-types/optional.ts
@@ -16,7 +16,7 @@ import {
     devMode
 } from "../../internal"
 
-type IFunctionReturn<T> = (parent?: AnyObjectNode) => T
+type IFunctionReturn<T> = (parent: AnyObjectNode | null) => T
 
 type IOptionalValue<C, T> = C | IFunctionReturn<C | T>
 
@@ -88,7 +88,7 @@ export class OptionalValue<
         )
     }
 
-    getDefaultInstanceOrSnapshot(parent?: AnyObjectNode | null): this["C"] | this["T"] {
+    getDefaultInstanceOrSnapshot(parent: AnyObjectNode | null): this["C"] | this["T"] {
         const defaultInstanceOrSnapshot =
             typeof this._defaultValue === "function"
                 ? (this._defaultValue as IFunctionReturn<this["C"] | this["T"]>)(parent)
--- a/packages/mobx-state-tree/src/types/utility-types/optional.ts
+++ b/packages/mobx-state-tree/src/types/utility-types/optional.ts
@@ -61,7 +61,7 @@ export class OptionalValue<
         initialValue: this["C"] | this["T"]
     ): this["N"] {
         if (this.optionalValues.indexOf(initialValue) >= 0) {
-            const defaultInstanceOrSnapshot = this.getDefaultInstanceOrSnapshot()
+            const defaultInstanceOrSnapshot = this.getDefaultInstanceOrSnapshot(parent)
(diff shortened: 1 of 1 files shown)


coolsoftwaretyler:
Hey @BrianHung - kind of like in #2049 - would you mind at least updating the description here to give me/future readers a sense of the purpose of this PR? I can understand the code changes in isolation, but I don't fully get the context of how we might use this change, test it, validate it, etc.

Even an updated description would work, but you know I'll be stoked for a few test cases, some kind of example code, etc., haha ;)

Tag me when you've updated things and I'll take a look!

BrianHung:
@coolsoftwaretyler Thanks for the feedback; hope the edit helps!

coolsoftwaretyler:
Thanks @BrianHung - that's really helpful.

This looks good and CI is passing. I'm going to merge it in. I think it should land in the next minor release, although I'm unsure when that'll be going out.

Really appreciate the PR and the back-n-forth.

chakrihacker:
@coolsoftwaretyler I feel we should not allow users to have a snapshot of the model, inside optional. The correct way is to have a `views` where it can derive the state from the existing model.

chakrihacker:
@coolsoftwaretyler I am talking about this statement

>  if we had a user model and wanted to deterministically generate a unique color based on their userName

This statement means a derivation from existing model data, `views` are used to derive values from existing data

If you are ok I can explain better over a call

coolsoftwaretyler:
@chakrihacker - you're correct that a view is a good use for computed values like that, but I can also see the need for adding parent context for types.optional.

Views don't end up in snapshots, but a types.optional with a default value would.

If we need to we can do a call, but I'd rather discuss async first.

Can you say more specifically what concerns you about this PR? My opinion is it gives tools to developers, but they don't have to use it.

It could also be good to hear from @BrianHung about any other reason that views wouldn't solve his requirements.

chakrihacker:
I was writing test cases for this and had a thought that why haven't we used views for this, so raising question

coolsoftwaretyler:
Gotcha. I think we should wait and let Brian talk a little more about the use case, but I think there are a few scenarios where a view wouldn't solve the problem.

BrianHung:
> This statement means a derivation from existing model data, views are used to derive values from existing data

Color should only be _initially_ derived from the existing username when it is undefined, and thereafter not be derived. That's why it is not a view and makes more sense to be in the optional default value function.

For example,

```tsx
import { types as t } from "mobx-state-tree" 
const User = t.model("User", {
  id: t.identifier,
  userName: t.string,
  color: t.optional(t.string, (snapshot) => hashStringToColor(snapshot.userName))
}).actions(self => ({
  setUserName(name) {
    self.userName = name
  },
  setColor(color) {
    self.color = color
  },
}))
```

chakrihacker:
@BrianHung What about `preProcessSnapShot` doesn't that help your use case??

```ts
import { types as t } from "mobx-state-tree" 
const User = t.model("User", {
  id: t.identifier,
  userName: t.string,
  color: t.optional(t.string, null)
}).preProcessSnapShot(snapshot => ({
  color: snapshot.color ?? hashStringToColor(snapshot.userName)
})).actions(self => ({
  setUserName(name) {
    self.userName = name
  },
  setColor(color) {
    self.color = color
  },
}))
```

BrianHung:
@chakrihacker Maybe the example I gave was too basic, since I agree that `preProcessSnapshot` solves the problem in this example. Let's change the situation and say that `userName` is stored elsewhere in the state tree, or that color should be a function of some sibling or ancestor state in the state tree.

Then you would need `getRoot` or `getParent`.
```tsx
color: t.optional(t.string, (snapshot, parent) => getRoot(parent)...
```

or this PR where you can access the parent node in the preProcessSnapshot to solve it. https://github.com/mobxjs/mobx-state-tree/pull/2065

PR #2051, opened 2023-07-25
Test/add property parsing tests

coolsoftwaretyler:
Adds explicit tests to function how `types.model` processes a given set a properties.

The goal here is not to test the internal `toPropertiesObject` explicitly, but to outline the behavior of MST model instantiation given different types of output, using the `toPropertiesObject` implementation as the current set of expectations.

[Here's a blog post about it, if you're interested](https://coolsoftware.dev/blog/how-does-mobx-state-tree-infer-model-properties/).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ package-lock.json
 .editorconfig
 /test-results/**/*.xml
 /website/build
+.DS_Store
--- a/packages/mobx-state-tree/__tests__/core/model.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/model.test.ts
@@ -1,4 +1,5 @@
 import { applySnapshot, getSnapshot, types } from "../../src"
+import { Hook } from "../../src/internal"
 
 test("it should call preProcessSnapshot with the correct argument", () => {
     const onSnapshot = jest.fn((snapshot: any) => {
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -11,3 +11,4 @@ package-lock.json
 .editorconfig
 /test-results/**/*.xml
 /website/build
+.DS_Store
```
coolsoftwaretyler, line 14: Added this because my Mac finder window sometimes adds it to folders.

packages/mobx-state-tree/__tests__/core/model.test.ts
```diff
@@ -17,152 +18,405 @@ test("it should call preProcessSnapshot with the correct argument", () => {
...
+                expect(modelSnapshot).toEqual({
+                    todo1: {
+                        task: "test"
+                    }
+                })
             })
         })
     })
+    describe("when a user defines a property using a function", () => {
+        // if (process.env.NODE_ENV === "production") {
```
coolsoftwaretyler, line 351: These tests fail, but I don't think they are intended to. My reading of [`toPropertiesObject`](https://github.com/mobxjs/mobx-state-tree/blob/b36e07a86a3b791b115d83983f6ee93143ac0dc5/packages/mobx-state-tree/src/types/complex-types/model.ts#L258) is that we only want to error out if `devMode()` comes back as truthy.

These test cases end up hitting the final `else` clause, which has a comment that reads `// WTF did you pass in mate?` - but really it's just that MST doesn't have a case to handle when a user tries to use a function as a property in production.

I don't think this is really a bug, because we expect developers to hit the failure in development before shipping to prod, but it's not technically accounted for.

@jamonholmgren - do you think we should:

1. Write a case in `toPropertiesObject` to handle functions/objects as props in production where MST doesn't fail, but sets the value to some kind of `undefined` or other falsy value?
2. Remove `devMode()` checks in the function and object cases, so that we can provide more useful errors in production environments
3. Something else?

packages/mobx-state-tree/__tests__/core/model.test.ts
```diff
@@ -17,152 +18,405 @@ test("it should call preProcessSnapshot with the correct argument", () => {
...
+                        foo: () => "bar"
+                    })
+                }).toThrowErrorMatchingInlineSnapshot(
+                    `"[mobx-state-tree] Invalid type definition for property 'foo', it looks like you passed a function. Did you forget to invoke it, or did you intend to declare a view / action?"`
+                )
+            })
+        }
+    })
+    describe("when a user defins a property using a plain JavaScript object", () => {
+        // test("it sets the default value to undefined in production", () => {
```
coolsoftwaretyler, line 386: [Basically same questions as my prior thread, but about objects instead of functions](https://github.com/mobxjs/mobx-state-tree/pull/2051/files#r1272988275) - @jamonholmgren

packages/mobx-state-tree/__tests__/core/model.test.ts
coolsoftwaretyler: In the interest of getting this merged to add more test coverage [I removed the failing specs](https://github.com/mobxjs/mobx-state-tree/pull/2051/commits/9dff9372b4aa208734ffee1c9f623c596350d753). I don't think this is a major pitfall, just a nuance I was interested in. Perhaps we'll revisit!

packages/mobx-state-tree/__tests__/core/model.test.ts
coolsoftwaretyler: In the interest of getting this merged to add more test coverage [I removed the failing specs](https://github.com/mobxjs/mobx-state-tree/pull/2051/commits/9dff9372b4aa208734ffee1c9f623c596350d753). I don't think this is a major pitfall, just a nuance I was interested in. Perhaps we'll revisit!

Issue #2053, opened 2023-07-26
detach can corrupt the identifierCache


scytacki:
**_Bug report_**

-   [x] I've checked documentation and searched for existing issues [and discussions](https://github.com/mobxjs/mobx-state-tree/discussions)
-   [x] I've made sure my project is based on the latest MST version
-   [x] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction.

**Sandbox link or minimal reproduction code**
```
test("items detached from arrays don't corrupt identifierCache", () => {
    const Item = types.model("Item", { id: types.identifier })
    const ItemArray = types.model("ItemArray",
      { items: types.array(Item)})
    .actions(self => ({
        removeSecondItemWithDetach() {
            detach(self.items[1])
        }
    }));

    const smallArray = ItemArray.create({ items: [
        { id: "A" }, { id: "B" }, { id: "C" }, { id: "D" }, { id: "E"}
    ]})
    expect(resolveIdentifier(Item, smallArray, "E")).toBeDefined()
    smallArray.removeSecondItemWithDetach()
    expect(smallArray.items.length).toBe(4)
    expect(resolveIdentifier(Item, smallArray, "B")).toBeUndefined()
    expect(resolveIdentifier(Item, smallArray, "E")).toBeDefined()

    const largeArray = ItemArray.create({ items: [
        { id: "A" }, { id: "B" }, { id: "C" }, { id: "D" }, { id: "E" },
        { id: "F" }, { id: "G" }, { id: "H" }, { id: "I" }, { id: "J" },
        { id: "K" }
    ]})
    expect(resolveIdentifier(Item, largeArray, "K")).toBeDefined()
    largeArray.removeSecondItemWithDetach()
    expect(largeArray.items.length).toBe(10)
    expect(resolveIdentifier(Item, largeArray, "B")).toBeUndefined()
    expect(resolveIdentifier(Item, largeArray, "J")).toBeDefined()
    // ***** The following expectation fails ******
    expect(resolveIdentifier(Item, largeArray, "K")).toBeDefined()
})
```

**Describe the expected behavior**
When an item is removed from an array of identified items the remaining items in the array should still be accessible via resolveIdentifier. They should also be available as references.

**Describe the observed behavior**
[...]

PR #2054, opened 2023-07-26
fix splitCache, extra nodes were being removed

scytacki:
https://github.com/mobxjs/mobx-state-tree/issues/2053

When detach is used to remove nodes from the tree, splitCache is called to pull this node and its children out of the identifierCache of the original tree. The implementation of this splitting had a bug which would remove additional nodes that are not children of the detached node.

The issue was that it was using a string match of the path of the node being removed with all the other nodes in the indentifierCache. So if the node being removed had a path of `/items/1`, then this would match `/items/10`, or `items/199`. 

Additionally this PR removes empty sets in the indentifierCache when that is the result of splitCache. This matches the behavior of `notifiedDied`.

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/identifier.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/identifier.test.ts
@@ -306,5 +306,6 @@ test("items detached from arrays don't corrupt identifierCache", () => {
     expect(largeArray.items.length).toBe(10)
     expect(resolveIdentifier(Item, largeArray, "B")).toBeUndefined()
     expect(resolveIdentifier(Item, largeArray, "J")).toBeDefined()
+    // The following expectation was failing in version 5.1.8 and earlier
     expect(resolveIdentifier(Item, largeArray, "K")).toBeDefined()
 })
--- a/packages/mobx-state-tree/__tests__/core/identifier.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/identifier.test.ts
@@ -2,12 +2,12 @@ import {
     types,
     tryResolve,
     resolvePath,
-    cast,
     SnapshotOrInstance,
     IAnyModelType,
     IAnyComplexType,
     resolveIdentifier,
-    getIdentifier
+    getIdentifier,
+    detach
 } from "../../src"
 
 if (process.env.NODE_ENV !== "production") {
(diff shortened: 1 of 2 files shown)


coolsoftwaretyler:
Thanks, @scytacki! I will take a look at this in a few days. Looks very good at first glance, and I'm guessing y'all have merged it into your fork/are using it, so I'm sure we'll merge it pretty quickly.

Really appreciate you merging it back into MST! Thanks for that!

chakrihacker (approved):

scytacki:
Thanks @coolsoftwaretyler. Yes we have merged it into our fork. We haven't deployed it to our production app yet, but we will likely be doing that this week.

packages/mobx-state-tree/__tests__/core/identifier.test.ts
```diff
@@ -268,3 +268,44 @@ test("identifierAttribute of the type", () => {
     const M3 = types.model({ myId: types.optional(types.identifier, () => "hi") })
     expect(M3.identifierAttribute).toBe("myId")
 })
+
+test("items detached from arrays don't corrupt identifierCache", () => {
```
coolsoftwaretyler, line 272: **praise:** @scytacki - thank you for the detailed test cases here. Really make it clear what's goin' on.

packages/mobx-state-tree/src/core/node/identifier-cache.ts
```diff
@@ -68,23 +68,31 @@ export class IdentifierCache {
         }
     }
 
-    splitCache(node: AnyObjectNode): IdentifierCache {
-        const res = new IdentifierCache()
-        const basePath = node.path
+    splitCache(splitNode: AnyObjectNode): IdentifierCache {
+        const newCache = new IdentifierCache()
+        // The slash is added here so we only match children of the splitNode. In version 5.1.8 and
```
coolsoftwaretyler, line 73: **praise:** @scytacki - excellent catch + makes sense here. Thanks for the comments - this is an important fix, would be easy for us to lose context over time

packages/mobx-state-tree/src/core/node/identifier-cache.ts
```diff
@@ -68,23 +68,31 @@ export class IdentifierCache {
...
+                    }
                     modified = true
                 }
             }
             if (modified) {
                 this.updateLastCacheModificationPerId(id)
             }
         })
-        return res
+        return newCache
```
coolsoftwaretyler, line 95: **praise:** @scytacki - I like this name a little more than `res` here. Nice improvement!

coolsoftwaretyler (approved):
@scytacki - thanks for contributing back! This looks great, makes sense. I'm going to merge it in. Should make it to the next patch or minor release, although I'm not precisely sure when we'll be publishing next. Hopefully soon.

PR #2059, opened 2023-08-09
Allow readonly string arrays in types.enumeration

### LINKED ISSUE #1314: Take a readonly array when possible
**_Bug report_**

```typescript
type TodoType = "task" | "bug";
// written as a separate variable for other references that are not tied to MST
const todoTypes: ReadonlyArray<TodoType> = ["task", "bug"];
const todoTypes: ["task", "bug"] as const // same as above, but shorter, for TS >= 3.4

export const Todo = types
  .model("Todo", {
    id: types.optional(types.number, () => Math.random()),
    title: types.string,
    finished: false,
    type: types.enumeration(todoTypes) // error
  })
```

The above code should compile

Line 10 shows the following error:
> Argument of type 'ReadonlyArray<TodoType>' is not assignable to parameter of type 'TodoType[]'.
  Type 'ReadonlyArray<TodoType>' is missing the following properties from type 'TodoType[]': pop, push, reverse, shift, and 6 more

This is due to `enumeration` accepting T[], and not a `ReadonlyArray<T>`/`readonly T[]` (latter as of TS 3.4).

This behavior is not just for this case, but also in other places in MST (e.g. `types.union`.

My suggestion is to accept the minimal necessary interface for arrays when possible. This should cover most, if not all cases, since they are usually not directly used anyway, at least for types. This is more important with TypeScript 3.4+ since it's now much less of a hassle to make an array a `ReadonlyArray`, especially with `as const`, but also with the [`readonly` modifier added to more use-uses (e.g. function parameters)](https://github.com/Microsoft/TypeScript/pull/29435).

If this is handled, upgrading to TypeScript 3.4 (or greater) should be considered, since it makes the changes easier to type. e.g.:
```typescript
// TS 3.3
export function enumeration<T extends string>(options: ReadonlyArray<T>): ISimpleType<UnionStringArray<T[]>>
export function enumeration<T extends string>(
    name: string,
    options: ReadonlyArray<T>
): ISimpleType<UnionStringArray<T[]>>

// TS 3.4
export function enumeration<T extends string>(options: readonly T[]): ISimpleType<UnionStringArray<T[]>>
export function enumeration<T extends string>(
    name: string,
    options: readonly T[]
): ISimpleType<UnionStringArray<T[]>>
```
At the cost of supporting only TS 3.4+ (maybe we can transpile these to `ReadonlyArray` in the output type declarations?)

jamonholmgren:
## Overview

This PR enhances the enumeration function types to accept both mutable (string[]) and immutable (readonly string[]) string arrays. This allows for better compatibility with `as const` syntax and provides a more flexible API for users.

Fixes #1314.

## Changes

* Updated the type signatures in the enumeration function to handle both `readonly string[]` and `string[]`.
* Added new test cases to verify support for `readonly string[]` as well as maintaining support for `string[]`.

## Impact

These changes enable developers to use as const syntax with the enumeration function without any type errors, providing a smoother development experience.

## Testing

New tests have been added to ensure that both readonly and plain string arrays are handled correctly. Note that tests don't seem to fail even when the typescript changes haven't been implemented, but there is a type error in the test in VS Code that goes away once this is implemented.

CleanShot 2023-08-09 at 15 05 05@2x

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/enum.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/enum.test.ts
@@ -46,3 +46,29 @@ test("should support optional enums inside a model", () => {
     const l = TrafficLight.create({})
     expect(l.color).toBe(ColorEnum.Orange)
 })
+test("should support plain string[] arrays", () => {
+    const colorOptions: string[] = ["Red", "Orange", "Green"]
+    const TrafficLight = types.model({ color: types.enumeration(colorOptions) })
+    const l = TrafficLight.create({ color: "Orange" })
+    unprotect(l)
+    l.color = "Red"
+    expect(TrafficLight.describe()).toBe('{ color: ("Red" | "Orange" | "Green") }')
+    if (process.env.NODE_ENV !== "production") {
+        expect(() => (l.color = "Blue" as any)).toThrowError(
+            /Error while converting `"Blue"` to `"Red" | "Orange" | "Green"`/
+        )
+    }
+})
+test("should support readonly enums as const", () => {
+    const colorOptions = ["Red", "Orange", "Green"] as const
+    const TrafficLight = types.model({ color: types.enumeration(colorOptions) })
+    const l = TrafficLight.create({ color: "Orange" })
+    unprotect(l)
+    l.color = "Red"
+    expect(TrafficLight.describe()).toBe('{ color: ("Red" | "Orange" | "Green") }')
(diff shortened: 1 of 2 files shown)


coolsoftwaretyler:
Hey @jamonholmgren - this looks good to me and makes sense! I'm gonna merge it and include in the second RC for 5.2.0.

PR #2063, opened 2023-08-10
docs: run publish-docs

coolsoftwaretyler:
This is the diff after running `yarn publish-docs` against https://github.com/mobxjs/mobx-state-tree/discussions/2062. 

Perhaps it should be added to that release, but since the auto-generation is deterministic, I think it's fine to leave it separate for now.

This is one of those things we should solve with a good CI flow.

### DIFF
--- a/docs/API/index.md
+++ b/docs/API/index.md
@@ -177,7 +177,7 @@ sidebar_label: "Globals"
 
 Ƭ **IDisposer**: *function*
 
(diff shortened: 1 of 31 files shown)


coolsoftwaretyler:
Ah I think we neglected to update some docs tests when we added new types.

I'm going to leave this for now and perhaps fix tomorrow. Should be a quick one. If anyone else sees this before I get to it, go ahead and fix it up!

PR #2067, opened 2023-08-20
add test cases for optional

### LINKED ISSUE #2056: Add tests and documentation for new parent argument to `types.optional`
In https://github.com/mobxjs/mobx-state-tree/pull/2050, we added an new way for `types.optional` to access their parent context in the function argument.

We should add tests to cover these new cases. The PR author wrote up some possible examples along with context there.

chakrihacker:

Add test cases for optional type, where it can access snapshot of the parent

## Steps to validate locally

Run the test cases

closes #2056

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/optional.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/optional.test.ts
@@ -146,3 +146,27 @@ test("undefined can work as a missing value", () => {
     const m3 = M.create({}) // is ok as well (even in TS)
     expect(m3.x).toBe(undefined)
 })
+
+test("optional can access parent state", () => {
+    const TodoModel = types.model({
+        id: types.identifier,
+        name: types.string,
+        // @ts-ignore
+        priority: types.optional(types.string, (snapshot) => {
+            console.log({ snapshot })
+            if (snapshot) {
+                if (snapshot.name.toLowerCase().includes("today")) {
+                    return "High"
+                }
+            }
+            return ""
+        })
+    })
+
+    const todo1 = TodoModel.create({
+        id: "1",
+        name: "Learn MobX Today"
+    })
+    console.log({ todo: getSnapshot(todo1) })
+    expect(todo1.id).toBe("1")
+})


chakrihacker:
@coolsoftwaretyler Can you help me here, why `snapshot` is `undefined` ??

coolsoftwaretyler:
@chakrihacker - did you run `yarn build` at the top level of the repo before this? `cd packages/mobx-state-tree && yarn jest  -t "access parent state"` is working from me after doing that.

BrianHung:
@coolsoftwaretyler @chakrihacker 

After diving deeper into mobx-state-tree's architecture, I think we should actually revert these two PRs.
https://github.com/mobxjs/mobx-state-tree/pull/2050
https://github.com/mobxjs/mobx-state-tree/pull/2065

I had initially written these PRs to give the `snapshotProcessor` and `optional` types access to the parent context upon initialization, similar to how the `reference` type gives access to [parent in the custom get and set options](https://github.com/mobxjs/mobx-state-tree/blob/94bb7d02de6ed3e2c3f92a6a6ddadc16644597f0/packages/mobx-state-tree/src/types/utility-types/reference.ts#L463-L466).

However, the two PRs above in their current implementations can't access the parent `StateTreeNode`. This is because with MST and lazy initialization, the [parent isn't created](https://github.com/mobxjs/mobx-state-tree/blob/94bb7d02de6ed3e2c3f92a6a6ddadc16644597f0/packages/mobx-state-tree/src/core/node/object-node.ts#L218) until _after_ the [child nodes are initialized](https://github.com/mobxjs/mobx-state-tree/blob/94bb7d02de6ed3e2c3f92a6a6ddadc16644597f0/packages/mobx-state-tree/src/core/node/object-node.ts#L140).

A possible solution would be maybe to optimistically create the parent, which is what [references do](https://github.com/mobxjs/mobx-state-tree/blob/94bb7d02de6ed3e2c3f92a6a6ddadc16644597f0/packages/mobx-state-tree/src/types/utility-types/reference.ts#L317), or give to the parent `ObjectNode` instead where users could call [`getValue` to create the parent](https://github.com/mobxjs/mobx-state-tree/blob/94bb7d02de6ed3e2c3f92a6a6ddadc16644597f0/packages/mobx-state-tree/src/core/type/type.ts#L406-L408).

However, with the first solution, reference types _require_ parents to work while `snapshotProcessors` and `optional` don't have the same expectations; and with the second solution, `AnyObjectNode` is an internal utility currently not exposed (and probably shouldn't).

This doesn't mean there aren't any solutions. As @chakrihacker pointed out in a comment, the solution to needing parent context in [`optional` would either to use a `snapshotProcessor`](https://github.com/mobxjs/mobx-state-tree/pull/2050#issuecomment-1686594091); and if you wanted parent context in either, you could use actions or helper functions.

```ts
.actions(parent => ({
  createChild(props) {
    return ChildType.create({
     ...createProps(props, parent)
    })  
}
}));
```

```ts
function createChildType(props, parent) {
  return ChildType.create({
     ...createProps(props, parent)
  })  
}
```

The main takeaway is, if you need to access the parent node when creating the child type, the child props have to be derived from the parent / outside context first and then passed into create, as Mobx-State-Tree optimistically creates the child snapshot before initializing the parent state tree node.

coolsoftwaretyler:
Hey @BrianHung - thanks so much for staying on top of this and doing the extra work to re-think and catch that!

@chakrihacker - thank you as well for digging in on these changes and prompting all of this work to make sure we keep MST reliable. I think we can close this PR, and I will work on some additional PRs to revert the changes. 

I'll close out some of the related issues as well.

coolsoftwaretyler:
Revert PR incoming: https://github.com/mobxjs/mobx-state-tree/pull/2074

I'll have this out in the next release candidate sometime in a couple hours.

Issue #2055, opened 2023-08-01
Add tests and documentation for debug names in complex types


coolsoftwaretyler:
In https://github.com/mobxjs/mobx-state-tree/pull/2049, we added [debug names](https://mobx.js.org/observable-state.html#options-) to MST complex types (array, map, and object).

The human-readable names appear in MobX error messages and in the [trace utility](https://mobx.js.org/understanding-reactivity.html).

We should add some additional test coverage for this, along with updated documentation reflecting it.

PR #2069, opened 2023-08-23
Add a test for debug name

BrianHung:

Adds a test to https://github.com/mobxjs/mobx-state-tree/pull/2049
Closes https://github.com/mobxjs/mobx-state-tree/issues/2055

Also modifies #2049 to use `this.name` instead of `this.describe`, since describe just iterates over children and their types and doesn't actually use name.

```ts
const Model = types.model("Name", { value: t.number });
const model = Model.create({ value: 5 });

const array = types.array(Model).create()
const map= types.map(Model).create()

// previous
console.log(getDebugName(model)) // { value: number }
console.log(getDebugName(array)) // { value: number }[]
console.log(getDebugName(map)) // map<string,{ value: number }>

// now
console.log(getDebugName(model)) // Name
console.log(getDebugName(array)) // Name[]
console.log(getDebugName(map)) // Map<string, Name>
```

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/late.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/late.test.ts
@@ -31,7 +31,7 @@ test("late should describe correctly circular references", () => {
     const Node = types.model("Node", {
         childs: types.array(types.late((): IAnyModelType => Node))
     })
-    expect(Node.describe()).toEqual("{ childs: Node[]? }")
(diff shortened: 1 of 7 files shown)


coolsoftwaretyler:
Makes sense! Thanks for the tests, @BrianHung. I'll merge this in and it'll make its way to the 5.2.0 release candidate tonight!

PR #2072, opened 2023-08-27
use mobx type for map initial data

BrianHung:

MobX has an exported type `IObservableMapInitialValues` for its `initialData`. Since `MSTMap` extends `ObservableMap`, it makes sense that `MSTMap` should use the same type.

### DIFF
--- a/packages/mobx-state-tree/src/types/complex-types/map.ts
+++ b/packages/mobx-state-tree/src/types/complex-types/map.ts
@@ -10,7 +10,8 @@ import {
     observable,
     ObservableMap,
     observe,
-    values
+    values,
+    IObservableMapInitialValues,
 } from "mobx"
 import {
     ComplexType,
@@ -141,7 +142,7 @@ export enum MapIdentifierMode {
 
 class MSTMap<IT extends IAnyType> extends ObservableMap<string, any> {
     constructor(
-        initialData?: [string, any][] | IKeyValueMap<any> | Map<string, any> | undefined,
+        initialData?: IObservableMapInitialValues<string, any> | undefined,
         name?: string
     ) {
         super(initialData, (observable.ref as any).enhancer, name)
@@ -296,11 +297,11 @@ export class MapType<IT extends IAnyType> extends ComplexType<
             Object.keys(hooks).forEach((name) => {
                 const hook = hooks[name as keyof typeof hooks]!
                 const actionInvoker = createActionInvoker(instance as IAnyStateTreeNode, name, hook)
-                ;(!devMode() ? addHiddenFinalProp : addHiddenWritableProp)(
-                    instance,
-                    name,
-                    actionInvoker
-                )
+                    ; (!devMode() ? addHiddenFinalProp : addHiddenWritableProp)(
+                        instance,
+                        name,
+                        actionInvoker
+                    )
             })
         })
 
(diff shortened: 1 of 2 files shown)


packages/mobx-state-tree/src/types/complex-types/map.ts
```diff
@@ -141,7 +142,7 @@ export enum MapIdentifierMode {
 
 class MSTMap<IT extends IAnyType> extends ObservableMap<string, any> {
     constructor(
-        initialData?: [string, any][] | IKeyValueMap<any> | Map<string, any> | undefined,
+        initialData?: IObservableMapInitialValues<string, any> | undefined,
```
coolsoftwaretyler, line 145: @jamonholmgren - Would a change to TypeScript typings like this technically be a breaking change? I think it's the right move to make, but I"m wondering if we should get some backwards compatibility here.

@BrianHung - what do you think about typing `initialData` with another union that includes `IObservableMapInitialValues`, but allows people to keep using older typings they were before?

BrianHung: Double check me on this but I thought it was backwards compatible?

```ts
export type IObservableMapInitialValues<K = any, V = any> =
    | IMapEntries<K, V>
    | IReadonlyMapEntries<K, V>
    | IKeyValueMap<V>
    | Map<K, V>

export type IMapEntry<K = any, V = any> = [K, V]
export type IReadonlyMapEntry<K = any, V = any> = readonly [K, V]
export type IMapEntries<K = any, V = any> = IMapEntry<K, V>[]
export type IReadonlyMapEntries<K = any, V = any> = IReadonlyMapEntry<K, V>[]
```

https://github.com/mobxjs/mobx/blob/main/packages/mobx/src/types/observablemap.ts#L45-L48

coolsoftwaretyler: Oh nice!

That's exactly the info I needed on this one. Thank you!

Would you mind adding a test to `packages/mobx-state-tree/__tests__/core/map.test.ts` that checks for the typing here and demonstrates it works? You could follow the pattern Jamon did in https://github.com/mobxjs/mobx-state-tree/pull/2059, where he wrote a test that exercised the basic functionality, and if you modify the typing in the source, or modify the input in the test to mismatch the type, TS will throw an error.

coolsoftwaretyler:
So, I expected that `types.map(types.string).create()` would work with:

1. `[string, any][]`
2. `IKeyValueMap<any>`
3. `Map<string, any>`
4. `undefined`

It seems to *only* work with 2 and 4. [See here](https://codesandbox.io/s/autumn-resonance-g2fksm?file=/src/index.ts).

The typings on that `initialData` parameter for the constructor seem to get modified somewhere between the `MSTMap` definition, and exposing `types.map` export. I don't totally see where/how. 

I'm going to write a test that just tests 2/4, and I think it'll work out here. We also don't seem to allow users to call that constructor directly anyway, so I feel good about this change. I'll push a commit and merge.

BrianHung:
Sorry, I've been busy with work; thanks for taking up the effort to write tests ❤️

coolsoftwaretyler:
Happy to! Tryna clean up PRs so I can muck around with the folder structure lol

PR #2077, opened 2023-08-29
bugfix: fix error judgement on splitCache

luo-occ:
## fix the bug of path judgement

inner function of splitCache has error when judging descendant node

```
node1.path='a/1'
node2.path='a/10'

nodes2.path.indexOf(node1) === 0 // true
// then it will assume node2 is the child of node1, which is not expected

```
## fix solution

add additional judgement of path that contains '/'

### DIFF
--- a/packages/mobx-state-tree/src/core/node/identifier-cache.ts
+++ b/packages/mobx-state-tree/src/core/node/identifier-cache.ts
@@ -71,10 +71,11 @@ export class IdentifierCache {
     splitCache(node: AnyObjectNode): IdentifierCache {
         const res = new IdentifierCache()
         const basePath = node.path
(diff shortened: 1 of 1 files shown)


chakrihacker:
Hey @luo-occ Can you provide an example repro for the issue??

coolsoftwaretyler:
Thanks for the PR @luo-occ, and I'm sorry for any confusion, I think this is actually closed by https://github.com/mobxjs/mobx-state-tree/pull/2054, which had a corresponding issue of https://github.com/mobxjs/mobx-state-tree/issues/2053.

I'm going to close this out because I think we're all good on this, but let me know if the issue persists in versions past 5.2.0, please!

PR #2084, opened 2023-09-16
perf: use map instead of weak map

BrianHung:

Uses map instead of weakmap for running actions. Each call has a unique id, so the memory overhead of a weakmap is not needed.

### DIFF
--- a/packages/mobx-state-tree/__tests__/core/actionTrackingMiddleware2.test.ts
+++ b/packages/mobx-state-tree/__tests__/core/actionTrackingMiddleware2.test.ts
@@ -237,3 +237,201 @@ test("#1250", async () => {
     expect(calls).toEqual(["setX (21) <- (undefined) - onFinish (error: false)"])
     calls.length = 0
 })
+
+/**
+ * Test that when createActionTrackingMiddleware2 is called with valid hooks and a synchronous action, it runs onStart and onFinish hooks.
+ */
+test("successful execution", () => {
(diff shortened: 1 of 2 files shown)


coolsoftwaretyler:
Hey @BrianHung - I'm sorry I didn't see this before. I've been really busy at work + I don't seem to get notifications for PRs in this repo? I'm going to turn on my notifications + get around to reviewing this and many of the other open PRs. Sorry to leave you hanging, and thanks for your patience!

BrianHung:
No worries! I always use patch package to use my PRs locally so don’t rush.

coolsoftwaretyler:
Hey @BrianHung - as with the other PRs, would you mind adding a few tests around this, even if we already have some? Looks like we have a test file in `packages/mobx-state-tree/__tests__/core/actionTrackingMiddleware2.test.ts` - it would be great to add coverage or some redundant tests that make sense or are related to this change.

Thanks!

coolsoftwaretyler:
@BrianHung - I went ahead and added some more tests here. It was fun to play around with here, and I appreciate you taking a detailed look at how we can improve the overhead on this one.

I'm going to merge this, but when you get a chance, will you please [update this discussion post](https://github.com/mobxjs/mobx-state-tree/discussions/2096) with any details about how you found this problem? Did you notice memory overhead of maps in a particular scenario? How much improvement did you get? I want to take whatever was the cause of your PR and add it to some of our measurement work around performance.

Thanks!

PR #2094, opened 2023-09-22
Fix wording and discuss a bad link in philosophy.md

StefanWallin:
Not sure what to do with the duplicated link

- rephrase a sentence that was referencing a previous method name `toggleTodo()` which has been renamed to `toggle()`.
- highlight a link that should be updated to a proper one by maintainers.

## Steps to validate locally
[Read up on the philosophy of MobX-state-tree](https://mobx-state-tree.js.org/intro/philosophy) and follow the links. More specifically, the two paragraphs directly after Runtime and Designtime type error images.

### DIFF
--- a/docs/intro/philosophy.md
+++ b/docs/intro/philosophy.md
@@ -66,11 +66,11 @@ _Runtime type error_
 
 _Designtime type error_
 
-Because state trees are living, mutable models, actions are straight-forward to write; just modify local instance properties where appropriate. See `toggleTodo()` above or the examples below. It is not necessary to produce a new state tree yourself, MST's snapshot functionality will derive one for you automatically.
+Because state trees are living, mutable models, actions are straight-forward to write; just modify local instance properties where appropriate. See the `toggle()`-action in the Todo-store above or the examples below. It is not necessary to produce a new state tree yourself, MST's snapshot functionality will derive one for you automatically.
 
 Although mutable sounds scary to some, fear not, actions have many interesting properties.
 By default trees can only be modified by using an action that belongs to the same subtree.
-Furthermore, actions are replayable and can be used to distribute changes ([example](https://github.com/coolsoftwaretyler/mst-example-boxes/blob/main/src/stores/socket.js)).
+Furthermore, actions are replayable and can be used to distribute changes ([unclear example, where is the code to look at?](https://github.com/coolsoftwaretyler/mst-example-boxes/blob/main/src/stores/socket.js)).
 
 Moreover, because changes can be detected on a fine grained level, JSON patches are supported out of the box.
 Simply subscribing to the patch stream of a tree is another way to sync diffs with, for example, back-end servers or other clients ([example](https://github.com/coolsoftwaretyler/mst-example-boxes/blob/main/src/stores/socket.js)).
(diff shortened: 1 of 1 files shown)


docs/intro/philosophy.md
```diff
@@ -66,11 +66,11 @@ _Runtime type error_
...
+Because state trees are living, mutable models, actions are straight-forward to write; just modify local instance properties where appropriate. See the `toggle()`-action in the Todo-store above or the examples below. It is not necessary to produce a new state tree yourself, MST's snapshot functionality will derive one for you automatically.
 
 Although mutable sounds scary to some, fear not, actions have many interesting properties.
 By default trees can only be modified by using an action that belongs to the same subtree.
-Furthermore, actions are replayable and can be used to distribute changes ([example](https://github.com/coolsoftwaretyler/mst-example-boxes/blob/main/src/stores/socket.js)).
+Furthermore, actions are replayable and can be used to distribute changes ([unclear example, where is the code to look at?](https://github.com/coolsoftwaretyler/mst-example-boxes/blob/main/src/stores/socket.js)).
```
StefanWallin, line 73: This link should be changed, and the anchor text change be reverted.

coolsoftwaretyler: Hey @StefanWallin - thank you for the feedback on this. I agree the example is unclear. This doc is pretty old. It was modified recently when we moved a bunch of those old example repos out of the MST repository itself, so I mostly did some find-and-replacing, without a lot of edit time on the example itself.

I would be willing to change this in such a way that:

1. Drop the example code, and instead write a few sentences clarifying what it means to be "replayable" and "distribute changes"
2. Point to some other example, either in a separate repo, or even just as inline code behind some markdown code-fences.
3. Some other option you prefer

One way or another, yes let's clarify this. I don't think we have any particular strong desire to keep the same examples we've had for years, other than to recognize the good work that went into the project before we came along. But I"m happy to edit things around.

Are you available to make one of these changes? If not, we can open an issue about it and find someone else to do it.

Let me know what you think, and thanks for your time!

StefanWallin: I'm quite new to MST, I can make the change to the PR, but am quite unsure what to put there right know to in a succinct way demonstrate replayability. Given that we found this issue, maybe inline code would be best? It would at least be clearer to the next sweeping edit than a anonymous link.

What do you think about a code snippet like this?
```javascript
import { onSnapshot, applySnapshot } from "mobx-state-tree"

const history = []

onSnapshot(store, snapshot => {
  history.push(snapshot)
})

replay(index) {
  applySnapshot(store, history[index])
}
```

StefanWallin: Now I found the link I think you wanted to use, the [time.js](https://github.com/coolsoftwaretyler/mst-example-boxes/blob/main/src/stores/time.js)-file from the example repo... It's sort of like mine but probably a lot clearer. I can change the link to that and make a minimal PR instead of introducing a code snippet.

coolsoftwaretyler: @StefanWallin - yes, please do! Thank you!

coolsoftwaretyler: Hey @StefanWallin - I went ahead and made that change so we can merge. Thanks for the help!

Issue #2092, opened 2023-09-21
Remove Lerna config and lift MST sub-package up


coolsoftwaretyler:
This is the last step to making MST no longer a monorepo.

With https://github.com/mobxjs/mobx-state-tree/pull/2091, we finally removed the last of the sub-packages out of this repository. But we are currently still set up in Lerna, and with sub-packages. We won't change that until we close most (or hopefully all) of the open PRs in the repository to avoid gnarly merge conflicts.

This issue should get closed once we have a PR to actually remove all the monorepo structure. We put out a discussion post over here: https://github.com/mobxjs/mobx-state-tree/discussions/2089 to let folks know to anticipate it.

PR #2098, opened 2023-10-07
Chore/undo monorepo setup

coolsoftwaretyler:

This PR resolves https://github.com/mobxjs/mobx-state-tree/issues/2092, which completes https://github.com/mobxjs/mobx-state-tree/milestone/4. It removes all of the [Lerna](https://github.com/lerna/lerna) configuration we needed when MobX-State-Tree was a monorepo. It also moves some configuration files around, and changes some formatting options since this is going to be a huge diff anyways.

Here is a list of all the things contained in this commit/PR:

1. Uninstalled Lerna
1. Moved the contents of `packages/mobx-state-tree` to the root. So there's now `__tests__/`, `scripts/`, and `src/` at the root. We combined some more config files as well (discussed below).
1. Combined the Rollup config. Before, there was a base config at the root, and a separate config in the sub-package. Now we just have one config file at the root.
1. Combined `tsconfig.json` and updated file paths. Did not move this into `package.json` because `__tests__/tsconfig.json` wants to inherit from it.
1. Deleted the separate `test-results/` sub-folders, since we only have one package to test now.
1. Deleted the `tslint.json` in the `__tests__/` folder - it was only extending the base `tslint.json` file.
1. There are some existing TSLint errors in `__tests__/core/type-system.test.js`, but those exist in `packages/mobx-state-tree/__tests__/core/type-system.test.ts` in `master` as is, so this is not a regression.
1. Deleted `yarn test:cyclic` script. It didn't seem to be testing much, so removed it and references to it.
1. Updated path in size-limit configuration, and moved that config to `package.json`
1. Updated TypeScript to 3.8.3 to resolve peer dependency warning in `yarn install`
1. Updated CircleCI config for new commands and file paths.
1. Moved Prettier config to `package.json`, changed `tabWidth` to 2, added Prettier scripts to check, list, and write to files.
1. Moved size-limit configuration into `package.json`.
1. Deleted empty watchman configuration
1. Removed the `release` script - since we don't have a monorepo any more, I think doing this through `yarn tag-new-version` and a separate step to publish to npm is sufficient. Would be happy to take a future PR to automate this further.
2. Deleted config for an external tool called Wallaby - doesn't seem to have been installed as a devDependency or anything. If anyone wants to use it, we can open a separate PR to bring it back.

## Steps to validate locally

1. Pull this branch
1. For a clean slate, `rm -rf node_modules/ && yarn`

Run through the `package.json` scripts. They should mostly work in any order, but here's an order that makes the most sense to me. Some scripts run each other, but should be pretty robust and not rely too heavily on order-of-operations.

1. `yarn prepare`
1. `yarn prettier:list` (consider making a prettier error in a file to see it fail)
1. `yarn prettier:check` (consider making a prettier error in a file to see it fail)
1. `yarn prettier:write` (consider making a prettier error in a file to see it write the fix)
1. `yarn clean`
1. `yarn build` (runs `yarn clean` as well)
1. `yarn build` (run it again to check that cleaning works as expected)
1. `yarn jest` (you may get an open handle warning, this happens on `master` as well)
1. `yarn test` (you may get an open handle warning, this happens on `master` as well)
1. `yarn jest:perf`
1. `yarn test:perf`
1. `yarn test:dev` (again, open handle warnings are not regressions)
1. `yarn test:prod` (open handle warnings are not regressions)
1. `yarn coverage` (check the `coverage` folder, not version controlled)
1. `yarn test:all` (open handle warnings are not regressions - also generates coverage report)
1. `yarn size`
1. `yarn tag-new-version` (please use an `alpha` version, this will push to GitHub)
1. `yarn deduplicate` (this probably won't actually do anything)
1. `yarn fix-typedoc`
1. `yarn build-docs` (also runs `fix-typedoc`)
1. Please **skip** publishing docs for now - we can run and troubleshoot this when we actually release `v5.3.0`
1. `yarn start` (will open a local tab of the docs site)

We have a command for `yarn lint`, which fails, but it fails the same on `master`, so this is not a regression.

After you've checked things working in the repo, you can check that the build output works. I released a `5.3.0-alpha.1` version to npm, which includes some of the recent fixes and additions to the library, and was built off this branch to make sure the new setup works as expected.


coolsoftwaretyler:
I'm going to close this and see if I can trigger CI again. Doesn't seem to be running on subsequent commits.

PR #2103, opened 2023-10-12
Make Bun as Default Package Manager

chakrihacker:


coolsoftwaretyler:
Hey @chakrihacker - I've been thinking the TypeScript upgrade will be really helpful to resolve some of our existing TypeScript issues. Are you going to be moving this PR forward, or do you mind if I take that on in the next couple of weeks?

package.json
```diff
@@ -86,7 +86,7 @@
     "tslint-config-prettier": "^1.18.0",
     "typedoc": "0.15.8",
     "typedoc-plugin-markdown": "2.2.11",
-    "typescript": "^3.8.3",
```
coolsoftwaretyler, line 89: I think [TypeScript 5.4 will give us a lot of useful tools](https://devblogs.microsoft.com/typescript/announcing-typescript-5-4-beta/#the-noinfer-utility-type) like `NoInfer`.

Also changing this to be a literal version, because [TypeScript doesn't follow semantic versioning](https://www.learningtypescript.com/articles/why-typescript-doesnt-follow-strict-semantic-versioning), so we can't rely on things like major version locking to keep TS behavior stable.

src/core/node/object-node.ts
```diff
@@ -215,7 +215,9 @@ export class ObjectNode<C, S, T> extends BaseNode<C, S, T> {
     const type = this.type
 
     try {
-      this.storedValue = type.createNewInstance(this._childNodes)
```
coolsoftwaretyler, line 218: I still want to fix this, but I've got this error on `master`, so it may not be because of the TS upgrade

Screenshot 2024-02-12 at 8 12 22 PM

coolsoftwaretyler:
We got a separate PR to update TypeScript over in https://github.com/mobxjs/mobx-state-tree/pull/2103. I'm going to pull that in, then rework this PR so it just focuses on the Bun changes.

coolsoftwaretyler:
I'm gonna close this in favor of https://github.com/mobxjs/mobx-state-tree/pull/2148 because of the conflicts introduced from https://github.com/mobxjs/mobx-state-tree/pull/2103

PR #2116, opened 2023-11-09
feat: pass the node to postProcess snapshot transformers

airhorns:

When transforming the snapshot for a node, it would be quite handy to get access to the node the snapshot was generated for, so you can traverse the tree to get at data you want to put in the snapshot. Stored properties of that node are already given to you in the hook in the snapshot, but computeds, data from parents, tree environment data, or volatiles aren't accessible for use in snapshot transformers unless the node is passed in.

This brings back half of #2065 which was reverted as the `preProcess` changes weren't a satisfactory API. I think that decision is wise, but this PR targets `postProcess` only. I think it should always be possible to pass the node in to `postProcess`, as we're never in the middle of constructing a node.

My concrete use case for this is caching the output of some expensive computed views in the snapshot itself. I don't think that needs to be a core MST feature, but for the brave among us who want to venture into that territory, I think it'd be great if MST made it possible. 

## Steps to validate locally

Added some tests, kept the existing test suite passing! Annoyingly, if we do want to pass the node in, the node has to exist whenever we go to snapshot, and that means `createObservableInstanceIfNeeded` has to be called for nodes with a postProcessor a bit earlier than it used to be. I think that is ok as not that many nodes have postProcessors.

cc @BrianHung @coolsoftwaretyler

### DIFF
--- a/docs/concepts/snapshots.md
+++ b/docs/concepts/snapshots.md
@@ -56,3 +56,5 @@ Useful methods:
 -   `getSnapshot(model, applyPostProcess)`: returns a snapshot representing the current state of the model
 -   `onSnapshot(model, callback)`: creates a listener that fires whenever a new snapshot is available (but only one per MobX transaction).
 -   `applySnapshot(model, snapshot)`: updates the state of the model and all its descendants to the state represented by the snapshot
+
+`mobx-state-tree` also supports customizing snapshots when they are generated or when they are applied with [`types.snapshotProcessor`](/overview/hooks).
--- a/docs/overview/hooks.md
+++ b/docs/overview/hooks.md
@@ -14,50 +14,28 @@ title: Lifecycle hooks overview
     <a style="font-style:italic;padding:5px;margin:5px;"  href="https://egghead.io/lessons/react-loading-data-from-the-server">Hosted on egghead.io</a>
 </details>
 
-All of the below hooks can be created by returning an action with the given name, like:
+`mobx-state-tree` supports passing a variety of hooks that are called throughout a node's lifecycle. Hooks are passes as actions with the name of the hook, like:
 
 ```javascript
-const Todo = types.model("Todo", { done: true }).actions(self => ({
-    afterCreate() {
-        console.log("Created a new todo!")
-    }
+const Todo = types.model("Todo", { done: true }).actions((self) => ({
+  afterCreate() {
+    console.log("Created a new todo!")
+  }
 }))
 ```
 
-_⚠ The section below is outdated, and should be updated to use [`types.snapshotProcessor`](/API/#snapshotprocessor) instead of the snapshot hooks⚠_
-
-The exception to this rule are the `preProcessSnapshot` and `postProcessSnapshot` hooks (see `types.snapshotProcessor` as an alternative):
-
-```javascript
-types
-    .model("Todo", { done: true })
-    .preProcessSnapshot(snapshot => ({
-        // auto convert strings to booleans as part of preprocessing
-        done: snapshot.done === "true" ? true : snapshot.done === "false" ? false : snapshot.done
-    }))
-    .actions(self => ({
-        afterCreate() {
-            console.log("Created a new todo!")
-        }
-    }))
-```
-
-Note: pre and post processing are just meant to convert your data into types that are more acceptable to MST. Typically it should be the case that `postProcess(preProcess(snapshot)) === snapshot. If that isn't the case, consider whether you shouldn't be using a dedicated a view instead to normalize your snapshot to some other format you need.
-
-| Hook                  | Meaning                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
-| --------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `afterCreate`         | Immediately after an instance is created and initial values are applied. Children will fire this event before parents. You can't make assumptions about the parent safely, use `afterAttach` if you need to.                                                                                                                                                                                                                                                                                                                                                                                                                                  |
-| `afterAttach`         | As soon as the _direct_ parent is assigned (this node is attached to another node). If an element is created as part of a parent, `afterAttach` is also fired. Unlike `afterCreate`, `afterAttach` will fire breadth first. So, in `afterAttach` one can safely make assumptions about the parent, but in `afterCreate` not                                                                                                                                                                                                                                                                                                                   |
-| `beforeDetach`        | As soon as the node is removed from the _direct_ parent, but only if the node is _not_ destroyed. In other words, when `detach(node)` is used                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |
-| `beforeDestroy`       | Called before the node is destroyed, as a result of calling `destroy`, or by removing or replacing the node from the tree. Child destructors will fire before parents                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
-| `preProcessSnapshot`  | Before creating an instance or applying a snapshot to an existing instance, this hook is called to give the option to transform the snapshot before it is applied. The hook should be a _pure_ function that returns a new snapshot. This can be useful to do some data conversion, enrichment, property renames, etc. This hook is not called for individual property updates. _\*\*Note 1: Unlike the other hooks, this one is \_not_ created as part of the `actions` initializer, but directly on the type!**\_ \_**Note 2: The `preProcessSnapshot` transformation must be pure; it should not modify its original input argument!\*\*\_ |
-| `postProcessSnapshot` | This hook is called every time a new snapshot is being generated. Typically it is the inverse function of `preProcessSnapshot`. This function should be a pure function that returns a new snapshot. _\*\*Note: Unlike the other hooks, this one is \_not_ created as part of the `actions` initializer, but directly on the type!\*\*\_                                                                                                                                                                                                                                                                                                      |
-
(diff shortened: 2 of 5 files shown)


coolsoftwaretyler:
Hey @airhorns - thank you for this. I think this is pretty interesting. I really appreciate the context you provided, tests, and documentation changes.

I've got a busy couple of days coming up, but I can look at this in depth over the weekend.

While browsing the diff, I saw this in the hooks documentation:

>  Note: pre and post processing are just meant to convert your data into types that are more acceptable to MST. Typically it should be the case that `postProcess(preProcess(snapshot)) === snapshot. If that isn't the case, consider whether you shouldn't be using a dedicated a view instead to normalize your snapshot to some other format you need.

Would this PR break that documented intention? Or have I misunderstood what this code would do/what that section is saying?

airhorns:
I think that this capability would indeed let you violate that intention. In general, I think that doc is correct, and the typical use of snapshot processors is to be pure functions that are the inverse of each other. For my use case (caching views in snapshots), the snapshot itself now has extra information in it for sure such that they aren't the inverse of one another, but, that information is recoverable just by running the view. So IMO my use case still complies with the spirit of that note, which I think is "don't use random state in snapshots that you might not have next time".

coolsoftwaretyler:
Thanks! That's pretty convincing in my opinion. I'll take a closer look this weekend!

docs/concepts/snapshots.md
```diff
@@ -56,3 +56,5 @@ Useful methods:
 -   `getSnapshot(model, applyPostProcess)`: returns a snapshot representing the current state of the model
 -   `onSnapshot(model, callback)`: creates a listener that fires whenever a new snapshot is available (but only one per MobX transaction).
 -   `applySnapshot(model, snapshot)`: updates the state of the model and all its descendants to the state represented by the snapshot
+
+`mobx-state-tree` also supports customizing snapshots when they are generated or when they are applied with [`types.snapshotProcessor`](/overview/hooks).
```
coolsoftwaretyler, line 60: **praise:** thanks for adding this, it definitely fleshes out the idea of the snapshot modification overall, or at least introduces it.

docs/overview/hooks.md
```diff
@@ -14,60 +14,84 @@ title: Lifecycle hooks overview
     <a style="font-style:italic;padding:5px;margin:5px;"  href="https://egghead.io/lessons/react-loading-data-from-the-server">Hosted on egghead.io</a>
 </details>
 
-All of the below hooks can be created by returning an action with the given name, like:
+`mobx-state-tree` supports passing a variety of hooks that are called throughout a node's lifecycle. Hooks are passes as actions with the name of the hook, like:
```
coolsoftwaretyler, line 17: **question:** @airhorns - why change this now? Was this just something you wanted to clean up, or do you feel like the changes in this PR specifically warrant the clarification?

I like these changes, but I'm just curious about the motivation for adding it to this PR.

docs/overview/hooks.md
```diff
@@ -14,60 +14,84 @@ title: Lifecycle hooks overview
...
-        console.log("Created a new todo!")
-    }
+const Todo = types.model("Todo", { done: true }).actions((self) => ({
+  afterCreate() {
+    console.log("Created a new todo!")
+  }
 }))
 ```
 
-_⚠ The section below is outdated, and should be updated to use [`types.snapshotProcessor`](/API/#snapshotprocessor) instead of the snapshot hooks⚠_
```
coolsoftwaretyler, line 27: **praise:** @airhorns - thank you!

docs/overview/hooks.md
```diff
@@ -14,60 +14,84 @@ title: Lifecycle hooks overview
...
+| `postProcessSnapshot` | Deprecated, prefer `types.snapshotProcessor`. This hook is called every time a new snapshot is being generated. Typically it is the inverse function of `preProcessSnapshot`. This function should be a pure function that returns a new snapshot. _\*\*Note: Unlike the other hooks, this one is \_not_ created as part of the `actions` initializer, but directly on the type!\*\*\_                                                                                                                                                                                                                                                                                                      |
```
coolsoftwaretyler, line 36: **praise:** @airhorns - thank you

docs/overview/hooks.md
```diff
@@ -14,60 +14,84 @@ title: Lifecycle hooks overview
...
+| --------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `preProcess(inputSnapshot)` | Transform a snapshot before it is applied to a node. The output snapshot must be valid for application to the wrapped type. The `preProcess` hook is passed the input snapshot, but not passed the node, as it is not done being constructed yet, and not attached to the tree. If you need to modify the node in the context of the tree, use the `afterCreate` hook. |
+| `postProcess(outputSnapshot, node)` | Transform a snapshot after it has been generated from a node. The transformed value will be returned by `getSnapshot`. The `postProcess` hook is passed the initial outputSnapshot, as well as the instance object the snapshot has been generated for. It is safe to access properties of the node or other nodes when post processing snapshots. |
```
coolsoftwaretyler, line 97: **praise:** @airhorns thank you. If anyone else is looking at this PR as opposed to #2065, I think this extra documentation makes the case for why this may be a balanced approach.

docs/overview/hooks.md
```diff
@@ -14,60 +14,84 @@ title: Lifecycle hooks overview
...
-        done: snapshot.done === "true" ? true : snapshot.done === "false" ? false : snapshot.done
-    }))
-    .actions(self => ({
-        afterCreate() {
-            console.log("Created a new todo!")
-        }
-    }))
-```
-
 Note: pre and post processing are just meant to convert your data into types that are more acceptable to MST. Typically it should be the case that `postProcess(preProcess(snapshot)) === snapshot. If that isn't the case, consider whether you shouldn't be using a dedicated a view instead to normalize your snapshot to some other format you need.
```
coolsoftwaretyler, line 27: **suggestion:** @airhorns - we should remove or modify this since it would no longer be strictly true with your changes.

coolsoftwaretyler (commented):
@airhorns - thanks again for all of this. I am pretty happy to approve it, much as I was with #2065 - although I moved a little too quickly on that one and we found some issues with it.

I really appreciate the work you did to find the balance here, and resolve the specific node construction issues brought up in the first attempt.

It's clear to me that the community wants this since it's the second PR to this effect we've received. I'm leaning towards merging it in, especially considering the updated tests and docs.

I have two remaining concerns:

1. This does violate some of the existing documentation and initial intentions of MST. I think the change is subtle, but I don't want to make a unilateral decision on it since we'll be changing a longstanding library, and I don't want to subvert expectations too quickly.
2. Truthfully, since I had to revert #2065 - I am maybe a little reluctant to go ahead and merge a related PR on my own.

@chakrihacker - would you mind taking a look at this PR, along with my existing review? If you're satisfied with it, I'd love to merge it in. If you have any objections, I'd like to see if we can find a path forward. I think @airhorns did a great job with this and I'd love to find a way to make it work.

Thanks for everyone's time! Really appreciate it.

docs/overview/hooks.md
airhorns: Was just trying to clarify it a bit while I was in here, can back this out if you prefer

docs/overview/hooks.md
airhorns: Updated this -- moved it down closer to the snapshot section as well

docs/overview/hooks.md
coolsoftwaretyler: No, I love it. I was just curious. Thank you for taking the time to do so! Resolved.

coolsoftwaretyler:
Thanks for the follow-up @airhorns! I left one of my comments open for @chakrihacker or anyone else who might come check out the PR. I'm feeling good about this. Hoping to get a second maintainer review on it soon.

Have a great weekend!

chakravarthi-bb:
Hey @airhorns PR is really good. I want to do some local testing. I will do it in the evening and update it.

docs/overview/hooks.md
```diff
@@ -14,60 +14,129 @@ title: Lifecycle hooks overview
...
+    return {
+        ...snapshot,
+        createdAt: node.createdAt.getTime()
+    }
+});
 
-Note, except for `preProcessSnapshot` and `postProcessSnapshot`, all hooks should be defined as actions.
+const todo = Todo.create({done: true, createdAt: new Date()});
+const snapshot = getSnapshot(todo); 
+// { done: true, generatedAt: 1699504649386 }
```
chakrihacker, line 93: `createdAt` please

src/types/utility-types/snapshotProcessor.ts
```diff
@@ -1,3 +1,4 @@
+import { untracked } from "mobx"
```
chakrihacker, line 1: Please remove unused import

src/types/utility-types/snapshotProcessor.ts
```diff
@@ -14,7 +15,11 @@ import {
   devMode,
   ComplexType,
   typeCheckFailure,
-  isUnionType
+  isUnionType,
+  Instance,
+  BaseNode,
```
chakrihacker, line 20: Please remove unused import

src/types/utility-types/snapshotProcessor.ts
```diff
@@ -14,7 +15,11 @@ import {
...
+  ObjectNode,
+  isModelType
```
chakrihacker, line 22: Please remove unused import

chakrihacker (changes requested):
Minor feedback is requested. Great work

coolsoftwaretyler:
Thanks for the review @chakrihacker. Looks like @airhorns has implemented your feedback, so I'm going to merge this and get started on a pre release.

coolsoftwaretyler (approved):

PR #2117, opened 2023-11-09
feat: add t export, update documentation

coolsoftwaretyler:

This PR adds a new exported namespace from `mobx-state-tree`, called `t`. The intention is to disambiguate the idea of "types" from "the thing MST exports that people use to create models and stuff".

It's very common for users and maintainers alike to be discussing both TypeScript and MobX-State-Tree in the same context, and overloading the word "types" can be a bit of a pain.

This change gives us the option to write `import { t } from "mobx-state-tree"`, and discuss that namespace as `t`, which should improve communication in both code and person-to-person communication. Jamon and I have been kicking this idea around for a while and I wanted to get it out there since it's a quick win, IMO.

## Steps to validate locally

This change adds some notes through the docs making note of it. I also ran `yarn build-docs`, which created a bigger diff than this really is.

To check my work, the relevant files are:

-   `README.md`
-   `__tests__/core/api.test.ts`
-   `docs/concepts/trees.md`
-   `docs/intro/getting-started.md`
-   `src/index.ts`

Any other changes are auto-generated.

I added a test that checks for that export. It will fail if you remove the change in `src/index.ts`. This has no impact on the existing `types` namespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,51 +28,51 @@ See the [Getting started](https://mobx-state-tree.js.org/intro/getting-started) 
 There's nothing quite like looking at some code to get a feel for a library. Check out this small example of an author and list of tweets by that author.
 
 ```js
-import { types } from "mobx-state-tree"
+import { types } from "mobx-state-tree" // alternatively: import { t } from "mobx-state-tree"
 
 // Define a couple models
 const Author = types.model({
-    id: types.identifier,
-    firstName: types.string,
-    lastName: types.string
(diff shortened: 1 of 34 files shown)


jamonholmgren (approved):
Excited for this one, and IMO in a future version we could change our documentation to use it.

coolsoftwaretyler:
Thanks Jamon! I agree. Gonna merge this in.

PR #2137, opened 2024-02-04
Fix missing fail import

k-ode:

Fixes a missing import that would crash `types.snapshotProcessor` on usage.

The reason it wasn't caught by typescript is that it assumed `fail` was a built-in jest function.


coolsoftwaretyler:
Thanks @k-ode! Makes sense. We have seen this in a few places before.

If you have a chance, would you mind writing a test case for this? We recently made that a PR requirement and I'd love to protect against regressions on this.

coolsoftwaretyler:
Thanks @k-ode! Merging now and I'll push a patch.

I've also added a new issue to fix this across the whole project. We've seen it before: https://github.com/mobxjs/mobx-state-tree/issues/2138

PR #2149, opened 2024-02-23
Convert to bun:test for our tests, remove codecov

coolsoftwaretyler:

This PR will convert our Jest test to [bun test](https://bun.sh/docs/cli/test). This offers us a few advantages:

1. Huge improvement in speed (locally, I see ~2-3x improvement)
2. Native support for TypeScript, so no need for `ts-jest`
3. Without the need for `ts-jest` and `jest` itself, we reduce our `devDependencies` as well
4. Since we [have moved over to Bun for dependency management and script running](https://github.com/mobxjs/mobx-state-tree/pull/2148), this makes our development tooling more consistent overall.
5. This also removes [codecov](https://www.npmjs.com/package/codecov), partially because we didn't take the time to implement it with Bun, but also because that package is deprecated, and I don't think our token has been working for some time. We discussed this in the maintainer chat and agreed.
6. Fixes a missing `MstError` from a rebase

## Steps to validate locally

1. Make sure you have [latest Bun installed](https://bun.sh/docs/installation)
7. Pull down this branch
8. Run `bun install`
9. Then run `bun test:all` - tests should pass.
10. If you time it with `time bun test:all`, you can measure its runtime. Here's what I got on my 2020 Macbook Pro with Intel i7/32 GB RAM:

```
bun test:all  14.55s user 6.84s system 100% cpu 21.302 total
```

Compare this to the `yarn test:all` run in `master`, which reports:

```
Ran all test suites.
✨  Done in 75.39s.
```

Ran it with `time yarn test:all` and got slightly better results but with a lot more CPU usage:

```
yarn test:all  178.77s user 24.75s system 432% cpu 47.026 total
```


coolsoftwaretyler:
I was able to get this working with a little bit of bash scripting. I don't love it, but since the Bun test runner [doesn't behave precisely like `jest -i`](https://github.com/oven-sh/bun/issues/1825#issuecomment-1961563691), someone recommended [this command instead](https://github.com/oven-sh/bun/issues/1825#issuecomment-2003453515), which we can chain together to get full tests running with minimal skips and changes.

The major downside right now is that coverage is no longer correctly generated. I will see if I can fix that up.

Bun also shipped support for `expect.assertions`, so I was able to keep that expectation in there

coolsoftwaretyler:
I added the specific `toThrow` error messages back to those expectations, and then realized we weren't appropriately failing with the one-liner script to run all tests, so I put together separate bash scripts that will fail the process if we got any failures (and output the full results if so).

I also removed `cross-env` for now since we were only using it in those test commands, and we've swapped to bash scripts which presume a specific type of environment variable setting.

Issue #1525, opened 2020-05-18
Model instance maybe fields becoming TypeScript optional fields when included in a types.union


fruitraccoon:
**_Bug report_**

* [x] I've checked documentation and searched for existing issues
* [x] I've made sure my project is based on the latest MST version
* [x] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

**Sandbox link or minimal reproduction code**
https://codesandbox.io/s/wispy-meadow-z49x9

I've pinned the MST and Mobx versions in the reproduction, so that the issue can be reproduced if new versions affect the behaviour in any way (good or bad!).

**Describe the expected behavior**
That the TypeScript interface derived from `Instance<typeof Model>` would always "fit" an instance of that model.

**Describe the observed behavior**
When a child model has an a "maybe" field, and is then used within a union type, attempting to pass an instance of this model from the union field to a React prop typed as `Instance<typeof Model>` fails with "`Property '...' is optional in type ... but required in type 'IModel'`"

It's probably easiest to view the linked reproduction code.

I _think_ what's happening is that when the model type is coming from within the union, "maybe" fields are becoming optional fields - but they are still mandatory fields on the derived interface.  I can see that typescript reports the model type differently when it's come from the union vs. when it has not.

I realise this is a pretty specific use case, but I thought it worth raising.  If there's any issue with the reproduction or more information is required, please let me know!

alex-shamshurin:
Hallo, I have the same issue. 

`Property is optional in type but required in type` ...  For example - property in model is `name: types.maybe(types.string)` type and ts object which I'd like to assign to this model is `name? string` which cause incompatibility error. I have tried to use `types.optional` but it doesn't help. 

At root level of model  I tried this type:

```
export type Complete<T> = {
  [P in keyof Required<T>]: Pick<T, P> extends Required<Pick<T, P>>
    ? T[P]
    : T[P] | undefined;
};

```
and being applied to root model it helps to transform typescript optional to `types.maybe`, but suddenly it does not work for inner models and problem still exists.

coolsoftwaretyler:
Hey @fruitraccoon - sorry it's been so long since anyone got back to you here. Not sure if this is still a relevant issue. I have two thoughts:

1. In the short term, you can provide your `itemWithIssueModel` variable with an explicit type annotation, like this:

```ts
[...]

Issue #1833, opened 2021-11-17
Accessing a field of a model that inherits from another leads to a missing type error


buck06191:
**_Bug report_**

* [x] I've checked documentation and searched for existing issues
* [x] I've made sure my project is based on the latest MST version
* [x] Fork [this](https://codesandbox.io/s/intelligent-wood-sfvg4) code sandbox or another minimal reproduction. 

**Sandbox link or minimal reproduction code**

The error can be seen in particular at the above code sandbox in the [`Person.ts`](https://codesandbox.io/s/intelligent-wood-sfvg4?file=/src/Person.ts) file.

**Describe the expected behavior**

When using `getRoot` to access a model that is composed over a variety of others e.g. a model of type `Food` where
```typescript
export const Food = types.union({ dispatcher }, Fruit, Orange, Apple);
```
and `Orange` and `Apple` are created via `Fruit.props({...})`, I'd expect that properties that only exist on one of the "child" models would be optional i.e. would return as `undefined` if the runtime model was one without that field. In the sandbox linked to for example I would expect to be able to call
```typescript
const {
        food: { seeded }
      } = getRoot<IRoot>(self);
```
and for `seeded` to be `undefined` for an `Apple` and a boolean for an `Orange`.

**Describe the observed behavior**

Only the properties from the "Parent" model (in this case the `Fruit` model) are recognised as valid properties by Typescript.
For example, in the code sandbox example when trying to access the `seeded` property there is a compilation error that `seeded` does not exist on the type.
![image](https://user-images.githubusercontent.com/9122944/142246265-4c88ae24-d36f-4ab4-aa99-4d6a46501d10.png)
Looking at the error it seems to be using an intersection of properties rather than a union.

AaronPorts:
Ran into this the other day. Types in dispatcher really intersect each other.

coolsoftwaretyler:
Thanks for the initial write up @buck06191, and I'm sorry that it took so long for us to get back to you! (@AaronPorts - I appreciate you jumping in here as well).

This seems to be a TypeScript issue, so I'm going to label it as such. We are definitely open for PRs to improve typings, so I'll add a label to that effect as well.

If either of y'all are interested in helping out, that would be great. But since it's been so long, no worries if not! We will find time or someone to resolve this eventually.

Issue #1664, opened 2021-03-09
Array and model types are not inferred correctly when broken down into their components


ada-waffles:
**_Bug report_**

* [x] I've checked documentation and searched for existing issues
* [x] I've made sure my project is based on the latest MST version
* [x] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

**Sandbox link or minimal reproduction code**

https://codesandbox.io/s/great-sea-tpn6b?file=/src/index.ts

**Describe the expected behavior**

An array type should have its creation, snapshot, and instance types correctly inferred when passed to a function that infers types as their components.

**Describe the observed behavior**

The instance type is inferred as `C | T` due to a strange behavior in TypeScript.

I have already opened an issue with TypeScript here: https://github.com/microsoft/TypeScript/issues/43168

In the case of MST, the definition of `IType#is` at https://github.com/mobxjs/mobx-state-tree/blob/v5.0.1/packages/mobx-state-tree/src/core/type/type.ts#L101 causes this issue due to the interaction with the conditional type in `STNValue`. The issue doesn't occur for named interfaces unless they are subtyped, which is why this isn't affecting all types, just those such as arrays which subtype `IType`.

I'm opening this issue to raise awareness in case anyone wants to try working around the issue in MST.

For now I'm working around the issue by manually passing `<Creation[] | undefined, Snapshot[], IMSTArray<typeof Type>>` to the functions like `union` that need the components. It's ugly, but it works.

duke9509:
This also affects models, since they also subtype `IType`.

I've found a less verbose workaround in the form of a helper function/interface:
```ts
export interface ITypeComponents<C, S, T> {
    readonly CreationType: C;
    readonly SnapshotType: S;
    readonly TypeWithoutSTN: T;
}

export const asIType = <C, S, T>(type: ITypeComponents<C, S, T>): IType<C, S, T> => type as IType<C, S, T>;
```

Passing the type inference through that interface avoids the issue: `t.union(asIType(MyModel), asIType(t.array(MyOtherModel)))`

coolsoftwaretyler:
Hey @ada-waffles - I'm sorry it took so long for anyone to get back to you here. Really appreciate the time you put into this.

Jamon and I are trying to give MST some of the attention it sorely deserves. One of the big items we're talking about is improved TS support, and I think this falls in that realm.

I don't have a path forward quite yet, but I'm going to label this issue and keep it open.

[...]

PR #2151, opened 2024-02-24
Improved typing for union types

thegedge:

After 9 or 10 types in a union, we devolve into an `IAnyType`. This isn't particularly useful, and potentially even dangerous (in a typechecking sense, given the behaviour of `any`).

This PR makes union types become proper TS sum types for any number of member types given to the union. It also results in types that are much simpler to manage, since we don't have to "unroll" for each possible number of types. We've also brought similar improvements to `types.enumeration`.

## Steps to validate locally

I added some new tests for union and enumeration types, and verified that the tests fail before this change and pass after.


__tests__/core/type-system.test.ts
```diff
@@ -33,14 +33,21 @@ type DifferingKeys<ActualT, ExpectedT> = {
...
->}`
+type NotExactErrorMessage<ActualT, ExpectedT> = ActualT extends Record<string, unknown>
+  ? ExpectedT extends Record<string, unknown>
+    ? `Mismatched property: ${DifferingKeys<ActualT, ExpectedT>}`
+    : "Expected a non-object type, but received an object"
+  : ExpectedT extends Record<string, unknown>
+  ? "Expected an object type, but received a non-object type"
+  : "Types are not exactly equal"
+
+type IsExact<T1, T2> = [T1] extends [T2] ? ([T2] extends [T1] ? Exact<T1, T2> : never) : never
```
thegedge, line 44: Realized this wasn't working super great for things that weren't object-y, like a string union (e.g., `"a" | "b"`)

src/types/utility-types/enumeration.ts
```diff
@@ -8,14 +8,11 @@ export type UnionStringArray<T extends readonly string[]> = T[number]
 // these overloads also allow both mutable and immutable arrays, making types.enumeration<Enum>(Object.values(Enum)) possible.
 // the only case where this doesn't work is when passing to the function an array variable with a mutable type constraint;
 // for these cases, it will just fallback and assume the type is a generic string.
-export function enumeration<T extends readonly string[]>(
-  options: T
-): ISimpleType<UnionStringArray<T>>
```
thegedge, line 13: I was kind of surprised that this made the snapshot/creation types of `types.enumeration(["a", "b", "c"])` `string` instead of `"a" | "b" | "c"`. This change aligns it with the other overloaded signature.

My main concern is that, although I think this is the correct thing to do, users may have been relying on the wrong types. If they did provide a wrong value at runtime it would have failed, but if they weren't doing that they may find themselves having to do a lot more `as "a" | "b" | "c"`.

src/types/utility-types/enumeration.ts
```diff
@@ -31,8 +28,11 @@ export function enumeration<T extends string>(
  * @param options possible values this enumeration can have
  * @returns
  */
-export function enumeration(name: string | string[], options?: any): ISimpleType<string> {
-  const realOptions: string[] = typeof name === "string" ? options! : name
+export function enumeration<T extends string[]>(
+  name: string | T,
+  options?: T
+): ISimpleType<T[number]> {
```
thegedge, line 34: Technically not necessary, since the overloads above define the signature, but I find it helps the implementer to have this be the proper signature.

thegedge:
Although, I can't quite explain it, but this appears to also resolve https://github.com/mobxjs/mobx-state-tree/issues/1525

On main:
CleanShot 2024-02-24 at 18 49 20

On this branch:
CleanShot 2024-02-24 at 18 50 15

__tests__/core/type-system.test.ts
coolsoftwaretyler: Ah yes, I see that. Nice, I like it!

This still doesn't give us the exact mismatch on string unions. I'm seeing:

```
Argument of type '"1" | "2" | "7" | "3" | "0" | "5" | "4" | "6" | "8" | "9"' is not assignable to parameter of type '"Types are not exactly equal"'.
  Type '"1"' is not assignable to type '"Types are not exactly equal"'
```

But I'm pretty happy with it as-is and we can iterate on it later.

src/types/utility-types/enumeration.ts
coolsoftwaretyler: Ah yeah. This is a tough call. This will personally break some of my own application code, but I think it's much more correct.

We spoke about this on Discord, but I'm willing to at least merge this in for now and ship it as a pre-release version. Maybe we move more slowly with this one than we have with prior pre-releases and get feedback.

Alternatively, we could ship the minor version, and revert this in a patch very quickly.

It's tough to balance stability with improving TypeScript, since most TypeScript changes can look like breaking changes. 

We won't rush on a release of this, but I like the change a lot, so I will make sure it's consumable via npm for you and anyone else who wants it. Thank you very much!

coolsoftwaretyler (commented):
Hey @thegedge - thank you again, so much, for this phenomenal work. I'm stoked to see how many issues it may resolve.

I left a few comments here. I have some minor suggestions and questions. Nothing blocking, so please feel free to either make those changes, or ask me to do it myself. I appreciate all the time you've already put into this.

As you mentioned, there are a few changes here that some folks may consider breaking. If anyone else sees this PR  and has concerns, I'd love to hear that. We will bring it up at the Maintainer's meeting on March 7th. I expect we will release a pre-release version for folks to try out, but probably take our time with releasing it, to make sure we don't cause a lot of downstream churn for folks'

src/types/utility-types/union.ts
```diff
@@ -152,104 +151,28 @@ export type _CustomCSProcessor<T> = Exclude<T, _NotCustomized> extends never
 export interface ITypeUnion<C, S, T>
   extends IType<_CustomCSProcessor<C>, _CustomCSProcessor<S>, T> {}
 
-// generated with packages/mobx-state-tree/scripts/generate-union-types.js
```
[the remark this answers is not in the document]
thegedge: Oh yeah, totally. Good eye!

Will 🔥

src/types/utility-types/enumeration.ts
thegedge: Yeah, it's a tough call to make!

The other thing I forgot to mention is that this makes it consistent with the named overload too. It'd likely be confusing to people to have these be different:

```
types.enumeration("Letter", ["a", "b"])  // "a" | "b"
types.enumeration(["a", "b"])            // string
```

but I'm guessing it's rare that people move from the latter to the former, in which case they'd encounter a similar issue outlined in my first comment.

src/types/utility-types/union.ts
```diff
@@ -35,7 +30,11 @@ export interface UnionOptions {
  * @internal
  * @hidden
  */
-export class Union extends BaseType<any, any, any> {
+export class Union<Types extends IAnyType[]> extends BaseType<
```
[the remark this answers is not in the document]
thegedge: I'll try my best. Hope this helps explain!

First, we need to know that there are tuple types (e.g., `["a", "b", "c"]`) and array types (e.g., `string[]`) in TypeScript. When we have this:

```ts
export function union<Types extends IAnyType[]>(...types: Types): IUnionType<Types>
```

That `Types extends IAnyType[]` could result in `Types` being either a tuple or array type, depending on how we call this function. When we do the following

```ts
types.union(types.literal("a"), types.literal("b"), types.literal("c"))
```

we'll find that `Types` is `[ISimpleType<"a">, ISimpleType<"b">, ISimpleType<"c">]`, and that's because TS is nice enough to narrow that `types` array to a tuple type (generally) because it knows statically that it is _exactly_ those three things. If we were to do the following

```ts
types.union(..."abc".split("").map(c => types.literal(c))
```

We'll find that `Types` is `ISimpleType<string>[]`. That's because `split` produced `string[]`, not `["a", "b", "c"]`. In theory, we could make `split` produce the right tuple type, but even then we wouldn't get what we want because:

```ts
const chars = ["a", "b", "c"] as const;
types.union(...chars.map(c => types.literal(c))
```

In that case, we get `ISimpleType<"a" | "b" | "c">[]` because `map` isn't typed to deal with (and I'm not sure we could make that work without higher-kinded types).

So that's how we got our tuple type, which is key to making the union type work the way we want it. The final bit is how do we pull out creation types and whatnot from that? Well, fortunately you can ask TypeScript to give you the type based on a certain index type for an array/tuple:

```ts
const a = ["a", "b", "c"] as const;
const b = a as string[];

// typeof a[number] == "a" | "b" | "c";
// typeof b[number] == string;
```

Makes sense, because if you're saying I have an array with exactly three items, the first being `"a"`, the second `"b"`, and the last `"c"`, if I index with some arbitrary number it has to be one of them, i.e., `"a" | "b" | "c"`. We could also do `typeof a[1]` to get `"b"`.

TS goes even further that if all parts of the union type share something in common, you can index that again. For example:

```ts
type Things = [{ prop: 1 }, { prop: boolean }];
type ThingsProp = Things["prop"];  // 1 | boolean
```

Using thius last bit of knowledge we can then construct our creation/snapshot/instance types:

```ts
// From our example above, `Types == [ISimpleType<"a">, ISimpleType<"b">, ISimpleType<"c">]`
// Given that, `Types[number]` will be `ISimpleType<"a"> | ISimpleType<"b"> |ISimpleType<"c">

type IUnionType<Types extends IAnyType[]> = ITypeUnion<
  Types[number]["CreationType"],   // "a" | "b" | "c"
  Types[number]["SnapshotType"],   // "a" | "b" | "c"
  Types[number]["TypeWithoutSTN"]  // "a" | "b" | "c"
>
```

From here, you may be able to see how this translates to the complex types like models and maps.

__tests__/core/type-system.test.ts
thegedge: Yeah, it's unfortunate. Unions are tough to handle here. I was hoping to maybe fall back to TS telling us what's missing by having it return `T1` for the argument in position `T2`, but this breaks things when `T2 extends T1`.

I'll keep iterating though! I'm hoping with some TLC we can make this really great, or maybe we'll find a library that does all of this for us.

src/types/utility-types/union.ts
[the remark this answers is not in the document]
coolsoftwaretyler: Thanks man! That's really helpful. I appreciate it a lot.

coolsoftwaretyler:
> Although I can't quite explain it, this appears to also resolve #1525
> 
> On main: CleanShot 2024-02-24 at 18 49 20
> 
> On this branch: CleanShot 2024-02-24 at 18 50 15
> 
> [EDIT] Same for #1664
> 
> Haven't dug deeper, but I'm wondering if it'll also knock off #1833

Hey @thegedge - I can't verify this fix. I'm seeing the issue still in a test in [bcac3c0](https://github.com/mobxjs/mobx-state-tree/pull/2151/commits/bcac3c0b6d7d446d9802cd4f86c5d644d4f7242f). How did you get the passing screenshot?

coolsoftwaretyler:
[c2b47ee](https://github.com/mobxjs/mobx-state-tree/pull/2151/commits/c2b47ee3714911f2af6a6060672dee914e95ecf4) confirms the fix for #1664.

coolsoftwaretyler:
[47c0702](https://github.com/mobxjs/mobx-state-tree/pull/2151/commits/47c070225735571689893dacde51e385bb0ab800) replicates https://github.com/mobxjs/mobx-state-tree/issues/1833 - so I don't think this fixes that quite yet either.

coolsoftwaretyler:
Merge conflict came from https://github.com/mobxjs/mobx-state-tree/pull/2152 since we're editing `__tests__/core/type-system.test.ts` in both. Let's resolve some of the other items before we resolve the conflicts. Should be able to pull from both current and incoming on that one, if I'm not mistaken.

src/types/utility-types/enumeration.ts
thegedge: Hmm, here's what I'm seeing on the main branch:

CleanShot 2024-02-26 at 09 14 00

And in my branch:

CleanShot 2024-02-26 at 09 16 39

So it's definitely possible you have a configuration difference, or maybe a slightly older version of TS?

thegedge:
Oh yeah, great call on regression tests for the issues we fixed!

I'll double check on #1525 this evening

src/types/utility-types/enumeration.ts
coolsoftwaretyler: Ah yeah, I see this in the MST repo as well. My application codebase is on TS 5.3.3, but there must be some kind of config problem on my end. I'll investigate and probably [update the TS tips page, which is very outdated](https://mobx-state-tree.js.org/tips/typescript).

I'm going to resolve this discussion and point to it in the preview release notes where I'll solicit feedback on breaking/not breaking version numbers. Thanks!

coolsoftwaretyler:
@thegedge - I [added a test for 1525](https://github.com/mobxjs/mobx-state-tree/pull/2151/commits/658154454dcc3b93d6e31c76171287550a2d6f25), and [it's failing in CI](https://app.circleci.com/pipelines/github/mobxjs/mobx-state-tree/1076/workflows/05bf26da-b181-4083-9ce9-07304a4b59b6/jobs/6168?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary)

We definitely don't need to block on this, but since it looks like you've seen the fix for 1525 at some point, would be cool to try and verify it. 

Let me know whatcha think!

coolsoftwaretyler:
I will remove that failing test and plan to merge this in sometime over the weekend unless I hear otherwise.

thegedge:
Hey, @coolsoftwaretyler. Thanks for 👀. We have a big launch week coming up so I've been super busy and neglected to check in. If I can't get to it this evening, let's ignore and follow up with #1525 later.

coolsoftwaretyler:
Perfect, thank you!

thegedge:
Amended your test commit, was just missing an important check from the issue to ensure narrowing happens. In this PR, it passes!

🙇 for adding all these tests. I'll keep that in mind for future work. A strong testing culture will be super helpful moving forward :)

coolsoftwaretyler:
Phenomenal!

I'll merge and then put out a pre-release build to npm, along with an RFC about how strict we want to be around semantic versioning here.

Issue #1524, opened 2020-05-17
TypeScript: how to get type of model properties without internal fields?


DnEgorWeb:
When I create a Model Snapshot (`typeof Model.SnapshotType`) additionally to the fields of my model I get NonEmptyObject. My question is how to get just my fields?  I.g. I have a Model:
```
MyModel = types.model('MyModel', {
  field_one: types.maybeNull(types.string),
  field_two: types.maybeNull(types.string),
})

export type MyModelType = typeof MyModel.Type
export type MyModelSnapshot = typeof MyModel.SnapshotType
```
Then if I want to type an object which contains properties from the model I do the following:
```
const myObject: Record<
  keyof MyModelSnapshot,
  { firstValue: string; secondValue: string; }
>  = {
  field_one: { firstValue: 'mock'; secondValue: 'mock; },
  field_two: { firstValue: 'mock'; secondValue: 'mock; },
}
```
And I get TS error on this object `Property '[$nonEmptyObject]' is missing in type { field_one: { firstValue: "mock"; secondValue: "mock"; }; field_two: { firstValue: "mock"; secondValue: "mock"; }; }`

DnEgorWeb:
As a workaround I applied Omit< MyModelSnapshot , symbol>

cmdcolin:
thanks so much for posting this fix!

just a note: I had a similar question here https://github.com/mobxjs/mobx-state-tree/discussions/1699

coolsoftwaretyler:
Hey @DnEgorWeb - thanks for putting this issue together and posting your workaround.

I think this is part of an underlying issue with the MobX-State-Tree TypeScript typings, so I'm going to label it as such and keep this open while the maintainers put together some priorities.

If you're interested in pitching in on the effort, I'd be happy to chat. But it's been a few years here so no worries if it's not on your radar anymore. We appreciate what you've already done here for us!

PR #2152, opened 2024-02-25
Eliminate `NonEmptyObject`

thegedge:

Resolves https://github.com/mobxjs/mobx-state-tree/issues/1524

`NonEmptyObject` was introduced in https://github.com/mobxjs/mobx-state-tree/pull/1269 to prevent passing in any object to a model type that had no props.

Although this works great, some people have found it harder to work with because now `keyof typeof MyModelType` now has this extra key they have to deal with. One could intersect that type with `string`, but it's unfortunate that users have to do this.

Instead, we use `Record<string, never>` which can only be represented by an empty object literal. Snapshot preprocessors also had to be tweaked because they can emit properties that aren't known to them in the case of composition with other types.

## Steps to validate locally

New test added to prove this works, and ran the existing test suite to ensure no breakages.

### DIFF
--- a/__tests__/core/type-system.test.ts
+++ b/__tests__/core/type-system.test.ts
@@ -1179,11 +1179,49 @@ test("object creation with no props", () => {
 
   MyType.create()
   MyType.create({})
-  MyType.create({ [Symbol("test")]: 5 })
 
-  // @ts-expect-error
-  MyType.create({ test: 5 })
+  // Instances can be created with their own instance type
+  MyType.create(MyType.create())
 
-  // @ts-expect-error
-  MyType.create({ another: 5 })
+  // TODO @ts-expect-error -- symbols aren't props (but may be one day)
+  // This currently is allowed, because excess property checking doesn't happen against symbols.
+  // See https://github.com/microsoft/TypeScript/issues/44794
+  true || MyType.create({ [Symbol("test")]: 5 })
+
+  // @ts-expect-error -- this is a view, not a prop
+  true || MyType.create({ test: 5 })
+
+  // @ts-expect-error -- unknown prop
+  true || MyType.create({ another: 5 })
+})
+
+test("object creation when composing with a model with no props", () => {
+  const EmptyType = types.model().views((_self) => ({
+    get test() {
+      return 5
+    }
+  }))
+
+  const NonEmptyType = types.model({
+    value: types.optional(types.number, 0),
+    negate: types.boolean
+  })
+
+  const Composed = types.compose(EmptyType, NonEmptyType)
+
+  Composed.create({ negate: true })
+  Composed.create({ negate: false })
+  Composed.create({ value: 5, negate: true })
+
+  // Instances can be created with their own instance type
+  Composed.create(Composed.create({ negate: true }))
+
+  // @ts-expect-error -- symbols aren't props (but may be one day)
+  true || Composed.create({ [Symbol("test")]: 5 })
+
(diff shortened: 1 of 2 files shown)


__tests__/core/type-system.test.ts
```diff
@@ -1151,7 +1150,6 @@ test("maybe / optional type inference verification", () => {
   assertTypesEqual(
     _ as ITC,
     _ as {
-      [$nonEmptyObject]?: any
```
coolsoftwaretyler, line 1154: **praise:** @thegedge - it's very exciting to me to see this gone, this makes the test a lot easier to understand, from my perspective.

__tests__/core/type-system.test.ts
```diff
@@ -1172,3 +1169,21 @@ test("maybe / optional type inference verification", () => {
...
+test("object creation with no props", () => {
+  const MyType = types.model().views((_self) => ({
+    get test() {
+      return 5
+    }
+  }))
+
+  MyType.create()
+  MyType.create({})
+  MyType.create({ [Symbol("test")]: 5 })
```
coolsoftwaretyler, line 1182: **suggestion:** @thegedge - this is a little unintuitive to me. What do you think about dropping a comment here about why this works (it would also help me understand, myself lol)

src/types/complex-types/model.ts
```diff
@@ -56,7 +56,6 @@ import {
   assertArg,
   FunctionWithFlag
 } from "../../internal"
-import { ComputedValue } from "mobx/dist/internal"
```
coolsoftwaretyler, line 59: **praise:** @thegedge - it's very exciting to me that we can remove this import, which feels like a code smell, importing from MobX internals directly.

coolsoftwaretyler (commented):
Hey @thegedge - this is awesome!

CI seems to be failing, but it passes locally for me (as you said it did for you, too). I tried re-running, but no luck.

I'm about to turn in for the night, but I can take a look tomorrow and see if I can figure out what's blocking CI.

Once that passes, I'll probably merge into `master`, and ship it as part of the pre-release build alongside https://github.com/mobxjs/mobx-state-tree/pull/2151.

Thanks again, this is awesome!

__tests__/core/type-system.test.ts
thegedge: Actually, I should have prompted a question here. My bad!

Would we like this to also fail? I can definitely put in the work to make it so, but curious what you think? I do recall seeing an issue about supporting symbol keys for props, so I could make this fail in anticipation of that.

thegedge:
> CI seems to be failing, but it passes locally for me (as you said it did for you, too). I tried re-running, but no luck.

Yeah, looks like it's typechecking locally, but not when I open the specific file. I think this is some misconfiguration of `tsconfig.json` (I'll follow up with a PR to deal with that).

thegedge:
Looks like that last attempt failed too. I'll keep poking. It's possible that we ended up at this `$nonEmptyObject` solution because it was the only one to satisfy all of the requirements (baked into the existing tests).

__tests__/core/type-system.test.ts
thegedge: I think we should make this fail. I just realized this fails for the case where an object does have props, so we should stay consistent!

thegedge:
Fixed. Added in some more checks and comments to hopefully help explain things.

coolsoftwaretyler:
LGTM! Thanks again for all the hard work.

PR #2165, opened 2024-03-13
docs: fix broken link

kylemeenehan:

Fixes a broken link in the documentation

## Steps to validate locally

Notice that the old link is no longer valid:
https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/__tests__/core/reference-custom.test.ts#L148

Notice that the new link is valid:
https://github.com/mobxjs/mobx-state-tree/blob/master/__tests__/core/reference-custom.test.ts#L127

Please note that I've updated the line number. I didn't check the history of that file, but I do believe this line number is the best entry point to explore the file in the context of this documentation.


coolsoftwaretyler:
Love it, thank you @kylemeenehan! I'm going to merge and will deploy shortly.

Issue #2138, opened 2024-02-06
Linters do not recognize `fail` as a required import in library code


coolsoftwaretyler:
We've run into this problem a few times, where TypeScript thinks `fail` is a Jest function, so it doesn't send an error when we're missing an import.

Here's a recent PR that fixes one instance of the problem:

https://github.com/mobxjs/mobx-state-tree/pull/2137

And we've seen it before: https://github.com/mobxjs/mobx-state-tree/issues/1989

I think this is a great issue for someone who is new to the project. We need to make sure we don't have missing imports, so we should either:

1. Tell TypeScript that `fail` needs to be defined (perhaps using folder paths to differentiate, since tests are separate)
2. Rename `fail` to something that doesn't collide with Jest, like `mstFail` or something

Either way, we should be able to verify we aren't missing any more `fail` imports with linting.

JuanJo4:
hey @coolsoftwaretyler, I'll work on this one ✋

coolsoftwaretyler:
Hi @JuanJo4, that's great! Let me know if you need a hand.

I think I'd slightly prefer the approach where we rename the function to be like, `mstFail`, so we don't collide with Jest.

JuanJo4:
> I think I'd slightly prefer the approach where we rename the function to be like, `mstFail`, so we don't collide with Jest.
agreed, and I decided to use a class instead of a function `class MstError extends Error`, it seems more natural to me to throw an Error and not a function.

I'll push a PR soon, not sure how to write a test for this though, any suggestions?

coolsoftwaretyler:
I also have a slight preference for a function here, I assume renaming the function and renaming its call sites is easier that way, but I'm happy to consider the class if it offers some other advantage.

As for a test, I would exercise the behavior - what parameters do we ask for, and what do we expect to happen? Maybe spy on whatever console method we use in the failures.

Issue #1989, opened 2022-12-09
fail is not defined exception instead of assertion


Sinled:
**_Bug report_**

* [x] I've checked documentation and searched for existing issues
* [x] I've made sure my project is based on the latest MST version
* [ ] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

**Describe the expected behaviour**
https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/core/node/BaseNode.ts#L166

it seems there is missing `fail` import, 
instead `"assertion failed: cannot finalise the creation of a node that is already dead" ` i receive `fail is not defined` error in my application

**Describe the observed behaviour**
Correct assertion should be thrown

coolsoftwaretyler:
Hey @Sinled - do you have a minimum reproduction of the error that you're seeing here?

Sinled:
@coolsoftwaretyler  It was 4 months ago, i can't remember exact reproduction.

But if you open this file
https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/core/node/BaseNode.ts#L166

`fail` function is called, but it is not imported anywhere in the file

but for example here 
https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/core/node/scalar-node.ts

`fail` function explicitly imported

as i understand it wont work unless `fail` is in global scope?

coolsoftwaretyler:
Thanks, @Sinled! That's still quite helpful.

I am going through old issues and triaging, so I appreciate you getting back to me. If you run into it again, a reproduction would be helpful, but it should be easy for us to find this and patch it up with the provided info.

Not sure on the timeline, just getting started with some maintenance here, but I will update this issue as things get prioritized and moved along.

coolsoftwaretyler:
Hey @nithinssabu - I'm going to assign this to you [since you asked to get tasked with some smaller bug fixes](https://github.com/mobxjs/mobx-state-tree/issues/2045#issuecomment-1648263158). I'd love to see a PR here that adds a test making sure we throw the correct error, and of course fix the missing import.

Let me know if you've got additional questions. Thanks!

coolsoftwaretyler:
Hmm, for whatever reason, GitHub is not letting me assign this to you. See if you can self-assign, @nithinssabu. Otherwise, this comment should serve as info enough to keep folks from doubling work.

[...]

PR #2171, opened 2024-03-31
fix: #2138 linter issues with fail function

JuanJo4:

It solves #2138, this has happened before (#2137 & #1989) and was solved only for those specific cases, this PR aims to fix once and for all.

This is what I did:
- removed `fail` function and created a new `MstError` class that extends from `Error`.
- adjusted the calls of the `fail` function accordingly.
- removed unused imports

## Steps to validate locally

I've added some basic unit tests, not sure if that's enough.


coolsoftwaretyler:
Thanks @JuanJo4! I'll take a look today or tomorrow.

__tests__/uitls.test.ts
```diff
@@ -0,0 +1,20 @@
+import { MstError } from "../src/utils"
```
coolsoftwaretyler, line 1: **issue:** @JuanJo4 - would you mind renaming this file to `__tests__/utils.test.ts`?  I think the file name is a typo currently.

__tests__/uitls.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+describe("MstError custom error class", () => {
```
coolsoftwaretyler, line 3: **praise:** @JuanJo4 - thank you for these test! Make sense to me!

coolsoftwaretyler (changes requested):
Hey @JuanJo4 - I love it. I came around to really liking the new class rather than the function-based approach. Initially, I thought `fail` was `throw`ing, but I had misremenbered. Since we were throwing whatever gets returned from `fail`, I think a new class that extends `Error` is perfect.

I found a typo in one of your file names. Would you mind fixing that? Once you do, we can merge this.

Thanks for doing this! Really appreciate it.

__tests__/uitls.test.ts
JuanJo4: oh! my bad... fixed.

coolsoftwaretyler:
Perfect! Merging.

I'll get a new 6.0.0 release out soon. We're in a preview mode for the next major and this should land there once we ship it.

Thanks for the help, and please let us know if there's anything else you'd like to work on.

Issue #2173, opened 2024-04-10
getMembers does not work as expected, behaviour change between 5.1.0 and 5.4.1


cemcakirlar:
**_Bug report_**

-   [X] I've checked documentation and searched for existing issues [and discussions](https://github.com/mobxjs/mobx-state-tree/discussions)
-   [X] I've made sure my project is based on the latest MST version
-   [X] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction.

**Sandbox link or minimal reproduction code**

https://codesandbox.io/p/sandbox/mobx-state-tree-todolist-forked-gn69gv?file=%2Fmodels%2FTodo.js

**Describe the expected behavior**
the code should display keys of props, volatiles, views and actions

```
{
  "_keys_properties": [
    "todos",
    "id",
    "title",
    "finished",
    "nameProp",
    "surnameProp"
  ],
  "_keys_volatile": [
    "newTodoTitle",
    "someVolatile",
    "someOtherVolatile"
  ],
  "_keys_views": [
    "$treenode",
    "toJSON",
    "unfinishedTodoCount",
    "addTodo",
    "handleChange",
    "handleNewTodoClick",
    "remainingTasks",
    "toggle",
    "view",
    "___debugMembers"
  ],
  "_keys_actions": [
    "addTodo",
    "handleChange",
    "handleNewTodoClick",
    "toggle"
  ]
}
```
A screenshot from my development environment for the same code but different more rich model
![image](https://github.com/mobxjs/mobx-state-tree/assets/4405476/39483a61-8527-4870-8a18-14a999b4f26e)

**Describe the observed behavior**

the code cannot  display keys of props, volatiles, views and actions correctly

it mixes actions into views, views into volatiles (not existing in the sample but see attached image, *_calculated props are views but shown in volatiles)

my initial inspection shows it was working fine with version 5.1.0, some code I copied from an old project did not give correct results, I am on 5.4.1

this may be related to types.compose(), a straight simple types.model() seems to be working fine with getMembers(), needs more inspection though.

coolsoftwaretyler:
Thanks for filing this @cemcakirlar, and sorry for the inconvenience!

For you or anyone else interested in figuring this out, here's a comparison of all the changes between the latest preview release and `v5.1.0`: https://github.com/mobxjs/mobx-state-tree/compare/v5.1.0...v6.0.0-pre.2

[...]

PR #2174, opened 2024-04-14
Fix/get members 2173

evinosheaforward:

Fix reported issue with actions being reported as views in getMembers:
https://github.com/mobxjs/mobx-state-tree/issues/2173

issue was bad if/else if causing actions to be returned as views.

The issues 

## Steps to validate locally

For brute force exploring the differences, I took the existing v5.1.0 reflection tests and ran them vs both 5.1.0 and master and compared the views, properties, volatiles, and actions. This did yield a result of extra views on master (extra view for each action), so I started a git bisect looking for the commit where that changed and found [62e7e8bafe8bf07f4694e429f734bb12323fd8ef](https://github.com/mobxjs/mobx-state-tree/commit/62e7e8bafe8bf07f4694e429f734bb12323fd8ef) as the "bad commit".

Then I updated the tests to fail in this PR.
Finally, the fix commit passes the new reflection tests


bun.lockb
evinosheaforward, line 1: Want to confirm this should be updated when I ran `bun install && bun run build`

coolsoftwaretyler:
Thanks @evinosheaforward! I will take a look today or early in the week. I expect we'll merge soon and ship a patch to v5, and include in v6.

evinosheaforward:
> Thanks @evinosheaforward! I will take a look today or early in the week. I expect we'll merge soon and ship a patch to v5, and include in v6.

I'll note that there may be another bug reported in #2173 
this PR fixes `actions -> views` but not `views -> volatiles`
I'll look into the latter and you can lmk if you want to ship the fixes separately or together

src/core/mst-operations.ts
```diff
@@ -891,8 +891,8 @@ export function getMembers(target: IAnyStateTreeNode): IModelReflectionData {
       else reflected.volatile.push(key)
       return
     }
-    if (descriptor.value._isMSTAction === true) reflected.actions.push(key)
     if (descriptor.value._isFlowAction === true) reflected.flowActions.push(key)
+    if (descriptor.value._isMSTAction === true) reflected.actions.push(key)
```
coolsoftwaretyler, line 895: **praise**: this is such a subtle bug. I had to read it a few times over. This has gotten me in other code bases before.

The problem is that we originally constructed a conditional statement like "If this is an MST action, push it into the actions block. Otherwise, if it's observable, we push it into the volatiles, and if that's not true, then this is a view and should be pushed there. The clauses were:

```ts
if (descriptor.value._isMSTAction == true) {
 . . .
 } else if (isObservableProp(target, key)) {
  . . . 
 } else {
  . . . 
 }
```

But when we added the [`_isFlowAction` check](https://github.com/mobxjs/mobx-state-tree/commit/62e7e8bafe8bf07f4694e429f734bb12323fd8ef#diff-3090c65cf4a55928afed579b85369bf9c433e2c9bd615adce4127686a944b56aR874), we made the original `._isMSTAction` no longer part of the `if / else`. So now the clauses are:

```ts
if (descriptor.value._isMSTAction == true) { } // Stands alone

// Separate if/else logic, and an _isMSTAction will also try to evaluate this one.
if (descriptor.value._isFlowAction == true) {
 . . .
 } else if (isObservableProps(target, key)) {
  . . . 
 } else {
  . . . 
 }
 ```
 
 So anything where `_isMSTAction === true` will get pushed into the `actions` array, and then try to evaluate itself in the second block of logic. It falls through to the final `else` clause and gets added to the `views` array.
 
@evinosheaforward thank you for the eyes on this one. I have a feeling this would have taken me far too long to spot.

**suggestion (non-blocking):** @evinosheaforward - On that note, will you please change the formatting of this whole conditional clause to use brackets, rather than the if/else shorthand? I think that will help prevent this issue in the future. As I mentioned in my first comment, I've hit this exact bug in other code bases before, and I find that it usually comes from too much shorthand, which can make it hard to spot the logic flaw.

File: __tests__/core/reflection.test.ts [hunk 2]
```diff
@@ -184,6 +185,8 @@ test("reflection - members chained", () => {
   expect(reflection.actions.includes("anotherAction")).toBe(true)
   expect(reflection.views.includes("viewName")).toBe(true)
   expect(reflection.views.includes("anotherView")).toBe(true)
+  expect(reflection.views.includes("actionName")).toBe(false)
+  expect(reflection.views.includes("anotherAction")).toBe(false)
```
coolsoftwaretyler, line 189: **praise:** @evinosheaforward - thanks for updating the test!

**issue (blocking):** can we also update these tests so they exercise the inline docs listed in `src/core/mst-operations.ts:866`, which say:

>  In the case where a model has two actions: `doSomething` and `doSomethingWithFlow`, where `doSomethingWithFlow` is a flow generator, the `actions` array will contain both actions, i.e. ["doSomething", "doSomethingWithFlow"], and the `flowActions` array will contain only the flow action, i.e. ["doSomethingWithFlow"].

**issue (blocking):** there's also an opportunity for us to exercise that same functionality in the `reflection - model` test in `__tests__/core/reflection.test.ts:59`

Let me know if you want a clearer suggestion on what to do with those, but I'd like to take this moment to improve our test coverage around this feature while we're looking at it.

coolsoftwaretyler (changes requested):
Hey @evinosheaforward - thanks again for this great PR! It definitely fixes the "actions in view reflection" part of #2173.

I think [the view/volatiles part](https://github.com/mobxjs/mobx-state-tree/issues/2173#issuecomment-2054097859) might be a separate issue. I'm inclined to handle that in a new PR to keep these things focused. I'll try and put together a test case for that one and respond in the original issue.

With that said, I have a few small requests to finish up this PR. I marked them as blocking because I want to get those changes in before merge, but if you don't have the time to address them this week, let me know and I can handle them.

src/core/mst-operations.ts
evinosheaforward: Yeah, very easy bug to write.
Since I was coming at it from the perspective of _knowing the bug_ it stood out.

I updated the syntax/formatting for clarify, good idea.

__tests__/core/reflection.test.ts
evinosheaforward: Thanks for the feedback!
>issue (blocking): there's also an opportunity for us to exercise that same functionality in the reflection - model test in __tests__/core/reflection.test.ts:59

I added the explicit test that the flow action is added to `actions` in the `reflection - model` test, good call.

>issue (blocking): can we also update these tests

I'd be glad to update this test, but it doesn't currently have any flow actions.
If you'd like me to add one, let me know if there the example from the other test would work:
```
  .actions((self) => {
    function actionName() {
      return 1
    }
    return {
      actionName,
      generatorAction: flow(function* generatorAction() {
        const promise = new Promise((resolve) => {
          resolve(true)
        })
        yield promise
      })
    }
  })
```

coolsoftwaretyler: Yup, that'd be perfect, thank you!

evinosheaforward: addressed this!

I also added a little line to confirm that an action associated with a flowaction is not added to the flowactions in getMembers. I saw no downside and potential upside, if you disagree just lmk!

coolsoftwaretyler: Nailed it! Will merge and get a patch release out sometime this week!

PR #2189, opened 2024-07-05
Improve typing for model type substitutability.

### LINKED ISSUE #2186: Broken model types in 6.0.0 version
**_Bug report_**

```ts
const BaseModel = types.model();
type BaseModelType = typeof BaseModel;

const DerivedModel = BaseModel.props({ a: '' });

const test: BaseModelType = DerivedModel;
```

No type errors

```
Type 'IModelType<{ a: IType<string | undefined, string, string>; }, {}, _NotCustomized, _NotCustomized>' is not assignable to type 'IModelType<{}, {}, _NotCustomized, _NotCustomized>'.
  The types of 'views(...).CreationType' are incompatible between these types.
    Type 'ModelCreationType<ExtractCFromProps<{ a: IType<string | undefined, string, string>; }>>' has no properties in common with type 'IStateTreeNode<IAnyType>'.ts(2322)
```

It has been broken in 6.0.0 version. It works in 5.4.1 version as expected.

thegedge:
Previously, a model derived from an empty model couldn't be used in a place that asked for the empty model. In general, we had a lot of substitutability issues (just with our typing).

This commit fixes that by bringing back what was previously known as "$nonEmptyObject", but improves on it by not having it be as pervasive as it was previously (for example, it would show up on models that had props).

Fixes #2186 

This PR fixes regressions introduced (by me! 😅) in the previous major release that attempted to remove `$nonEmptyObject` from all types. Little did we know, it did support stricter typing in some cases.

This PR reintroduces a similar concept, but it does so in a far more limited form to fix the aforementioned issue.

## Steps to validate locally

The test is a good example, and validating would simply be running the type checker.

### DIFF
--- a/__tests__/core/type-system.test.ts
+++ b/__tests__/core/type-system.test.ts
@@ -18,7 +18,8 @@ import {
   IAnyType,
   ModelPrimitive,
   ModelPropertiesDeclaration,
-  SnapshotOut
+  SnapshotOut,
+  type ISimpleType
 } from "../../src"
 import { expect, test } from "bun:test"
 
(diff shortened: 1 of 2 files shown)


coolsoftwaretyler:
Thanks, @thegedge! I'll take a look this weekend and merge/release a patch shortly.

Issue #1101, opened 2018-12-06
TS+ES6 error: ObservableMap$$1 cannot be invoked without 'new'


andrey-tectonic:
**_Bug report_**

* [x] I've checked documentation and searched for existing issues
* [x] I've made sure my project is based on the latest MST version
* [ ] _(not applicable)_ Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

TypeScript project with `compilerOptions.target: "es6"` gives the error in `MSTMap` constructor:

```
Class constructor ObservableMap$$1 cannot be invoked without 'new'
```

MST npm package contains only ES5 code. `MSTMap` extends `mobx.ObservableMap` with old-fashioned way without class syntax.

```js
function MSTMap(initialData) {
    return _super.call(this, initialData, mobx.observable.ref.enhancer) || this;
}
```

and this gives the error when `mobx.es6.js` used.

Temporary solution that worked for me - set `compilerOptions.target` to `"es5"`. I suggest it's possible to include es6 version of MST in npm package (same way as mobx.es6.js).

Carduelis:
The same and it's still not resolved.

Also, for some reason I bumped to the issue of that all maps are `undefined` by default instead of empty Map objects, so it leads to error in `values`  function which does not accept `undefined` as an argument

mweststrate:
all es5 should be valid es6 I'd expect. Please create a small repository that demonstrates the problem.

andrey-ridge:
Demo repo: https://github.com/andrey-ridge/mst-1101-demo

Please refer to `README.md` to build and run demo. Expected output should be like this:

```
Testing TodoMap with ES_VER=5...OK
Testing TodoMap with ES_VER=6...
TypeError: Class constructor ObservableMap cannot be invoked without 'new'
```

coolsoftwaretyler:
@chakrihacker - I think we should leave this open. I'm seeing errors from the reproduction project, and I'd like to support this if we could.

It's been a while and not a ton of activity here, but that doesn't mean there aren't folks out there who might need it, and I'm guessing we can fix this with some kind of build config without actually changing library code?

PR #2197, opened 2024-07-14
chore: target es6 in tsconfig

coolsoftwaretyler:

Fixes https://github.com/mobxjs/mobx-state-tree/issues/1101 by upgrading our target to `es6`.

## Steps to validate locally

[Repro repo should work](https://github.com/andrey-ridge/mst-1101-demo/tree/master).

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -18,7 +18,7 @@
     "importHelpers": true,
     "stripInternal": true,
     "downlevelIteration": true,
-    "lib": ["ES2016"],
+    "lib": ["es6"],
     "useDefineForClassFields": true,
     "isolatedModules": true,
     "skipLibCheck": true
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "outDir": "lib/",
-    "target": "es5",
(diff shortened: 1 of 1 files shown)


coolsoftwaretyler:
Ok, so it didn't work with these changes, but the `.module` export does work if we update our target to `ES6` in `tsconfig.json`.

tsconfig.json
```diff
@@ -1,10 +1,10 @@
 {
   "compilerOptions": {
     "outDir": "lib/",
-    "target": "es5",
+    "target": "es6",
     "sourceMap": false,
     "declaration": true,
-    "module": "es2015",
+    "module": "es6",
```
coolsoftwaretyler, line 7: I just changed this for consistency's sake, but the `es2015 === es6`

coolsoftwaretyler:
I think we're actually gonna ship two tsconfig files and build separately so we can maintain backwards compatibility with older MST versions. Should be able to just ship this build alongside it.

coolsoftwaretyler:
I'm going to close this for now and revisit once we get a few of the other pending PRs merged in since we're changing up a little in tsconfig anyway. My goal is to put out a new, additional build with an upcoming v7 preview.

PR #2198, opened 2024-07-14
Fix typings for snapshot processor

thegedge:

Previously, we weren't properly pulling out the input/output snapshot types of the referenced type in snapshot processors, forcing explicit generic argument specification or typing on the processor functions to avoid `_NotCustomized`:

CleanShot 2024-07-14 at 13 33 08@2x

This PR fixes that issue by choosing the given types input/output snapshot if the snapshot processor isn't customized.

Two other small additions in here:
1. `bun.lockb` appeared to be out of date, so I added the updates
2. Fixed some docs that had `xyzProcess` instead of `xyzProcessor`

## Steps to validate locally

Tests added to validate.


coolsoftwaretyler:
Will take a look this week!

coolsoftwaretyler (changes requested):
Hey @thegedge, three things: 

1. **praise:** thank you for fixing up the docs! That helps a lot.
2. **nitpick**: I added the `describe` import from `bun:test` so we can use that with TS. I think that may actually be why you found you wanted `@types/jest` in https://github.com/mobxjs/mobx-state-tree/pull/2200. We no longer use Jest. `bun:test` is *mostly* a drop in replacement with its own types, but we do need to actually import from that namespace. I pushed that up.
3. **issue**: I'm seeing a type error in the `produces the right types when not customized` test:

Screenshot 2024-07-21 at 2 37 43 PM

This goes away if instead we write either:

```ts
        assertTypesEqual(snapshot, _ as ModelCreationType<{ name: string | undefined }>)
```

Or


```ts
        assertTypesEqual(snapshot, _ as SnapshotIn<typeof Model>)
```

I'm partial to the second one since that's more accessible to users. But I wanted to check to see if you also see this issue or not, and if the `SnapshotIn<>` solution changes what we're actually asserting here.

Let me know what you think. Happy to merge this in once we are aligned on this TS error in the test file.

thegedge:
I'm partial to the second one too!

coolsoftwaretyler:
Hey @thegedge - ready for another look on my end?

thegedge:
Oh yeah, totally

coolsoftwaretyler:
LGTM! I'm gonna merge this and target it as part of the v7 release because I think TS changes are a little nicer behind major versions, even though this is mostly fixing a problem rather than making a sweeping change.

thegedge:
> LGTM! I'm gonna merge this and target it as part of the v7 release because I think TS changes are a little nicer behind major versions, even though this is mostly fixing a problem rather than making a sweeping change.

I think that's fair, given this is a prerequisite for https://github.com/mobxjs/mobx-state-tree/pull/2199, which is a far more significant typing change!

Issue #1764, opened 2021-08-17
Type issue when assigning a store with a reference


RickLucassen:
**_Bug report_**

* [x] I've checked documentation and searched for existing issues
* [x] I've made sure my project is based on the latest MST version
* [x] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

**Sandbox link or minimal reproduction code**
https://codesandbox.io/s/mobx-state-tree-todolist-forked-o9yj3

**Describe the expected behavior**
In the example we're defining a root store with another store as a property. That store has a property of type array of a model.
When that model doesn't have a types.reference present everything is fine. However, when the model does have a types.reference present typescript complains about an incorrect type.

Typescript should resolve these types correctly.

**Describe the observed behavior**
As seen in the `rootStore`, where there is a version with and without the reference, the one with the reference generates a typescript error about the `someModel` type not matching.

RickLucassen:
It seems that wrapping the `createTodoStoreWithReference` with `castToSnapshot` does make it work. Would this be the proposed solution?

Still strange that without the `castToSnapshot` there is a difference in behavior between the two stores (with and without a reference).

coolsoftwaretyler:
Hey @RickLucassen - sorry it took so long for anyone to get back to you here. We are trying to give some more care and attention here to MobX-State-Tree

I don't know for 100% sure that `castToSnapshot` is the best case solve, but if it did help your issue, that makes sense. There are some known issues with the TypeScript inferences in MST, and we're highly interested in fixing that.

If you're still having the problem, or if you solved it elsewise, I'd love to hear what worked for you. Or if you moved on, no worries.

For now, I'll leave this issue and label with with TS and help/PR welcome. Hopefully we can find some ways to improve this moving forward.

thegedge:
Curious: is this still an issue? I'm not able to see any type errors:

![CleanShot 2024-07-06 at 13 51 45@2x](https://github.com/mobxjs/mobx-state-tree/assets/833576/8a920b1e-08db-46e8-b9f0-b07bc2bc8058)

coolsoftwaretyler:
@thegedge - I see this in the reproduction CodeSandbox:

<img width="408" alt="Screenshot 2024-07-07 at 12 59 03 PM" src="https://github.com/mobxjs/mobx-state-tree/assets/29899554/64e065fe-c482-4ee8-be3c-0f3f89323f43">

[...]

Issue #1568, opened 2020-08-18
Error with reference types after updating to latest version


peejayess:
**_Bug report_**

* [x ] I've checked documentation and searched for existing issues
* [x ] I've made sure my project is based on the latest MST version
* [ x] Fork [this](https://codesandbox.io/s/mobx-state-tree-todolist-forked-3060p?file=/index.ts) code sandbox or another minimal reproduction. 

**Sandbox link or minimal reproduction code**
https://codesandbox.io/s/mobx-state-tree-todolist-forked-3060p?file=/index.ts
Note:  You will have to wait for a few seconds for the error to be highlighted in the index.ts file

**Describe the expected behavior**
I have a working application where  I have the following code:

```
export const ProjectStore = types.model("ProjectStore", {
  projects: types.array(Project),
  currentProject: types.maybeNull(types.reference(Project))
});
```

```
export const RootStore = types.model("RootStore", {
  projectStore: ProjectStore
});
```

```
const rootStore = RootStore.create({
  projectStore: ProjectStore.create()
});
```

**Describe the observed behavior**
After upgrading to the latest version of MST (3.17.2) I now have a type checking error at:

` projectStore: ProjectStore.create()`

which is caused by line:

`currentProject: types.maybeNull(types.reference(Project))`

The error message is:

[...]

PR #2199, opened 2024-07-20
 Type create/validate to accept an instance type 

thegedge:

We've had issues with typings when instantiating references

CleanShot 2024-07-20 at 11 42 42@2x

https://github.com/mobxjs/mobx-state-tree/issues/1568 for example, and maybe even https://github.com/mobxjs/mobx-state-tree/issues/1764.

This PR resolves that by typing `create` to accept an instance type. I've added a "kitchen sink" test for create to test both snapshots and instances.

## Steps to validate locally

The test added in this PR will type issues (like in above the screenshot) if you revert 6176e94efbcf1094c846ef2c4f19fd0a08249d01.

### DIFF
--- a/src/core/type/type.ts
+++ b/src/core/type/type.ts
@@ -24,6 +24,7 @@ import {
   getStateTreeNodeSafe,
   assertArg
 } from "../../internal"
+import type { Writable, WritableKeys } from "ts-essentials"
 
 /**
  * @internal
@@ -70,6 +71,8 @@ export type STNValue<T, IT extends IAnyType> = T extends object ? T & IStateTree
 /** @hidden */
 const $type: unique symbol = Symbol("$type")
 
+type ExcludeReadonly<T> = T extends {} ? T[WritableKeys<T>] : T
+
 /**
  * A type, either complex or simple.
  */
@@ -93,7 +96,7 @@ export interface IType<C, S, T> {
    *
    * @returns An instance of that type.
    */
-  create(snapshot?: C, env?: any): this["Type"]
+  create(snapshot?: C | ExcludeReadonly<T>, env?: any): this["Type"]
 
   /**
    * Checks if a given snapshot / instance is of the given type.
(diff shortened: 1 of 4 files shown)


thegedge:
Hold off. I realized our tsconfig isn't properly typechecking tests, and this introduces a typing bug: it allows passing in a value for a view in model types.

Technically this works (it just gets ignored) but I'm thinking ideally we make it fail typechecking so users avoid making that error. This is actually tough to do, but I think we can do this by marking the getters as readonly and excluding those from the instance type in `create`.

I'm going to play with that idea, and if it works out, great! If it doesn't, I think I may remove the one test that's failing:

CleanShot 2024-07-20 at 12 16 39@2x

thegedge:
Will rebase this after https://github.com/mobxjs/mobx-state-tree/pull/2200 merged, which should fix the typechecking / build stuff

src/core/type/type.ts
```diff
@@ -93,7 +96,7 @@ export interface IType<C, S, T> {
    *
    * @returns An instance of that type.
    */
-  create(snapshot?: C, env?: any): this["Type"]
+  create(snapshot?: C | ExcludeReadonly<T>, env?: any): this["Type"]
```
thegedge, line 99: This allows this test to continue passing:

https://github.com/mobxjs/mobx-state-tree/blob/c2f459b6126d4d1a990837541ced7f81f1490033/__tests__/core/type-system.test.ts#L1249-L1250

In other words, we want the type checker to let users know when they pass in a value for a getter view. I'm not sure whether or not this is the right decision, but this change is an addition (i.e., is backwards compatibile) to existing type checking, so I say we go for it and relax it if users encounter issues with it.

coolsoftwaretyler:
Hey @thegedge - just merged #2200 so you should be good to rebase and then figure out the new typing issues. Let me know when this is ready for another look.

thegedge:
@coolsoftwaretyler this is good for review now! <3

I was thinking we could also get this one in for https://github.com/mobxjs/mobx-state-tree/pull/2201

coolsoftwaretyler:
Perfect! Great idea. Will take a look sometime this week and get this into the next preview release.

coolsoftwaretyler:
Alright, looks excellent and we will get this into the v7 preview. I think we'll consider this a breaking change, even though it should mostly be backwards compatible. I think type changes are just best treated as breaks overall, in case people are relying on the old type behavior.

PR #2200, opened 2024-07-20
Fix tsconfig and typechecking

thegedge:

This PR does two things:
1. Updates `tsconfig.json` so we start typechecking tests, and
2. adds `@types/jest`, which was missing and noticed after (1).

## Steps to validate locally

* Check out the first commit in this PR
* `rm -rf node_modules`
* `bun install`
* `bun typecheck`

That should fail. Check out the `HEAD` of this branch and then `bun install` and typecheck again, everything should pass.


tsconfig.json
```diff
@@ -23,6 +23,5 @@
     "isolatedModules": true,
     "skipLibCheck": true
   },
-  "files": ["src/index.ts"],
-  "include": ["**/*.{js,ts}"]
```
thegedge, line 27: Not sure why, but `{a,b}` doesn't appear to work. I get

> error TS18003: No inputs were found in config file '/Users/gedge/src/github.com/mobxjs/mobx-state-tree/tsconfig.json'. Specified 'include' paths were '["**/*.{js,ts}"]' and 'exclude' paths were '["lib/"]'.

I also dropped `files`, because it seemed redundant with "include", but that changed how things get emitted, hence the changes to the build. Do we have any good tests to verify integrity of the build?

package.json
```diff
@@ -56,6 +56,7 @@
     "@size-limit/preset-big-lib": "^5.0.3",
     "@types/bun": "^1.0.6",
     "@types/node": "^12.0.2",
+    "@types/jest": "^29.5.12",
```
coolsoftwaretyler, line 59: **issue:** we don't actually use Jest. We're using [bun test](https://bun.sh/docs/cli/test), which is mostly drop-in compatible with Jest, and the Bun runner will sort of just smooth over missing imports at runtime.

But to fix the TS error you get with `bun run build`, instead we should add imports to the failing files like `__tests__/core/1525.test.ts` with:

```ts
import { describe, it } from "bun:test"
````

I'll take care of that since it's pretty straightforward and I'm looking at it now. It's awesome that we're type checking the tests now, but we can get it to work from Bun rather than re-importing Jest dependencies. Will push that up shortly.

Comment (coolsoftwaretyler) [reply package.json]: Updated with b687398d2cbb42a536d5a309ba3d7b9b6031c5b4, and I found a few places where we needed `expect.fail()` instead of `fail()`. `bun run build` and `bun run test:all` still works on that commit.

File: tsconfig.json (hunk 1 above)
Comment (coolsoftwaretyler) [reply tsconfig.json:27]: Not that I know of, unfortunately. I think some of the tests run against the build itself, so I think that:

```sh
bun run build
bun run test:all
```

Is generally indicative of it working. But I am not very confident in this.

The output files look basically the same when I compare the file structure between `master` and this branch. I'm inclined to just merge this and pull it into the `7.0.0` release so we can use the preview builds and breaking semantics to protect against any regressions here without impacting current users who are auto-updating minor/patch versions.

coolsoftwaretyler:
I'm going to merge this so we can unblock https://github.com/mobxjs/mobx-state-tree/pull/2199.

Issue #1627, opened 2020-12-16
Union dispatcher snapshot undefined


vallode:
**_Question_**

* [x] I've checked documentation and searched for existing issues
* [ ] I tried the [spectrum channel](https://spectrum.chat/?t=dad48299-3dfc-4e10-b6da-9af1e39498a3)

In our MST/Typescript project we like to use `Model.is(...)` quite a bit, this sometimes results in errors like these popping up:

```
Cannot read property 'exampleProperty' of undefined
```
```js
1 |  dispatcher: (snapshot) => snapshot.exampleProperty || types.null,
  |                                    ^
```

This is _usually_ caused by not enforcing the passed model to be non-nullable via an extra check like:  
```js
exampleObject && Model.is(exampleObject)
```
It's easily fixed but not always caught by people due to the rather "mysterious" error thrown and no typing error.

Is there a way to enforce `Model.is` to throw type errors with undefined values? Would this be something that MST would have to change in their typings or maybe the inference is already there but I am not using it correctly. (I'd be happy to create an PR).

coolsoftwaretyler:
Hey @vallode - sorry it took anyone so long to get back to you here in the MST project.

Unfortunately I don't have an answer for you at the moment, but I will add some labels to help us as we clean up the maintenance overall, and hopefully we can get this slated for a fix.

I know it's been literally years, so I understand if you aren't interested/have moved on, but did you ever find a fix on your own? Are you still interested in submitting a PR? We're definitely open to improvements in the typing system.

vallode:
Hey, no worries. I'd have to try and dig out the codebase this was a part of in order to look back into it, I remember the issue being quite simple.

Changing:
```ts
dispatcher: (snapshot) => snapshot.exampleProperty || types.null,
```

To:
```
dispatcher: (snapshot) => snapshot?.exampleProperty ?? types.null,
```

Would likely resolve this, happy to open a PR and play around with it but give me a little time ;)

PR #2203, opened 2024-08-04
Improve typing of the union dispatcher function

thegedge:

Closes https://github.com/mobxjs/mobx-state-tree/issues/1627

Currently, the dispatcher function for union types has two issues with the typing on the dispatcher function:

1. The input snapshot is typed as any.
2. It expects any type to be returned.

This PR improves both of these by

1. typing the input snapshot so that it's an input snapshot for the union types, and
2. it restricts the possible types returned from the dispatcher to the types specified in the union.

## Steps to validate locally

As always, this is all about the typechecking, so one would just have to play around with `types.union` and ensure it fails when you expect it to fail, and passes when you'd expect it to pass.

I've included a test to capture two key scenarios, but let me know if there should be more!

### DIFF
--- a/__tests__/core/type-system.test.ts
+++ b/__tests__/core/type-system.test.ts
@@ -1,5 +1,3 @@
-import { Exact } from "ts-essentials"
-
 import {
   types,
   getSnapshot,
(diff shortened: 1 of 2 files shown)


coolsoftwaretyler:
Thanks @thegedge! I'll take a look soon. Will you please resolve the merge conflict? On my phone right now and GitHub is not showing precisely what's in conflict.

thegedge:
Ahh, I think `master` on my fork wasn't up to date, so was just a conflict on the addition of the test. Fixed!

__tests__/core/type-system.test.ts
```diff
@@ -1,5 +1,3 @@
-import { Exact } from "ts-essentials"
```
thegedge, line 1: I didn't realized it, but `Exact` from `ts-essentials` isn't really "exact"! It's just "if A and B are record types, the shapes of the two are the same", which is definitely not what we wanted. My bad!

(TS is hard)

coolsoftwaretyler:
Thank you! Will take a look soon!

PR #2204, opened 2024-08-04
Get `bun test` working locally

thegedge:

Before this PR, `bun test` would have a ton of failures if you ran it locally. We didn't see this in CI because our runners would test each file separately from one another.

After doing some digging, I found a couple of issues:
1. AFAICT, `bun test` maintains globals across modules (unlike jest), so you have to be careful to reset it.
2. Sometimes tests would hang because failures wouldn't call the `done` callback.
3. Some other issue that I could not explain with this test: https://github.com/mobxjs/mobx-state-tree/blob/fc60d76b3725e053b89532e6f26804bf9e893d75/__tests__/core/model.test.ts#L212-L230

And the various solutions:

1. Introducing a setup script that ensures we reset state before/after each test.
2. Switch to `await` / `async` vs callbacks.
3. Check for overlapping view/properties when instantiating the views vs discovering this issue when getting the atoms, which we were doing here: https://github.com/mobxjs/mobx-state-tree/blob/fc60d76b3725e053b89532e6f26804bf9e893d75/src/types/complex-types/model.ts#L651-L658

## Steps to validate locally

All of the following should succeed:

- `bun test` 
- `bun run test:dev`
- `bun run test:prod`


src/types/complex-types/model.ts
```diff
@@ -648,13 +652,10 @@ export class ModelType<
   getSnapshot(node: this["N"], applyPostProcess = true): this["S"] {
     const res = {} as any
     this.forAllProps((name, type) => {
-      try {
-        // TODO: FIXME, make sure the observable ref is used!
-        const atom = getAtom(node.storedValue, name)
-        ;(atom as any).reportObserved()
-      } catch (e) {
-        throw new MstError(`${name} property is declared twice`)
```
thegedge, line 656: Perhaps the change that needs the most attention in this PR is this one. As mentioned in the description, something about throwing here was causing some kind of state leakage across test files.

By moving the "declared twice" check to `instantiateViews`, everything started working again, but I guess the reality is that we probably need this scattered about all of the `instantiateXYZ` functions and not just `views`  🤔 

Any ideas on what kind of state could be leaking between test modules due to this?

coolsoftwaretyler:
Ooh very exciting! Thanks for figuring this out! Will take a look soon.

package.json
```diff
@@ -17,8 +17,8 @@
   "scripts": {
     "clean": "shx rm -rf dist lib",
     "build": "bun run clean && tsc && cpr lib/src dist --filter=\\.js$ && rollup -c",
-    "test:dev": "./scripts/bun-test-dev.sh",
```
coolsoftwaretyler, line 20: **praise:** thank you for also cleaning up these commands/scripts!

src/types/complex-types/model.ts
coolsoftwaretyler: I'd have to experiment to see exactly what's going on, but what if `.reportObserved()` here is important in some other function elsewhere, or part of some cleanup task, and by throwing on `getAtom`, we leave something hanging that was expecting to complete?

That's just speculation, but it stands to reason if throwing earlier is preventing it, and there's two lines in the try block, that a problem in line 1 and the subsequent discarding of line 2 would be a problem.

Let's see if we can figure out the difference here before shipping. I've definitely shipped things I didn't understand and broken folks before, haha.

If you get around to figuring it out, let me know. Otherwise I can do more investigation when I'm back from my trip next week.

thegedge: <3 I'm definitely down for understanding this a bit more.

I'll investigate a bit more on the weekend. Main thing is that there's a reproducible command:

```
bun test __tests__/core/model.test.ts __tests__/core/env.test.ts
```

(after reverting 0ec3bcc9c9f60f49dcb52b29a6903e4e9940aaf7)

thegedge: Okay, I think I've narrowed it down: it's something related to not properly cleaning up the snapshot reaction on the root when something throws while creating an object node.

thegedge: I can fix it by ensuring the disposer is called, but unfortunately breaks one test that expects a certain error message when a view has the same name as prop:

```
224 |       expect(() =>
225 |         UserModel.create({
226 |           id: "chakri",
227 |           name: "Subramanya Chakravarthy"
228 |         })
229 |       ).toThrow("[mobx-state-tree] name property is declared twice")
              ^
error: expect(received).toThrow(expected)

Expected substring: "[mobx-state-tree] name property is declared twice"
Received message: "[mobx-state-tree] Node not available for property name"
```

thegedge: So I was able to "sort of" fix that case, but then added a test for where an action name overlaps a prop and got something completely different.

@coolsoftwaretyler I'm somewhat convinced what's already here is the right fix. My line of thinking is that it's impossible to create any kind of instance with these "broken" models, so it's best to fail early rather than late. Any "observing" done during initialization won't do anything, because nothing ever gets far enough to trigger any kind of event listener or other user code (AFAICT).

I might even go a bit further and say that the `views` and `actions` functions should do the validation, not even `instantiateXYZ`. When doing so, we can assume that `model.create` will always produce a valid instance with no conflicts, which would in turn may allow us to drop various safety checks we do post-instantiation (good for perf!).

If you're game for that, we can go down this route (I'll update to ensure actions and volatiles also don't override any props). If you want to take a smaller step, I can push my other change that sprinkles some more checks and ensures the snapshot reaction is also cleaned up when initialization fails.

coolsoftwaretyler: Thanks so much, @thegedge, and sorry for the late response. I mentioned it in Discord but figured I'd flag publicly: I had been traveling and then picked up a pretty nasty illness. Recovering now and finally getting around to things.

I am very interested in the idea of skipping validation on instantiation. Although I'd like to explore that in a separate issue or PR. I'm concerned about use cases (like my own) where people chain `.actions` and `views` calls for TypeScript reasons and end up paying a higher perf cost. But we can benchmark that.

For now, I feel satisfied here and like I understand the changes well enough to merge.

Thank you so much for getting our bun tests working as they should. This is going to be a huge boon.

thegedge: <3

@coolsoftwaretyler since you're game with this change, we do need to follow up and apply the same logic for actions and volatiles too (before we set up a new release). I'll make that happen this weekend.

I'll also add a bunch of tests for that too, and once we have an RC ready, try it out on Gadget's CI to make sure everything is gucci. I don't think there's any rush to get an RC out there, since we're optimizing our own DX right now, but let's make a mental note to call out this change in the release notes :)

Issue #2216, opened 2024-10-11
TypeScript does not recognize overriding mandatory property with optional one in sub-model


andreykurilin:
**Minimal reproduction code**

```typescript

import { types } from 'mobx-state-tree';

export const ErrorStore = types
  .model('ErrorStore', {
    value: '',
  })
  .actions((self) => ({
    set(value: string) {
      self.value = value;
    },
    reset() {
      self.value = '';
    },
  }))
  .views((self) => ({
    get formattedValue() {
      return self.value;
    },
  }));

export const NoDisplayError = ErrorStore.named('NoDisplayError').views(() => ({
  get formattedValue() {
    return undefined;
  },
}));


const InputStore = types.model("InputStore", {
    value: "",
    // developer should explicitly specify whether validation errors should be displayed or not
    error: types.union(ErrorStore, NoDisplayError)
});


const PasswordInputStore = InputStore.named("PasswordInputStore").props({
    // we have strict requirements for password, so validation errors should be displayed by default
    error: types.optional(types.union(ErrorStore, NoDisplayError), () => ErrorStore.create({}))
});


const passwordS = PasswordInputStore.create({});
```

**Describe the expected behavior**

Initialization of `PasswordInputStore` should not require specifying optional `error` property.

**Describe the observed behavior**

TypeScript fails as `error` property is not provided.

```error
      TS2345: Argument of type '{}' is not assignable to parameter of type 'ModelCreationType<{ value: string | undefined; error: EmptyObject & Partial<{ value: string | undefined; }>; }>'.
  Property 'error' is missing in type '{}' but required in type '{ error: EmptyObject & Partial<{ value: string | undefined; }>; }'.

```

coolsoftwaretyler:
Thanks for the issue report @andreykurilin, and I'm sorry for the inconvenience!

I see @thegedge has assigned this to himself, so I expect good things here. 

I put together a CodeSandbox with the repro code in case that helps demonstrate the issue for others: https://codesandbox.io/p/sandbox/g3w8ft?file=%2Fsrc%2Findex.ts%3A8%2C15

thegedge:
Sorry for the delay, but it's been really busy for me lately.

Thought I'd explore what's happening here:

![CleanShot 2024-11-09 at 10 16 51@2x](https://github.com/user-attachments/assets/7dbed15d-a0d6-4f7e-adf1-2de0feb51192)

[...]

PR #2218, opened 2024-11-09
Fix model types when extending in any way.

thegedge:

Fixes https://github.com/mobxjs/mobx-state-tree/issues/2216

Previously, we used to have TS resolve the extensions by just doing something like `A & B`. The problem is that in TypeScript that means "it's both A and B", but what happens if you have a required prop in one and an optional prop in the other? Well, the least common denominator wins, which is requiredness, because it satisfies both the required and optional properties.

MST doesn't work that way though. Whenever we extend a model we "clobber" the old property of the same name. To model this in TS we need to omit properties from the previous version of the model and replace them with the newer ones.

## Steps to validate locally

The added test should capture both the essence of the aforementioned issue, while also capturing other issues we have.

### DIFF
--- a/__tests__/core/type-system.test.ts
+++ b/__tests__/core/type-system.test.ts
@@ -1,38 +1,37 @@
+import { describe, expect, test } from "bun:test"
 import {
-  types,
-  getSnapshot,
-  unprotect,
-  getRoot,
-  getParent,
-  SnapshotOrInstance,
-  cast,
-  SnapshotIn,
-  Instance,
-  castToSnapshot,
-  IType,
-  isStateTreeNode,
-  isFrozenType,
-  TypeOfValue,
   IAnyType,
+  IType,
+  Instance,
   ModelPrimitive,
   ModelPropertiesDeclaration,
+  SnapshotIn,
+  SnapshotOrInstance,
   SnapshotOut,
-  type ISimpleType,
-  isOptionalType,
-  isUnionType,
-  type IOptionalIType,
-  type ITypeUnion,
-  isMapType,
+  TypeOfValue,
+  cast,
+  castToSnapshot,
+  getParent,
+  getRoot,
+  getSnapshot,
   isArrayType,
-  isModelType,
+  isFrozenType,
+  isIdentifierType,
+  isLateType,
   isLiteralType,
+  isMapType,
+  isModelType,
+  isOptionalType,
   isPrimitiveType,
   isReferenceType,
-  isIdentifierType,
   isRefinementType,
-  isLateType
+  isStateTreeNode,
+  isUnionType,
+  types,
+  unprotect,
+  type IOptionalIType,
+  type ISimpleType
 } from "../../src"
-import { describe, expect, test } from "bun:test"
 import type {
   DatePrimitive,
   IAnyComplexType,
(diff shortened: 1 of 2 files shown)


src/types/complex-types/model.ts
```diff
@@ -458,7 +468,7 @@ export class ModelType<
     return this.cloneAndEnhance({ properties })
   }
 
-  volatile<TP extends object>(fn: (self: Instance<this>) => TP) {
```
thegedge, line 461: Changing all of these to be clearer. `object` is generally not what one wants.

CleanShot 2024-11-09 at 11 02 56@2x

I think what we want is only the last.

src/types/complex-types/model.ts
```diff
@@ -199,23 +203,28 @@ export interface IModelType<
...
-  ): IModelType<PROPS, OTHERS & TP, CustomC, CustomS>
+  volatile<VS extends AnyObject>(
+    fn: (self: Instance<this>) => VS
+  ): IModelType<PROPS, OmitMerge<OTHERS, VS>, CustomC, CustomS>
 
-  extend<A extends ModelActions = {}, V extends Object = {}, VS extends Object = {}>(
+  extend<A extends ModelActions = {}, V extends AnyObject = {}, VS extends AnyObject = {}>(
     fn: (self: Instance<this>) => { actions?: A; views?: V; state?: VS }
-  ): IModelType<PROPS, OTHERS & A & V & VS, CustomC, CustomS>
+  ): IModelType<PROPS, OmitMerge<OTHERS, A & V & VS>, CustomC, CustomS>
```
thegedge, line 227: This is the fix for the issue, but I changed the above to resolve a similar issue.

Technically we likely want to deal with this in `compose` too, but I'd rather deal with that when I make `compose` use variadic tuples in the generic signature.

coolsoftwaretyler:
Thanks @thegedge! Will take a look this weekend.

__tests__/core/type-system.test.ts
```diff
@@ -1,38 +1,37 @@
+import { describe, expect, test } from "bun:test"
```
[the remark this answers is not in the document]
coolsoftwaretyler: @thegedge - I strongly agree with this. We still haven't released v7 yet. Here's what I'd like to do:

1. Let's merge this PR (I am still reviewing but it looks good so far)
2. I will *actually* release the next major version, including this
3. Once we've released that version, let's do a big formatting PR. I have no strong opinions about formatting other than it should be automatic, checked before merge, and consistent. So we can just use whatever settings you have if you want the PR, or I can put the PR together and use whatever settings I have lol.
4. We can merge that PR right behind the major version, which should keep future patches/minors/majors easy to check berfore/after the big formatting PR. Basically making it easy to see what happened between v6 and v7, and then changes inside of 7 itself.

I put together an issue here: https://github.com/mobxjs/mobx-state-tree/issues/2219

__tests__/core/type-system.test.ts
```diff
@@ -1403,3 +1402,84 @@ test("#1627 - union dispatch function is typed", () => {
...
+        return self.value
+      },
+
+      get eggs(): string {
+        return self.value
+      }
+    },
+    state: {
+      anotherValue: "test" as string,
+      soManyValues: "test" as string
```
coolsoftwaretyler, line 1430: **praise:** many values, very wow.

src/types/complex-types/model.ts
```diff
@@ -74,6 +73,9 @@ export interface ModelPropertiesDeclaration {
   [key: string]: ModelPrimitive | IAnyType
 }
 
+/** intersect two object types, but omit keys of B from A before doing so */
+type OmitMerge<A, B> = Omit<A, keyof B> & B
```
coolsoftwaretyler, line 77: **praise:** NICE

src/types/complex-types/model.ts
coolsoftwaretyler: I definitely agree here.

coolsoftwaretyler (approved):
Love it. Thank you very much!

Will merge and get this into the next breaking change

Issue #2219, opened 2024-11-10
Add formatting rules


coolsoftwaretyler:
Just a small issue to capture [this comment](https://github.com/mobxjs/mobx-state-tree/pull/2218#discussion_r1835440649): we need to add auto-formatting to this repository so that all PRs are consistent and to reduce some of the noisy formatting diffs we tend to get.

coolsoftwaretyler:
Here's the plan:

1. Let's merge this PR (I am still reviewing but it looks good so far)
2. I will actually release the next major version, including this
3. Once we've released that version, let's do a big formatting PR. I have no strong opinions about formatting other than it should be automatic, checked before merge, and consistent. So we can just use whatever settings you have if you want the PR, or I can put the PR together and use whatever settings I have lol.
4. We can merge that PR right behind the major version, which should keep future patches/minors/majors easy to check berfore/after the big formatting PR. Basically making it easy to see what happened between v6 and v7, and then changes inside of 7 itself.

@thegedge - I can handle this to coordinate with the major version release, unless you want to get that sweet commit credit.

thegedge:
> I have no strong opinions about formatting other than it should be automatic, checked before merge, and consistent.

💯 I have no strong opinions either (I have my preferences, of course!), but agreed that it just needs to be automatic to not waste anyone's time.

Prettier is of course the go-to tool these days, but I know not everyone is behind its prescriptiveness. I'm not as deep into the ecosystem, so perhaps you have some other suggestions, @coolsoftwaretyler?

If not, I say we use prettier and call it a day. My recommendations:

1. I'd be for switching tot he user of semi-colons. Prettier will add them for us, and we avoid the dreaded (or at least I think it's dreaded) `;(a as xyz).vaue = expression` syntax (used to avoid accidentally making a function call).
2.

thegedge:
Actually, ignore my recommendations. I think we should respect the choices of the organization, which is in disagreement with (1) but not (2).

Let's just copy the prettier settings from https://github.com/mobxjs/mobx

coolsoftwaretyler:
Great find, I agree!

coolsoftwaretyler:
Alrighty, [v7.0.0 is out](https://github.com/mobxjs/mobx-state-tree/releases/tag/v7.0.0), so I'll tackle this sometime soon.

PR #2226, opened 2024-11-27
Feat/add formatting config

coolsoftwaretyler:

Closes https://github.com/mobxjs/mobx-state-tree/issues/2219.

This changes our Prettier config to match the [config for MobX](https://github.com/mobxjs/mobx/blob/main/.prettierrc).

It also runs `bun prettier:write` against the whole codebase, so there's a huge diff here.

We actually have [Husky](https://github.com/typicode/husky) set up to do this as a pre-commit hook already. I think the reason we've had a bunch of diffs is because some of the source files hadn't been given the correct formatting at some point, so as we've made changes throughout the codebase, the Husky hook has formatted them for us.

I've also added a step to run `prettier:check` in CI, so if people skip the hook, we'll see the problem in PRs.

## Steps to validate locally

You should have `[typescript]editor.formatOnSave: true` in `.vscode/settings.json`. Turn that to `false`, then make a change to a file that breaks the Prettier config.

Save the file, add it, and commit it. Husky should clean up your code. 

`bun test` should pass, and `bun run build` should pass and work.

CI should pass on this branch, including the new `check-prettier` job.


coolsoftwaretyler:
Hey @thegedge - whaddya think? I think we've actually had this in place, but the starting point for all our files was off, so it looked like it was broken.

This adopts MobX's style, and also sets up a CI job to check Prettier on PRs.

I notice we're still on TSLint which [has been deprecated for five years lol](https://palantir.github.io/tslint/). I'm gonna open a separate issue to move to ESLint and its TypeScript tooling.

thegedge:
🙌 🎉 

I checked out this branch, saved a file, and it still had changes. Seems like a couple of things:

1. I have things set up, globally, to organize imports. Not sure if we're deviating from mobx here, but I really like having imports sorted and organized (and it removes unused ones too!). What do you think of including https://www.npmjs.com/package/prettier-plugin-organize-imports? Otherwise, I can just turn this off for MST :)
2. Ternaries are formatting differently for me (see screenshot, `src/core/mst-operations.ts`)

CleanShot 2024-11-27 at 16 27 18@2x

coolsoftwaretyler:
Hey @thegedge - I like that import organizing rule as well. Since you're one of the top contributors these days, let's slightly deviate from MobX on this one so you don't need to make any changes yourself. I think that's pretty fair.

I also like removing unused imports automatically. That's a big selling point for me.

I can't reproduce your ternary issue. Any chance that's some other global config overriding the project? This is what I see with format-on-save:


https://github.com/user-attachments/assets/a3aed3da-0f25-43b3-ae20-2d2c24843b36

And if I turn off format-on-save and run `yarn prettier:write`:

https://github.com/user-attachments/assets/27615d1a-eb74-4f62-a9f0-7b1af77808ca

coolsoftwaretyler:
Oh dang, there's some kind of funky import order dependency here that's failing a bunch of tests. I might be inclined to actually drop that commit and just go with the MobX defaults, rather than wrestle with these weird types for now.

If we cut a new branch for bigger rewrite/refactoring, we could turn that plugin on for that endeavor and try to avoid whatever import magic is going on right now.

thegedge:
> I can't reproduce your ternary issue. Any chance that's some other global config overriding the project? This is what I see with format-on-save:

Maybe! I'll double check sometime this week, but feel free to not wait for me. Happy to ship as-is!

coolsoftwaretyler:
Sweet. Shippin it!

PR #2229, opened 2024-12-09
Chore/2227 eslint migration

alyssadicarlo:

This PR addresses the migration from TSLint to TypeScript ESLint. The main goal is to ensure that TypeScript files in src/ and __tests__/ are properly linted using ESLint. However, despite multiple configuration updates, ESLint still reports that the files are being ignored.

The error I’m seeing is:

`You are linting "src/**/*.ts", but all of the files matching the glob pattern "src/**/*.ts" are ignored.
`

After running eslint --debug, I have confirmed that ESLint is correctly loading the .eslintrc.js file, but for some reason, it still considers src/ to be ignored. The cause of this issue is currently unknown, and I’m submitting this PR to request feedback on how to resolve it.

### Questions / Help Needed

- Are there any implicit ignore rules that could be affecting ESLint?
- Could there be any package.json configuration or default ignores at play? To my knowledge, it only automatically ignores node_modules and dist folders.
- Is there any issue with how ESLint is processing ignorePatterns? Should I be including overrides?

### Planned Steps
- [x] Installed TypeScript ESLint dependencies and [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier).
- [x] Removed TSLint dependencies and TSLint configuration file.
- [x] Ran the tslint-to-eslint-config migration tool as recommended in the [official docs](https://typescript-eslint.io/users/what-about-tslint/).
- [ ] Run eslint `bun lint`
- [ ] Handle existing inline lint rule disables (like // tslint:disable-next-line:no-shadowed-variable

I’d appreciate any insight on where the ignore issue may be coming from. Thanks so much for your help!

### DIFF
--- a/docs/API/index.md
+++ b/docs/API/index.md
@@ -1,10 +1,10 @@
 ---
 id: "index"
-title: "mobx-state-tree - v7.0.0"
+title: "mobx-state-tree - v7.0.1"
 sidebar_label: "Globals"
 ---
 
-[mobx-state-tree - v7.0.0](index.md)
+[mobx-state-tree - v7.0.1](index.md)
 
 ## Index
 
@@ -179,7 +179,7 @@ sidebar_label: "Globals"
 
 Ƭ **IDisposer**: *function*
 
-*Defined in [src/utils.ts:35](https://github.com/mobxjs/mobx-state-tree/blob/d5d9f75f/src/utils.ts#L35)*
+*Defined in [src/utils.ts:35](https://github.com/mobxjs/mobx-state-tree/blob/f8bb1472/src/utils.ts#L35)*
 
 A generic disposer.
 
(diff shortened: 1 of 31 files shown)


coolsoftwaretyler:
Thanks for this! I'll take a look soon. If anyone else sees this and has thoughts on it, feel free to chime in.

coolsoftwaretyler:
Hey @alyssadicarlo - thanks for this!

So I'm not exactly sure what the deal is with the error, but I was able to sidestep it by basically starting from scratch. I think there's some kind of issue in the migration script that was generated, and I don't think we need to bang our heads into the wall debugging it.

What do you think about this:

1. [Set up a baseline config file as per typescript-eslint setup](https://typescript-eslint.io/getting-started#step-2-configuration)
2. Go through the lint output and [set all of the errors to warn](https://eslint.org/docs/latest/use/configure/rules#:~:text=To%20change%20a%20rule's%20severity,doesn't%20affect%20exit%20code))
3. Then we can go through and either fix issues that ought to be fixed, or we can decide to adopt those rules specifically.

For this PR, if you do steps 1 and 2, that should be sufficient to get started, and it'll leave a ton of low hanging fruit items for folks in the future for cleanup.

You may also want to rebase against `master`, it looks like you're missing the prettier config changes from https://github.com/mobxjs/mobx-state-tree/pull/2226/files

PR #2231, opened 2024-12-13
Fix/2230 type instantiation excessively deep

### LINKED ISSUE #2230: v7.0.0. type instantiation is excessively deep and possibly infinite
I'm using nested store structure:
```typescript
const ModelProps = Model.props(aProp).props(bProp).props(cProps);
export interface IModelProps extends Instance<typeof ModelProps> {}

const ModelVolatile = ModelProps.volatile(aVolatiles).volatile(bVolatiles).volatile(cVolatiles);
export interface IModelVolatile extends Instance<typeof ModelVolatile> {}

const ModelViews = ModelVolatile.views(aViews);
export interface IModelViews extends Instance<typeof ModelViews> {}

const AAction = ModelViews.actions(a1Actions).actions(a2Actions);
export interface IAAction extends Instance<typeof AAction> {}
const BAction = AAction.actions(b1Actions).actions(b2Actions);
export interface IBAction extends Instance<typeof BAction> {}
...
```
After 7.0.0 update I'm getting 2589 typescript error: type instantiation is excessively deep and possibly infinite.

coolsoftwaretyler:

Fixes #2230 by using mapped types instead of `Omit`. Seems to pass the test case.

## Steps to validate locally

You should not see TS errors about excessively deep type instantiation. All other tests and type checks should work.

### DIFF
--- a/__tests__/core/2230.test.ts
+++ b/__tests__/core/2230.test.ts
@@ -1,0 +1,155 @@
+//github.com/mobxjs/mobx-state-tree/issues/2230
+
+import { describe, test } from "bun:test"
+import { types, Instance } from "../../src/index"
+
+describe("2230 - type instantiation is excessively deep and possibly infinite", () => {
+    test("does not happen", () => {
+        const ModelProps = types
+            .model({
+                prop01: "",
+                prop02: "",
+                prop03: ""
+            })
+            .props({
+                prop11: "",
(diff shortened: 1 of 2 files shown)


thegedge (approved):
Verified failing before and fixed after

Thanks for fixing, @coolsoftwaretyler <3 🙇

PR #2234, opened 2024-12-17
fix: revert 2218 to fix ts excessively deep types

coolsoftwaretyler:

This PR closes https://github.com/mobxjs/mobx-state-tree/issues/2230, by reverting https://github.com/mobxjs/mobx-state-tree/pull/2218. That means we'll have to repoen https://github.com/mobxjs/mobx-state-tree/issues/2216, unfortunately.

In the long term, we'll want to fix these types at the root of the repository. But since we don't have a full roadmap for that, we are going to make a practical choice to prioritize fixing #2230 over fixing #2216.

For anyone who prefers the older fix, you can either patch the package yourself, or run from `v7.0.0`, which will still be available and have #2218 available in its source code.

## Steps to validate locally

Tests should pass, old behavior should be restored. I have modified `__tests__/core/2230.test.ts` to include all of the reproducing code from #2230.


thegedge (approved):
🚀

PR #2245, opened 2025-02-15
Make ts-essentials a regular dependency

### LINKED ISSUE #2233: New type errors in Model.create with version 7.0.0
**_Bug report_**

Originally reported in the v7.0.0 discussion post: https://github.com/mobxjs/mobx-state-tree/discussions/2224#discussioncomment-11550012

Here is a CodeSandbox reproduction.

On v6.0.1, we get a type error correctly: https://codesandbox.io/p/sandbox/naughty-bassi-forked-39c7cz?file=%2Fsrc%2Findex.ts&workspaceId=ws_QGa4cGacuoY7BgaNFKGDj8

Screenshot 2024-12-13 at 12 31 05 PM

On v7.0.0, MST will fail at runtime, but TypeScript does not warn us about that problem: https://codesandbox.io/p/sandbox/gv3tn2?file=%2Fsrc%2Findex.ts

Screenshot 2024-12-13 at 12 31 45 PM

We should receive type errors on the `a` property when passing a string to a `types.number`.

No type error in version 7.0.0

thegedge:

Resolves #2233

When I was working on improving type definitions, I brought in `ts-essentials` as a dev dependency. What I failed to realize was that this library still gets referenced in the generated type declaration files, so we can't have it as a dev dependency.

Also, doing the same thing locally without skipping the lib check (tsconfig.json), you get the following:

```
⠿ npm exec tsc
node_modules/mobx-state-tree/dist/core/type/type.d.ts(2,35): error TS2307: Cannot find module 'ts-essentials' or its corresponding type declarations.
```

This PR fixes that issue by making it a real dependency.

I've opted to follow up with testing that can catch this kind of user, to expedite getting this fix out to our users.

## Steps to validate locally

`npm pack`, and then using a `file:` dependency pointing at the tarball it created. Before this PR you get the above error, and after this error you get a happy `tsc`.

### DIFF
--- a/__tests__/core/api.test.ts
+++ b/__tests__/core/api.test.ts
@@ -1,6 +1,6 @@
-import * as mst from "../../src"
-import { readFileSync } from "fs"
 import { expect, test } from "bun:test"
+import { readFileSync } from "fs"
+import * as mst from "../../src"
 
 function stringToArray(s: string): string[] {
     return s.split(",").map(str => str.trim())
@@ -156,7 +156,7 @@ test("all methods mentioned in API docs", () => {
 })
 
 test("only accepted dependencies", () => {
-    const validDeps: string[] = []
+    const validDeps: string[] = ["ts-essentials"]
 
     const deps =
         JSON.parse(readFileSync(__dirname + "/../../package.json", "utf8")).dependencies || {}
--- a/package.json
+++ b/package.json
@@ -52,6 +52,9 @@
     "files": [
         "dist/"
     ],
+    "dependencies": {
+        "ts-essentials": "^9.4.1"
+    },
     "devDependencies": {
         "@size-limit/preset-big-lib": "^5.0.3",
         "@types/bun": "^1.0.6",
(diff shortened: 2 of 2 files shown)


thegedge:
@coolsoftwaretyler not sure what you think, but we should probably stick to this being a bump to the patch version, so that we're more likely to get it installed for anyone with something like `"mobx-state-tree": "^7.0.x"`

__tests__/core/api.test.ts
```diff
@@ -156,7 +156,7 @@ test("all methods mentioned in API docs", () => {
 })
 
 test("only accepted dependencies", () => {
-    const validDeps: string[] = []
+    const validDeps: string[] = ["ts-essentials"]
```
thegedge, line 159: 😢 I made us have a dependency

coolsoftwaretyler: Womp womp

We'll figure it out! I think it's a reasonable one for now, and we can look into removing it later. Let's ship a patch to fix this in the interim, and I'll put together a follow-up to investigate removing it.

Issue #2277, opened 2025-10-24
Error when React 19 component has MST node as prop with observer defined later


filipkis:
**_Bug report_**

-   [x] I've checked documentation and searched for existing issues [and discussions](https://github.com/mobxjs/mobx-state-tree/discussions)
-   [x] I've made sure my project is based on the latest MST version
-   [x] Fork [this](https://codesandbox.io/p/sandbox/mobx-state-tree-todolist-forked-pj732k) code sandbox or another minimal reproduction.

**Sandbox link or minimal reproduction code**

[CodeSandbox](https://codesandbox.io/p/devbox/mst-component-params-demo-forked-n3txrx?file=%2Fsrc%2FApp.jsx)

**Describe the expected behavior**

When a component is defined as bellow, where `item` prop is a MST Node, if that node changes the component should update without any issue.

```jsx
const Comp = ({ item }) => {
  return <div>Hello {item?.name}</div>;
};

const CompLaterObserved = observer(Comp);
```

**Describe the observed behavior**

However, on React 19, the above `observer` usage pattern spits out errors (one for each property of the node) shown at the end. The component still updates and everything works apparently, but it's annoying and very noisy in the log (especially since it's one per property, so a big model might spint dozens of errors).

If the observer is wrapped directly e.g.

```jsx

const Comp = observer(({ item }) => { ...});
```

then no errors are shown. 

I'm suspecting this might be a `mobx` issue and not an MST one, but I'm not sure even where to start creating a minimal example of pure `mobx` error.

Here is an example of an error:

```
index.js:27 Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree. (Object type: 'Item', Path upon death: '/item', Subpath: 'name', Action: ''). Either detach nodes first, or don't use objects after removing / replacing them in the tree.
    at warnError (mobx-state-tree.module.js:3811:16)
    at ObjectNode.value (mobx-state-tree.module.js:1932:13)
    at ObjectNode.value (mobx-state-tree.module.js:2026:12)
    at ObservableValue.dehanceValue (mobx.esm.js:1695:19)
    at ObservableValue.get (mobx.esm.js:1751:17)
    at ObservableObjectAdministration.getObservablePropValue_ (mobx.esm.js:4782:34)
    at Object.get [as name] (mobx.esm.js:5188:26)
    at ObservableObjectAdministration.get_ (mobx.esm.js:4832:24)
[...]

PR #2278, opened 2025-10-25
Fix React 19 observer pattern warnings (issue #2277)

coolsoftwaretyler:

Related to https://github.com/mobxjs/mobx-state-tree/issues/2277. 

I'm not convinced this is what we *should* do, because it really changes some of the strictness about reading from dead nodes. But it does *seem* to fix the problem in reproducers.

React 19 is doing something that triggers MobX's _interceptReads on MST node properties, which calls assertAlive() for every property access. This creates warnings for alive nodes being read during rendering, although they seem to be false positives (the observability is working as expected).

One way we could fix this is to:

- Modify ObjectNode.unbox() to only assert alive when:
  1. Running an action (writes/mutations), OR
  2. The node is actually dead

This would allow passive property reads on alive nodes without warnings, and maintain existing error detection for genuinely dead nodes. It'll also preserve write protection during actions.

I added React 19 compatibility test suite that sort of mimics the behavior here. Other tests are passing.

## Steps to validate locally

1. Run tests
2. Use this in a scenario similar to [the original reproducer](https://codesandbox.io/p/devbox/mst-component-params-demo-forked-4q23gt?file=%2Fsrc%2FApp.jsx%3A27%2C27-27%2C35&workspaceId=ws_QGa4cGacuoY7BgaNFKGDj8). See no warnings.

We need to consider if this is what we really want to do, so I'd love some feedback on what could go wrong if we relax the liveliness checks.

### DIFF
--- a/__tests__/core/react-19-compatibility.test.ts
+++ b/__tests__/core/react-19-compatibility.test.ts
@@ -1,0 +1,272 @@
+import { t, setLivelinessChecking, getLivelinessChecking, unprotect } from "../../src/index"
+import { getStateTreeNode } from "../../src/internal"
+import { describe, expect, it, beforeEach, afterEach, spyOn } from "bun:test"
+
+/**
+ * Tests for React 19 compatibility, specifically addressing issue #2277
+ * where observer() applied after component definition causes warnings
+ * due to React 19's more aggressive prop inspection.
+ */
+describe("React 19 compatibility", () => {
+    const ItemModel = t.model("Item", {
+        id: t.identifier,
+        name: t.string,
+        description: t.string,
+        count: t.number
+    })
+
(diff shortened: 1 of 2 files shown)


coolsoftwaretyler:
It does look like the issue happened between React 19.1.0 and React 19.2.0. Continuing to investigate what changed.

thegedge:
Yep, I'm game for keeping the default behaviour. I could perhaps propose a tweak: one new mode that warns in development and ignores in production, and make that the default.

Okay just keeping it as-is and guiding users towards how to disable it. Perhaps (maybe just in development) append a bit to the message to point them to `setLivelinessChecking`? I love it when error messages help me out :)

PR #2287, opened 2026-03-20
fix: invalidate safe references after snapshot reconciliation

### LINKED ISSUE #2212: onInvalidated not called when Bad Reference hits Store - React Native Expo
**_Bug report_**

1. Clone this repo: https://github.com/ChristopherGabba/BugHuntingForFrankAndTyler.git 
2. Create expo development build via simulator or via actual device: `npx eas build --profile development --platform ios`
3. Run `npx expo install -c` in terminal 
4. The ignite boilerplate app will open up. Tap `Sign in`
5. Go to the `Podcasts` Tab at the bottom
6. Tap on any `Toggle Favorite` button for any podcast. I hard coded this:

```typescript
export const EpisodeStoreModel = types
  .model("EpisodeStore")
  .props({
    episodes: types.array(EpisodeModel),
    favorites: types.array(
      types.safeReference(EpisodeModel, {
        onInvalidated(event) {
          console.log("INVALIDATED", JSON.stringify(event, null, 4))
        },
      }),
    ),
    favoritesOnly: false,
  })
```
with this:
```typescript
    addFavorite(episode: Episode) {
      const bsEpisode: ReferenceIdentifier[] = ["BS_EPISODE_REFERENCE"]
      store.setProp("favorites", bsEpisode)
    },
```

I would expect the `onInvalidated` hook to call and remove the bad reference immediately instead of crashing the app. 

I've SEVERELY struggled with this behavior with MST with regard to references, and that's why I migrated to `safeReference` so that it wouldn't crash. The app should not crash when a bad reference arrives, it should handle gracefully (aka hit the invalidated function call). 

Real Life Scenario:
1. Fetch data from backend
2. Data was modified by another user (lets say a Todo action and assigned it to another user that isn't in the store)
3. Fetch data again
4. Bad reference arrives and app crashes

The app crashes immediately.

anchmelev:

If this pull request is approved, it will make `safeReference` invalidate bad snapshot identifiers that arrive through `applySnapshot` / reconciliation on an already-attached tree, instead of leaving broken references in place until they throw on access.

Today this already works during initial creation, but not when an invalid reference is introduced into an existing tree. In that case MST skips invalidation while reconciliation is still in progress, so `onInvalidated` never fires and `safeReference` can still crash the app when read.

This change keeps the existing API and delays that invalidation check until the current MST action finishes. That gives reconciliation time to complete, including cases where the target node appears later in the same snapshot, and then MST can either:

- attach the normal invalidation watcher when the target exists
- fire `onInvalidated` with `cause: "invalidSnapshotReference"` and let `safeReference` clean itself up

This PR also adds regression coverage for:

- applying invalid safe references to an existing tree
- snapshots where a reference appears before its target later in the same apply

Closes #2212.

## Steps to validate locally

- `bun run typecheck`
- `bun run build`
- `bun run test:all`

Focused checks:
- `bun test __tests__/core/reference-onInvalidated.test.ts`
- `bun test __tests__/core/reference.test.ts`
- `bun test __tests__/core/action.test.ts`
- `bun test __tests__/core/async.test.ts`
- `bun test __tests__/core/actionTrackingMiddleware2.test.ts`

Additional app-level verification:
- Ran a minimal Expo repro on the iOS simulator against this local MST branch  
issue-2212-ios-pass

- Confirmed invalid snapshot references in `safeReference` model / array / map values trigger `onInvalidated`
- Confirmed invalid refs are removed instead of throwing on first access
- Confirmed references still resolve correctly when the target appears later in the same snapshot

Note:
- The original external repro repository linked in the issue currently returns 404, so I recreated the failing scenario in a minimal Expo app for runtime verification.

### DIFF
--- a/__tests__/core/reference-onInvalidated.test.ts
+++ b/__tests__/core/reference-onInvalidated.test.ts
@@ -299,6 +299,80 @@ describe("safeReference", () => {
         expect(store.single).toBeUndefined()
     })
 
+    test("invalid references applied to an existing tree should be removed", () => {
+        const events: Array<[string, string | number]> = []
+        const onInvalidated: OnReferenceInvalidated<Instance<typeof Todo>> = ev => {
+            events.push([ev.cause, ev.invalidId])
+        }
+        const SafeRef = types.safeReference(Todo, { onInvalidated })
+
+        const Store = types.model({
+            todos: types.array(Todo),
+            single: SafeRef,
+            arr: types.array(SafeRef),
+            map: types.map(SafeRef)
+        })
+
+        const store = Store.create({
+            todos: [{ id: "1" }, { id: "2" }],
+            single: "1",
+            arr: ["1"],
+            map: { a: "1" }
+        })
+        unprotect(store)
+
+        applySnapshot(store, {
+            todos: [{ id: "1" }, { id: "2" }],
+            single: "100",
+            arr: ["100", "2"],
+            map: { a: "100", b: "2" }
+        })
+
+        expect(events).toEqual([
+            ["invalidSnapshotReference", "100"],
+            ["invalidSnapshotReference", "100"],
+            ["invalidSnapshotReference", "100"]
+        ])
+        expect(store.single).toBeUndefined()
+        expect(store.arr.length).toBe(1)
+        expect(store.arr[0]!.id).toBe("2")
+        expect(store.map.size).toBe(1)
+        expect(store.map.get("b")!.id).toBe("2")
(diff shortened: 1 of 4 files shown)


coolsoftwaretyler (approved):
LGTM, this is awesome. Thank you for your work! Will ship the new version soon.

coolsoftwaretyler:
Available in 7.2.0 https://github.com/mobxjs/mobx-state-tree/releases/tag/v7.2.0

PR #2299, opened 2026-08-18
feat: support number options in types.enumeration

### LINKED ISSUE #2243: Enumeration number problem
For some reason numbers are not allowed in enumerations:

```ts
const ordersRequest = types.model({
  paymentPeriod: types.optional(types.array(types.enumeration('PaymentPeriod', [15, 30])), [15]),
})
```

i.e.  for a type like: `(15 | 30)[]`.

As far as I understand, I have to use `types.union` for this:

```ts
const PaymentPeriodEnum = 

const ordersRequest = types.model({
paymentPeriod: types.optional(types.array(types.union(types.literal(15), types.literal(30))), [15]),
})
```

Can numbers be added to enumerations?

mrpmohiburrahman:

If this pull request is approved, `types.enumeration` will accept numbers as well as strings, so a `(15 | 30)` field can be written as an enumeration instead of a hand-rolled union.

Closes #2243.

### The problem

```ts
types.enumeration("PaymentPeriod", [15, 30])
```

fails three ways today. It doesn't compile — `Type 'number' is not assignable to type 'string'` — because the overloads are constrained to `T extends string`. Cast past that and dev mode throws `expected string as argument 1, got 15 instead`. In production the assertion is skipped and the options are coerced silently, so the type ends up accepting the *string* `"15"` and rejecting the number `15`:

```
describe()      = { n: ("15" | "30") }
is({ n: 15 })   = false
is({ n: "15" }) = true
```

The workaround in the issue — `types.union(types.literal(15), types.literal(30))` — already works, which is what suggested the fix would be small.

### Root cause

`literal()` is `<S extends Primitives>` and `ModelPrimitive` already includes `number`, so the layer underneath has accepted numbers all along. `enumeration` narrows them away in three places, and only one is the defect proper:

```ts
const type = union(...realOptions.map(option => literal("" + option)))
```

`"" + option` is a no-op on strings and a silent coercion on anything else. The generic constraint and the `assertIsString` call are the guards that stop anyone reaching it.

### The change

- Widen the three signatures and the exported `UnionStringArray` constraint to `string | number`. `UnionStringArray` keeps its name — it's exported, so renaming it would be breaking.
- Replace `assertIsString(option, i + 1)` with an `assertArg` that takes either, in the shape `identifier.ts:272` already uses. `assertIsString` itself is untouched: its four other callers take genuine strings, and widening the shared helper would stop it catching real mistakes there.
- Pass `option` to `literal()` unchanged.
- The implementation signature now returns `ISimpleType<T>` instead of `ISimpleType<T[number]>`, because `T[number]` stops compiling once `T` can be a number. Both overloads are unchanged, so this isn't visible to callers.

`boolean`, `Date` and `bigint` are also in `ModelPrimitive`. I left them out on purpose — that felt like an API decision worth asking about rather than assuming.

### One behaviour change worth flagging

Outside dev mode, `types.enumeration(["a", 1] as any)` currently coerces `1` to the string `"1"`; after this it becomes `literal(1)`. Anyone relying on that coercion in production would see a snapshot change. Dev mode already threw for this case, so no working dev-mode code can depend on it.

## Steps to validate locally

```
bun install
bun test __tests__/core/enum.test.ts
```

Four tests are added, alongside the six already there:

- **`should support numeric enums`** — the number is accepted, the string of that number is rejected, `describe()` returns `(15 | 30)`, and an out-of-range value throws in dev mode.
- **`should support zero as a numeric enum option`** — added at review request. `0` is the option that would break any truthiness check on the option path, so it is pinned down separately: the number is accepted, `"0"` is not, `describe()` renders `(0 | 1)`, and it round-trips through a snapshot. It passed without a source change, so it is a regression guard rather than a fix.
- **`should support numeric enums inside an optional array`** — the reproduction from the issue, verbatim.
- **`should support mixed string and number enums`** — `["a", 1]`, including that `"1"` is not accepted.

To watch them fail, check out this branch and revert only the source file:

```
git checkout origin/master -- src/types/utility-types/enumeration.ts
bun test __tests__/core/enum.test.ts   # 4 fail
```

The six existing string tests are untouched and still pass. That was the case I was most wary of, since `"" + option` sat on the path every string enum takes.

### What I ran

| command | result |
| --- | --- |
| `bun run typecheck` | clean |
| `NODE_ENV=development bun test` | 1158 pass, 1 skip, 0 fail (1154 on `master`) |
| `MST_TESTING=1 NODE_ENV=production bun test` | 1023 pass, 1 skip, 0 fail (1019 on `master`) |
| `bun run prettier:check` | clean |
| `bun run lint` | 4 errors, identical on `master` (pre-existing, untouched) |

I didn't run `bun run build` or the size-limit check, so `bun run test:all` wasn't run end to end. On docs, I only touched `docs/overview/types.md`, which is hand-written — `docs/API/` is generated and CONTRIBUTING asks for it in a separate commit.


src/types/utility-types/enumeration.ts
```diff
@@ -1,47 +1,54 @@
-import { ISimpleType, union, literal, assertIsString, devMode } from "../../internal"
+import { ISimpleType, union, literal, assertArg, devMode } from "../../internal"
 
+// kept under its original name because it is exported; renaming it would be a breaking change.
```
coolsoftwaretyler:
```suggestion
```

**suggestion:** looks like an artifact comment from an LLM working on a specific task, doesn't really add much value to the codebase. Please remove.

coolsoftwaretyler:
Taking a look now.

__tests__/core/enum.test.ts
```diff
@@ -73,3 +73,34 @@ test("should support readonly enums as const", () => {
         )
     }
 })
+test("should support numeric enums", () => {
```
coolsoftwaretyler, line 76: **suggestion:** let's also test `0` to double check coercion. Something like this maybe?

```
test("should support zero as a numeric enum option", () => {
    const Zero = types.model({ n: types.enumeration("Zero", [0, 1]) })
    expect(Zero.is({ n: 0 })).toBe(true)
    expect(Zero.is({ n: 1 })).toBe(true)
    expect(Zero.is({ n: "0" as any })).toBe(false)
    expect(Zero.is({ n: 2 as any })).toBe(false)
    expect(Zero.describe()).toBe("{ n: (0 | 1) }")

    const z = Zero.create({ n: 1 })
    unprotect(z)
    z.n = 0
    expect(getSnapshot(z)).toEqual({ n: 0 })

        expect(() => (z.n = 2 as any)).toThrow(/Error while converting `2` to `Zero`/)
    }
})
```

coolsoftwaretyler (changes requested):
Looks awesome @mrpmohiburrahman. I have two small asks before we merge.

src/types/utility-types/enumeration.ts
mrpmohiburrahman: Removed in de10974.

__tests__/core/enum.test.ts
mrpmohiburrahman: Added in e0cf9a9.

coolsoftwaretyler (approved):

Issue #1702, opened 2021-04-23
Compatibility with MobX 6 useProxies: "ifavailable" option


BATCOH:
**_Bug report_**

* [x] I've checked documentation and searched for existing issues
* [x] I've made sure my project is based on the latest MST version
* [x] Fork [this](https://codesandbox.io/s/y64pzxj01) code sandbox or another minimal reproduction. 

**Sandbox link or minimal reproduction code**
https://codesandbox.io/s/mobx-state-tree-todolist-ifavailable-t9w2z?file=/index.js

**Describe the expected behavior**
MST should work with `useProxies: "ifavailable"` option from `mobx`.

**Describe the observed behavior**
MST uses `Object.defineProperty` internally in [`addHiddenFinalProp`/`addHiddenWritableProp` utils](https://github.com/mobxjs/mobx-state-tree/blob/8ce9992ff5480e1e7c66803b6692354a1121d801/packages/mobx-state-tree/src/utils.ts#L207-L227) that lead to crash in `ifavailable` mode.
See also [#2876@mobx](https://github.com/mobxjs/mobx/issues/2876)
Since `mobx@6.3.0` released with a convenient `defineProperty` from `object-api`, it seems to be a good idea to use it in MST.

I'm actually working on PR right now and will add it as soon as I can.

jamonholmgren:
5.0.2 is released: https://github.com/mobxjs/mobx-state-tree/releases/tag/v5.0.2

Please test it on your project and report issues.

Thanks again @BATCOH !

PR #2301, opened 2026-08-22
fix: make compiled output compatible with MobX 7

### LINKED ISSUE #2298: MobX 7 compatibility
**Description**

The mobx-state-tree package currently declares MobX ^6.3.0 as a peer dependency in its package.json.

MobX 7 has now been released, so I wanted to check whether mobx-state-tree is compatible with MobX 7.

Would it be possible to clarify whether MobX 7 is supported? If it is compatible, could the peer dependency be updated to include MobX 7 (for example, ^6.3.0 || ^7.0.0)?

If it is not currently compatible, it would also be helpful to know whether there are any known issues or planned work to support MobX 7.

**Relevant links:**

MobX 7.0.0 release: https://github.com/mobxjs/mobx/releases/tag/mobx%407.0.0

hrastnik:

Fixes #2298 — makes MST work with MobX 7. Per review, this is a breaking change: MobX 7 only.

The `ObservableMap` error in that issue is a build configuration problem rather than a source bug. `tsconfig.json` had `"target": "es5"`, which downlevels the one place MST subclasses a MobX class (`MSTMap extends ObservableMap`):

```js
var MSTMap = /** @class */ (function (_super) {
    __extends(MSTMap, _super);
    function MSTMap(initialData, name) {
        return _super.call(this, initialData, function (v) { return v; }, name) || this;
    }
```

MobX 6 also shipped ES5 functions, so `.call()` worked. MobX 7 ships a native `class ObservableMap`, and native constructors throw when invoked without `new`.

Changes:

1. **`tsconfig.json`** — `target: es5` → `es2015`. This is the actual fix.
2. **`tsconfig.json`** — removed `downlevelIteration`, which only existed to serve the ES5 target and is inert at ES2015.
3. **`src/utils.ts`** — `mobxShallow` read `_getGlobalState().useProxies` to decide whether to pass `{ proxy: false }`. MobX 7 removed `configure({ useProxies })` and no longer accepts that option, and since `_getGlobalState()` is typed `any` the field silently became `undefined`, selecting the `{ proxy: false }` branch. MobX 7 is always Proxy-backed, so the check is gone and `mobxShallow` is now just `{ deep: false }`.
4. **`package.json`** — peer dependency narrowed to `^7.0.0`; dev dependency on MobX 7.
5. **tests** — removed `configure({ useProxies })` calls, which no longer typecheck against v7.

No other MobX 7 removals affect MST: it uses no legacy decorators, no namespaced annotations (`observable.ref`, `action.bound`, `comparer.*`), no `trace`, and no `ObservableSet`. Every `_`-prefixed internal it relies on (`_getAdministration`, `_interceptReads`, `_allowStateChangesInsideComputed`, `getAtom`) still exists in v7.

Two notes:

- **Two deleted tests.** `useProxies: 'never'` array behavior and `#1702 - should not throw with useProxies: 'ifavailable'` both existed only to exercise configuration modes MobX 7 removed, so they're no longer expressible. Happy to restore either as a stub if you'd rather keep the issue reference.
- **`version` field.** Left untouched at 7.3.2 — yours to bump on release.

## Steps to validate locally

```
bun install
bun run build
bun run test:all
```

To reproduce the original failure and confirm the fix, save this as `smoke.cjs` in the repo root:

```js
const { types } = require("./dist/mobx-state-tree.js")
const Todo = types.model("Todo", { id: types.identifier, title: types.string })
const Store = types.model("Store", { todos: types.map(Todo) })
const s = Store.create({ todos: { a: { id: "a", title: "A" } } })
console.log("ok:", s.todos.get("a").title, "size:", s.todos.size)
```

With `"target": "es5"` restored in `tsconfig.json` and `bun run build`, `node smoke.cjs` reproduces the error from #2298:

```
TypeError: Class constructor ObservableMap cannot be invoked without 'new'
    at new MSTMap (dist/mobx-state-tree.js:5323:23)
```

With `es2015`, it prints `ok: A size: 1`.


package.json
```diff
@@ -84,7 +84,7 @@
         "yarn-deduplicate": "^3.1.0"
     },
     "peerDependencies": {
-        "mobx": "^6.3.0"
+        "mobx": "^6.3.0 || ^7.0.0"
```
coolsoftwaretyler:
```suggestion
        "mobx": "^7.0.0"
```

**issue:** I think I'm going to call this change a breaking update and push a new major release. We'll drop support for MobX 6, so let's remove it from the peer dependencies as well.

src/utils.ts
```diff
@@ -24,9 +24,12 @@ export const EMPTY_OBJECT: {} = Object.freeze({})
...
  */
-export const mobxShallow = _getGlobalState().useProxies
-    ? { deep: false }
-    : { deep: false, proxy: false }
+// MobX 7 always uses Proxy-backed observables: it dropped the ES5 fallback, removed
+// `configure({ useProxies })` and no longer accepts `{ proxy: false }`. Its global state
+// therefore has no `useProxies` field, so only opt into the non-proxy shape when MobX
+// explicitly reports proxies as disabled (MobX 6 with `useProxies: "never"`).
+export const mobxShallow =
+    _getGlobalState().useProxies === false ? { deep: false, proxy: false } : { deep: false }
```
coolsoftwaretyler: **suggestion:** let's just remove this. I think we should consider this PR a breaking change in MST and release a new major version anyway that only supports MobX 7. that way we can remove this check entirely if MobX 7 is always using proxies.

coolsoftwaretyler (changes requested):
Looks good and works, thanks @hrastnik!

I think we're just going to consider this a breaking change since we have to change our TS target anyway. Asked for a couple changes to help support that. Let me know when it's ready and I can merge/release.

coolsoftwaretyler (approved):
LGTM - I'm gonna ship it as a new major release.

